PR #1, opened 2016-02-22
eval alt

davidmarkclements:
On average I think we're losing maybe 15-30ms by not using eval, however I think there's more room to optimise here (seeing some deopts in both cases, e.g. not a Smi and Unexpected Object). Also util.format is unoptimizable so we might want to rewrite that part too. 

The trade off here is using eval vs limited args and possibly up to 30ms hit (quite a lot - could be optimized more though - also we're still miles ahead). Not using eval may help adoption. 

```sh
with eval:
benchPino*10000: 467.518ms
benchPinoObj*10000: 515.426ms
benchPino*10000: 408.458ms
benchPinoObj*10000: 500.840ms

without:
benchPino*10000: 461.231ms
benchPinoObj*10000: 525.623ms
benchPino*10000: 440.931ms
benchPinoObj*10000: 513.321ms
```

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -34,28 +34,25 @@ function sermon (stream, opts) {
   }
 
   for (var key in levels) {
-    // needed because we cannot put arguments manipulation
-    // in another function without paying a perf drop
+    funcs[key] = function plog (a, b, c, d, e, f, g, h, i, j, k) {
+      var base = 0
+      var obj = null
+      var params = null
+      var msg
+      if (Object(a) === a) {
+        obj = a
+        params = [b, c, d, e, f, g, h, i, j, k]
+        base = 1
+      } else {
+        params = [a, b, c, d, e, f, g, h, i, j, k]
+      }
+      if ((params.length = arguments.length - base) > 0) {
+        msg = format.apply(null, params)
+      }
 
-    eval('' + // eslint-disable-line no-eval
-      'funcs.' + key + ' = function ' + key + ' () {\n' +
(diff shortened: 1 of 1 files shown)


mcollina:
I find the eval thing much more readable.

Also, are unit tests passing? I think `util.format` prints `undefined` for every passed arg that is `undefined`, at least on node v4.

davidmarkclements:
yes - all tests pass - when we set params.length to arguments.length - base it cuts down the params array to actual inputs length and discards the rest

certainly there's things we can do to make it more readable - also is it possible that eval'ing strings is only more readable to you because you wrote it? :{P

mcollina:
Agreed, I've refactored this a bit, and added you as a contributor :D.

davidmarkclements:
sweet - so one other thing - now you've refactored and using closure instead of func-object the function no longer needs to be called "plog" - if you make it anonymous then at least the stack traces can infer function name from object key instead of all logger types being called plog

mcollina:
You are right :).

davidmarkclements:
doing it now - also changing sermon to pino

davidmarkclements:
https://github.com/mcollina/pino/commit/ea686df747154c4dc9a6a84c6aa492a3d88a0363

mcollina:
Added you to NPM as well :).

mcollina:
go ahead and do a patch release :).

davidmarkclements:
done

PR #4, opened 2016-03-07
Some nits after discussing this with @wyatt

mcollina:
@davidmarkclements please review :)


geek:
LGTM

davidmarkclements:
completely agree - each var is already declared so no need for typeof check, more pragmatic to duck type based on used props than check instanceof error - so more versatile in that sense. Not sure it'll give noticeable perf benefits but from code practice pov it's sound

PR #5, opened 2016-03-09
timestamp opt (20-25% speed increase)

davidmarkclements:
When we profile an isolated benchmark with http://npm.im/0x (again) we can see that toISOString is hot (10.8% on stack 8.72% on top of stack) 

screen shot 2016-03-09 at 14 01 02

If we alter out timestamps from ISO to epoch, we get the following flamegraph showing  that `Date.now()` is far less expensive that `Date.prototype.toISOString`

screen shot 2016-03-09 at 14 01 22

Removing the bottleneck gives us a 20-25% perf increase:

benchPino*10000: 369.998ms
benchPino2*10000: 278.406ms
benchPinoObj*10000: 431.747ms
benchPinoObj2*10000: 347.675ms
benchPino*10000: 344.995ms
benchPino2*10000: 272.239ms
benchPinoObj*10000: 420.123ms
benchPinoObj2*10000: 340.240ms

We can keep consistent log output formatting via the `pino` command, by applying the toISOString call at the point of transformation.


davidmarkclements:
oh - also added CLI usage

mcollina:
we should make this configurable via a CLI option.

davidmarkclements:
My contention is we should provide the most optimal approach to logging

We're positioning pino as the fastest logger, so do we really want to add an option that slows it down?

davidmarkclements:
ok after discussion we decided to go with an option with an ominous name "slowtime", a warning in the readme, and the ability to transform json logs with the pino command line tool using the `-t` flag.

PR #8, opened 2016-03-15
No slowdown when logging through child loggers.

mcollina:
@davidmarkclements and @danjenkins. Here you are.

Some notes (all bad):

1. creating a logger is going to be relatively _slow_ (might still be faster than bunyan)
2. the child logger does not inherits anything really, if you change the parent you are going to have issues
3. you can pass a `meta` property to any pino instance, and all the properties in there will be added
4. creating a child of a child does not work, currently
5. the data is made immutable when you create the logger, prejsonified it and stored

Good news: *there is no drawback*.

Any ideas on how to solve point 1-5?

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,9 +155,9 @@ var instance = pino({
 ```
 
 <a name="child"></a>
-### logger.child(opts)
+### logger.child(bindings)
 
-Creates a child logger, setting all key-value pairs is `opts` as
+Creates a child logger, setting all key-value pairs in `bindings` as
 properties in the log lines. All serializers will be applied to the
 given pair.
 
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Into this:
 ## API
 
   * <a href="#constructor"><code><b>pino()</b></code></a>
+  * <a href="#child"><code>logger.<b>child()</b></code></a>
   * <a href="#level"><code>logger.<b>level</b></code></a>
   * <a href="#fatal"><code>logger.<b>fatal()</b></code></a>
   * <a href="#error"><code>logger.<b>error()</b></code></a>
@@ -152,6 +153,25 @@ var instance = pino({
   }
 }
 ```
+
+<a name="child"></a>
+### logger.child(opts)
+
+Creates a child logger, setting all key-value pairs is `opts` as
+properties in the log lines. All serializers will be applied to the
+given pair.
+
+Example:
+
+```js
+logger.child({ a: 'property' }).info('hello child!')
+// generates
+// {"pid":46497,"hostname":"MacBook-Pro-di-Matteo.local","level":30,"msg":"hello child!","time":1458124707120,"v":0,"a":"property"}
+```
+
+It leverage the output stream of the parent and
+its log level. These settings are immutable and pulled out during the
+instantiation of the child logger.
 
 <a name="level"></a>
 ### logger.level
--- a/pino.js
+++ b/pino.js
@@ -147,16 +147,16 @@ function setup (result, funcs, level, stream, opts, serializers, stringify) {
   result.level = opts.level || 'info'
   result.child = child
 
-  function child (opts) {
+  function child (bindings) {
     if (!opts) {
-      throw new Error('missing options for child logger')
+      throw new Error('missing bindings for child logger')
     }
 
     var data = ''
     var value
-    for (var key in opts) {
-      value = opts[key]
-      if (opts.hasOwnProperty(key) && value !== undefined) {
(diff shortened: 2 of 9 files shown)


mcollina:
So, how ofter do you create child loggers? It's probably possible to make creation of the logger a fast operation as well. Currently it's _very very bad_.

I'll give it another pass tomorrow

mcollina:
@danjenkins so I need to make child log creation fast :D.

mcollina:
So, it's faster now (50% faster), but it still 25% slower than not doing it in the first place.

```
benchPinoCreation*10000: 503ms
benchBunyanCreation*10000: 1558ms
benchBoleCreation*10000: 1854ms
benchPinoCreation*10000: 496ms
benchBunyanCreation*10000: 1489ms
benchBoleCreation*10000: 1881ms
```

This is due to the fact that the logging functions will not be optimized yet. If you keep the child around enough time, you will see no worries.


Anybody that will do a code review for me?

davidmarkclements:
@mcollina yah I'll take a look but no capacity till maybe, Fri aft - I want to profile this all with 0x and take a stab as well

mcollina:
Feature wise, after we land this I think pino is "done", and I will mark 1.0.0.

davidmarkclements:
so question - why in the childCreation benchmarks are we doing a log ten times? Doesn't that skew the results?

README.md
```diff
@@ -153,6 +154,25 @@ var instance = pino({
 }
 ```
 
+<a name="child"></a>
+### logger.child(opts)
```
davidmarkclements: not sure that "opts" is the right word? bunyan docs refers to them as "bound fields" so maybe "fields" would be a better description? or "bindings"?

README.md
```diff
@@ -153,6 +154,25 @@ var instance = pino({
...
+
+Creates a child logger, setting all key-value pairs is `opts` as
```
davidmarkclements: "key-value pairs IN `opts`"

pino.js
```diff
@@ -152,13 +152,19 @@ function setup (result, funcs, level, stream, opts) {
       throw new Error('missing options for child logger')
     }
 
-    var meta = JSON.stringify(opts)
-    meta = meta.slice(0, meta.length - 1)
-    meta = meta.slice(1)
+    var data = ''
+    var value
+    for (var key in opts) {
```
davidmarkclements: what if opts has enumerable proto props?

davidmarkclements:
I think we need to add benchmarks for more complex objects, particularly because we stringify(value) (e.g. any nested objects go through JSON.stringify)

davidmarkclements:
yah so I'm not seeing any cost of child logging vs normal logging:

benchPinoObj*10000: 373.550ms
benchPinoChild*10000: 358.387ms
benchPinoObj*10000: 352.645ms
benchPinoChild*10000: 350.087ms

mcollina:
> so question - why in the childCreation benchmarks are we doing a log ten times? Doesn't that skew the results?

All benchmarks are set up in the same way. We might even move all of that into a helper module, so we avoid that situation. Logging ten times is a fairly heavy use case for a logger: we are testing logging 10 times in the same tick.This is more for filling up the underlining buffer for the stream, rather than anything else.

> also - bole "child" isn't really a fair test is it?

No it is not, but pino is faster anyway, so why not?

README.md
mcollina: agreed for bindings.

mcollina:
> there is sometimes a 20ms diff when logging just strings

It highly possible because we are logging more data, I don't think we can do much about it at this point. Maybe with 100k repetitions, but I think we are mostly there.

mcollina:
I have updated for bindings. I think I am happy to call this done.
Child creation can still be better, but we are competing against ourselves at this point.

@davidmarkclements @danjenkins review this tomorrow and then I release :)

davidmarkclements:
I'm happy :+1:  lgtm as the cool kids say

README.md
```diff
@@ -153,6 +154,25 @@ var instance = pino({
...
+
+Example:
+
+```js
+logger.child({ a: 'property' }).info('hello child!')
+// generates
+// {"pid":46497,"hostname":"MacBook-Pro-di-Matteo.local","level":30,"msg":"hello child!","time":1458124707120,"v":0,"a":"property"}
+```
+
+It leverage the output stream of the parent and
```
danjenkins: this should be leverages shouldn't it?

mcollina:
Yes you are right, not sure why Github didn't pick it up automagically.

Released in v1.0.0

PR #10, opened 2016-03-16
Explain why there are no transports.

### LINKED ISSUE #9: API for "transports"
Hi, I'm working on a project where Winston & custom "transports" are used.

I would be happy to replace Winston with pino, because it looks so much better :sunglasses: 

Do you have any API idea for supportings "transports" (I'll be happy to collaborate with some transports for Redis, AMQP, etc.)?

This is what Winston documentation says about transports:

https://github.com/winstonjs/winston/blob/master/docs/transports.md

mcollina:
Fixes #9.

@davidmarkclements @danjenkins please review :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ It also includes a shell utility to pretty-print its log files.
 * [Benchmarks](#benchmarks)
 * [API](#api)
 * [How do I rotate log files?](#rotate)
-* [What happened to my transports?](#transports)
+* [How to use Transports with Pino](#transports)
 * [Acknowledgements](#acknowledgements)
 * [License](#license)
 
@@ -301,14 +301,15 @@ In order to rotate your log files, add in `/etc/logrotate.d/myapp`:
 ```
 
 <a name="transports"></a>
-## What happened to my transports?
+## How to use Transports with Pino
 
-There will never be an API for transports, or support for ObjectMode Writable streams.
-This library is fast because it does way less than the others. We went a
-great length to make sure this library is _really fast_, and transports
+Transports are not part of Pino. There will never be an API for transports,
+or support for ObjectMode Writable streams.
+This library is fast because it does way less than the others. We went
+to great lengths to make sure this library is _really fast_, and transports
 will slow things down.
 
-So, how do you do a transport? You create another node process to do that, and you pipe them together.
+So, how do you do a transport? With Pino, we create a separate process for our transport and pipe to it.
 It's the Unix philosophy.
 
 Something like:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -299,6 +300,28 @@ In order to rotate your log files, add in `/etc/logrotate.d/myapp`:
...
+So, how do you do a transport? You create another node process to do that, and you pipe them together.
+It's the Unix philosophy.
+
+Something like:
+
+```js
+var split = require('split2')
+var pump = require('pump')
+
+pump(process.stdin, split2(JSON.parse), myAwesomeTransport)
```
davidmarkclements: can you show how to create myAwesomeTransport?

README.md
```diff
@@ -299,6 +300,28 @@ In order to rotate your log files, add in `/etc/logrotate.d/myapp`:
...
 
+<a name="transports"></a>
+## What happened to my transports?
+
+There will never be an API for transports, or support for ObjectMode Writable streams.
+This library is fast because it does way less than the others. We went a
+great length to make sure this library is _really fast_, and transports
+will slow things down.
+
+So, how do you do a transport? You create another node process to do that, and you pipe them together.
```
davidmarkclements: With Pino, we create a separate process for our transport and pipe to it.

README.md
```diff
@@ -299,6 +300,28 @@ In order to rotate your log files, add in `/etc/logrotate.d/myapp`:
...
+Something like:
+
+```js
+var split = require('split2')
+var pump = require('pump')
+
+pump(process.stdin, split2(JSON.parse), myAwesomeTransport)
+```
+
+Doing it inside the same process will add unnecessary load to your single-threaded node.js, effectively slowing you down.
```
davidmarkclements: Using transports in the same process causes unnecessary load and slows down Node's single threaded event loop.

README.md
```diff
@@ -10,6 +10,7 @@ It also includes a shell utility to pretty-print its log files.
 * [Benchmarks](#benchmarks)
 * [API](#api)
 * [How do I rotate log files?](#rotate)
+* [What happened to my transports?](#transports)
```
davidmarkclements: maybe: How to use Transports in Pino

davidmarkclements:
some remarks :)

README.md
```diff
@@ -299,6 +300,28 @@ In order to rotate your log files, add in `/etc/logrotate.d/myapp`:
 }
 ```
 
+<a name="transports"></a>
+## What happened to my transports?
+
+There will never be an API for transports, or support for ObjectMode Writable streams.
+This library is fast because it does way less than the others. We went a
```
danjenkins: `We went a great length` doesn't read right, I think it should be `We went to great lengths`

README.md
danjenkins: node.js should have a capital N _sorry_

README.md
danjenkins: Hmmm, I'm not sure, transports aren't _in_ Pino, they're outside of Pino, so maybe `How to use Transports with Pino`

mcollina:
Updated.

PR #13, opened 2016-03-16
roll our own format func

davidmarkclements:
this isn't ready for acceptance, more the beginning of an experiment

I've taken core's util format, altered as follows:

1) accept single param called "args", must be an array
2) do not support non-string args for first arg (util format does this, but who uses that??)
3) convert syntax to conform to standard linting

No conclusive savings as yet, but format is not inlining - target text too big

As an experiment I minified it, which solves target text too big, but instead we get: "target AST is too large [early]" 

Possibly decomposing into smaller funcs is the solution here, needs more tinkering

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,26 +58,32 @@ This produces:
 
 As far as we know, it is the fastest logger in town:
 
+`info('hello world')`:
+
 ```
-benchBunyan*10000: 1230ms
-benchWinston*10000: 2139ms
-benchBole*10000: 1615ms
-benchPino*10000: 314ms
-benchBunyan*10000: 1135ms
-benchWinston*10000: 2025ms
-benchBole*10000: 1635ms
-benchPino*10000: 312ms
-benchBunyanObj*10000: 1480ms
-benchWinstonObj*10000: 2252ms
-benchPinoObj*10000: 409ms
-benchBoleObj*10000: 1765ms
-benchBunyanObj*10000: 1415ms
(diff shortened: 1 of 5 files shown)


davidmarkclements:
we also need benchmarks for multiple args and then for multiple args that use interpolation

once #8 is merged we can add multi-arg to benchmarks folder

davidmarkclements:
rolled out separate module: http://npm.im/quick-format

```
benchPinoInterpolateAll*10000: 398.659ms
benchPinoInterpolateAll2*10000: 399.629ms
benchPinoInterpolateExtra*10000: 1321.408ms
benchPinoInterpolateExtra2*10000: 445.608ms
benchPinoInterpolateAll*10000: 369.683ms
benchPinoInterpolateAll2*10000: 387.859ms
benchPinoInterpolateExtra*10000: 1301.741ms
benchPinoInterpolateExtra2*10000: 444.020ms
```

The main trick is in using stringify instead of inspect in the format function, that's it. 

Whilst there's a 20ms diff with core util.format after 20000 ops in cases where we don't have trailing params, overall the perf benefits are huge.

I did rewrite the function using a `str.split(/%/)` approach, and managed to get the function to inline, but it turned out to be 30ms slower on average than the original - regardless of inlining (flamgraph showed that the bottleneck was the C functions resulting from `str.split(/%/)`. So went back to the core function with a couple of tweaks (mainly, stringify, and cache f.length).

PR #15, opened 2016-03-21
better serialization

davidmarkclements:
in our benchmarks there's no real gain or loss, however in terms of functionality we get better serialization of objects when using interpolation - formally any object with a circular reference would result in the whole object being serialized as [Circular]. Now we label circular references properly, but 2x faster than json-stringify-safe (instead using fast-safe-stringify). This is how we end up with roughly the same benchmarks (because we replace json-stringify-safe for pino and use it instead of tryStringify for quick-format). 

also log format version (v prop) is bumped since circular reference output is different

benchmarks: 

benchPinoMulti*10000: 311.975ms
benchPinoMulti2*10000: 293.021ms
benchPinoInterpolate*10000: 306.162ms
benchPinoInterpolate2*10000: 304.980ms
benchPinoInterpolateAll*10000: 448.825ms
benchPinoInterpolateAll2*10000: 434.358ms
benchPinoInterpolateExtra*10000: 537.417ms
benchPinoInterpolateExtra2*10000: 530.803ms
benchPinoInterpolateDeep*10000: 5872.619ms
benchPinoInterpolateDeep2*10000: 5894.609ms
benchPinoMulti*10000: 287.413ms
benchPinoMulti2*10000: 289.061ms
benchPinoInterpolate*10000: 304.053ms
benchPinoInterpolate2*10000: 300.175ms
benchPinoInterpolateAll*10000: 467.555ms
benchPinoInterpolateAll2*10000: 451.627ms
benchPinoInterpolateExtra*10000: 582.479ms
benchPinoInterpolateExtra2*10000: 572.805ms
benchPinoInterpolateDeep*10000: 5850.142ms
benchPinoInterpolateDeep2*10000: 5877.385ms

### DIFF
--- a/benchmarks/multiArg.js
+++ b/benchmarks/multiArg.js
@@ -73,6 +73,12 @@ var run = bench([
     }
     setImmediate(cb)
   },
+  function benchPinoInterpolate (cb) {
+    for (var i = 0; i < max; i++) {
+      plog.info('hello %s', 'world')
+    }
+    setImmediate(cb)
+  },
   function benchBunyanInterpolateAll (cb) {
     for (var i = 0; i < max; i++) {
       blog.info('hello %s %j %d', 'world', {obj: true}, 4)
--- a/package.json
+++ b/package.json
@@ -39,9 +39,8 @@
   "dependencies": {
     "chalk": "^1.1.1",
     "core-util-is": "^1.0.2",
-    "fast-json-parse": "^1.0.0",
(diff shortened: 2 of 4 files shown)


mcollina:
@davidmarkclements would you mind adding the releases on the github page, pointing to the relevant PRs that were merged and released?

davidmarkclements:
done #16

PR #18, opened 2016-03-31
Added the error serializer

mcollina:
@davidmarkclements please review


davidmarkclements:
I like that we're making this explicit

shouldn't we therefore remove the `.stack` check?
https://github.com/mcollina/pino/blob/errSerializer/pino.js#L100

I think that's wrong anyway, because it will interpret this `{stack: 'meow'}` as an error object too. 

Also - is it redundant to add a test for an alternative type of error, e.g. `try { x = u } catch (e) { pino.error({err: e}) }`  will be a `ReferenceError`

mcollina:
That check does a separate job:

```js
error(new Error('aaaa'))
```

Which will also set the message.

If you pass `{ err: new Error('aaa') }` it is something different: we are using the standard serializer thing.

PR #24, opened 2016-04-04
60% perf increase !!

davidmarkclements:
Ok so this one breaks all the tests, but it still actually functionally equivalent.

Here's the idea - caching log strings up to 2MB, then writing to log (this is why
the tests fail because it's no longer exact input/output). When the process exits
we write whatevers left in the cache (not sure if this is problematic, seems to work, 
maybe wouldn't work if the cache was larger). 

We should probably use the `death` module or some equivalent to make sure we catch all exit scenarios, but just wanted to throw this out there for discussion. 

Here's new benchmarks:

basic:
```
benchBunyan*10000: 1134.796ms
benchWinston*10000: 1828.768ms
benchBole*10000: 1595.314ms
benchPino*10000: 151.062ms
benchBunyan*10000: 1109.045ms
benchWinston*10000: 1877.307ms
benchBole*10000: 1620.304ms
benchPino*10000: 148.871ms
```

object:

```
benchBunyanObj*10000: 1275.334ms
benchWinstonObj*10000: 1913.518ms
benchBoleObj*10000: 1539.534ms
benchPinoObj*10000: 188.245ms
benchBunyanObj*10000: 1251.408ms
benchWinstonObj*10000: 1847.023ms
benchBoleObj*10000: 1555.642ms
benchPinoObj*10000: 184.770ms
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,7 +397,7 @@ This has a couple of important caveats:
 * 4KB of spare RAM will be needed for logging
 * As opposed to the default mode, there is not a one-to-one relationship between calls to logging methods (e.g. `logger.info`) and writes to a log file (or log stream)
 * There is a possibility of the most recently buffered log messages being lost (up to 4KB of logs)
-  * For instance, a powercut will mean no logs are written 
+  * For instance, a powercut will mean up to 4KB of buffered logs will be lost
   * A sigkill (or other untrappable signal) will probably result in the same
   * If writing to a stream other than `process.stdout` or `process.stderr`, there is a slight possibility of lost logs or even partially written logs if the OS buffers don't have enough space, or something else is being written to the stream (or maybe some other reason we've not thought of)
 
--- a/README.md
+++ b/README.md
@@ -162,8 +162,8 @@ Returns a new logger. Allowed options are:
   default `true`
 * `name`: the name of the logger, default `undefined`
 * `serializers`: an object containing functions for custom serialization of objects. These functions should return an JSONifiable object and they should never throw
-* `slowtime`: Outputs ISO time stamps (`'2016-03-09T15:18:53.889Z'`) instead of Epoch time stamps (`1457536759176`). **WARNING**: This option carries a 25% performance drop, we recommend using default Epoch timestamps and transforming logs after if required. The `pino -t` command will do this for you (see [CLI](#cli))
-* `extreme`: Enables extreme mode, yields an additional 60% performance (from 250ms down to 100ms per 10000 ops). There are trade-off's should be understood before usage. See [Extreme mode explained](#extreme).
+* `slowtime`: Outputs ISO time stamps (`'2016-03-09T15:18:53.889Z'`) instead of Epoch time stamps (`1457536759176`). **WARNING**: This option carries a 25% performance drop, we recommend using default Epoch timestamps and transforming logs after if required. The `pino -t` command will do this for you (see [CLI](#cli)). default `false`.
+* `extreme`: Enables extreme mode, yields an additional 60% performance (from 250ms down to 100ms per 10000 ops). There are trade-off's should be understood before usage. See [Extreme mode explained](#extreme). default `false`
 
 `stream` is a Writable stream, defaults to `process.stdout`.
 
--- a/README.md
+++ b/README.md
@@ -9,10 +9,13 @@ It also includes a shell utility to pretty-print its log files.
 * [Usage](#usage)
 * [Benchmarks](#benchmarks)
 * [API](#api)
+* [Extreme mode explained](#extreme)
+* [How to use Pino with Express](#express)
 * [How do I rotate log files?](#rotate)
 * [How to use Transports with Pino](#transports)
 * [Caveats](#caveats)
 * [Changelog](#changelog)
+* [Team](#team)
 * [Acknowledgements](#acknowledgements)
 * [License](#license)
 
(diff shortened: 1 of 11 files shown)


davidmarkclements:
@mcollina ready for review :) 

in particular can you review extreme mode description in readme

davidmarkclements:
ok so I forgot the exit handlers, and also to test the write-on-exit so that's added. 

using console.log (and console.error) to output final buffer (to guarantee the write), 
is all good, but bear in mind it adds a newline. Which means logs will have newline holes in them if  logs are being appended between start/stop

but yeah, def ready for review now.

README.md
```diff
@@ -157,6 +163,7 @@ Returns a new logger. Allowed options are:
 * `name`: the name of the logger, default `undefined`
 * `serializers`: an object containing functions for custom serialization of objects. These functions should return an JSONifiable object and they should never throw
 * `slowtime`: Outputs ISO time stamps (`'2016-03-09T15:18:53.889Z'`) instead of Epoch time stamps (`1457536759176`). **WARNING**: This option carries a 25% performance drop, we recommend using default Epoch timestamps and transforming logs after if required. The `pino -t` command will do this for you (see [CLI](#cli))
+* `extreme`: Enables extreme mode, yields an additional 60% performance (from 250ms down to 100ms per 10000 ops). There are trade-off's should be understood before usage. See [Extreme mode explained](#extreme).
```
mcollina: You should state that the `extreme`  default is `false`.

benchmarks/basic.js
```diff
@@ -8,6 +8,9 @@ var winston = require('winston')
 var fs = require('fs')
 var dest = fs.createWriteStream('/dev/null')
 var plog = pino(dest)
+delete require.cache[require.resolve('../')]
```
mcollina: why this?

pino.js
```diff
@@ -35,10 +37,27 @@ function pino (opts, stream) {
...
+  if (cache) {
+    onExit(function () {
+      if (stream === process.stdout) {
+        console.log(logger.buf)
+      } else if (stream === process.stderr) {
+        console.error(logger.buf)
+      } else {
+        stream.write(logger.buf)
+      }
+      process.exit()
```
mcollina: are you sure this is ok? We are losing the exit code, and possibly interacting with other handlers. Maybe we should do it the other way, let people to add some code to their app? I'm especially concerned about all the SIG* handlers, because if I have some logic added there to do some things when these happens, the process will end up exiting.

test.js
```diff
@@ -522,3 +523,60 @@ test('throw if creating child without bindings', function (t) {
...
+    cb()
+  }))
+
+  child.on('close', function () {
+    t.is(actual, expected)
+    t.is(actual2.trim(), expected2)
+
+    os.hostname = hostname
+    Date.now = now
+    global.process = proc
```
mcollina: we are using `tap`, so you can do `t.tearDown(func)` so we are 100% sure that these get emitted correctly.

mcollina:
Generally 👍 . A couple of problems that we might need to solve before merging, but nothing major.

benchmarks/basic.js
davidmarkclements: for a fair test (results are different otherwise) - deliberately loading two separate instances of pino (avoids polymorphic blah blah etc.) - I don't think it matters for basic but it matters elsewhere, I did it on all benchmarks to be sure

pino.js
davidmarkclements: no you're right, all we have to do is remove process.exit and it should be fine, sorry that shouldn't be there

test.js
davidmarkclements: you mean inside child.on('close' ?

davidmarkclements:
what are you're thoughts on the blank lines from console.log?

pino.js
davidmarkclements: you know I was too fast with this, I see what you're saying now

if you remove exit the process doesn't exit sooooo hmm

davidmarkclements: what if we look for registered listeners on the signals

davidmarkclements:
ok so - exit handling,

here's what we do - pass error codes through exit and beforeExit
uncaughtException pass error code 1
SIGHUP 129
SIGINT 130
SIGQUIT 131
SIGTERM 143

THEN - in the exit handler, we check if anyone else is listening on the captured
event, if they are **we don't exit**, and instead register the handler all over again.

We write out the buffer regardless, but also clear it at that point for cases
where another listener is added, prevents process exit, and continues to log output.

davidmarkclements:
Just adding benchmarks for unsafe and unsafe extreme, check out objects:

```
benchBunyanObj*10000: 1272.168ms
benchWinstonObj*10000: 1944.737ms
benchBoleObj*10000: 1561.756ms
benchPinoObj*10000: 331.931ms
benchPinoUnsafeObj*10000: 328.004ms
benchPinoExtremeObj*10000: 157.349ms
benchPinoUnsafeExtremeObj*10000: 149.055ms
```

multi args also benefit, e.g. info('hello %s %j %d', 'world', {obj: true}, 4) goes from 2s for winston to 270ms in unsafe-extreme mode. Even if just unsafe, you get 15% knocked off normal mode. Deep interpolation is almost as winston levels of slowness though, but unsafe mode drops it down to lowest bench for deep (this is since fixing fast-safe-stringify - which is still 2x json-stringify-safe, bole seems to be doing some kind of lazy serialization, not sure if that's the edge) 

```
benchBunyanInterpolateAll*10000: 1301.128ms
benchWinstonInterpolateAll*10000: 2025.412ms
benchBoleInterpolateAll*10000: 1905.032ms
benchPinoInterpolateAll*10000: 562.888ms
benchPinoExtremeInterpolateAll*10000: 357.521ms
benchPinoUnsafeInterpolateAll*10000: 482.484ms
benchPinoUnsafeExtremeInterpolateAll*10000: 271.497ms
benchBunyanInterpolateExtra*10000: 2653.329ms
benchWinstonInterpolateExtra*10000: 2206.925ms
benchBoleInterpolateExtra*10000: 3394.202ms
benchPinoInterpolateExtra*10000: 712.303ms
benchPinoUnsafeInterpolateExtra*10000: 540.384ms
benchPinoExtremeInterpolateExtra*10000: 489.714ms
benchPinoUnsafeExtremeInterpolateExtra*10000: 355.044ms
benchBunyanInterpolateDeep*10000: 4419.056ms
benchWinstonInterpolateDeep*10000: 15263.350ms
benchBoleInterpolateDeep*10000: 4849.581ms
benchPinoInterpolateDeep*10000: 11768.594ms
benchPinoUnsafeInterpolateDeep*10000: 3308.311ms
benchPinoExtremeInterpolateDeep*10000: 11796.496ms
benchPinoUnsafeExtremeInterpolateDeep*10000: 3290.613ms
```

davidmarkclements:
exit handler improvements - this is mainly for the benchmarks, since we load multiple instances of pino, we have to figure out if the additional event handlers are coming from pino or elsewhere, otherwise the whole thing is messed up in benchmarks (e.g. no ctrl+c to exit etc)

pino.js
```diff
@@ -175,33 +220,55 @@ function asErrValue (err) {
...
-    if (len > 1) {
-      msg = format(params, null)
-    } else if (len) {
-      msg = params[0]
+    onExit.insertCode = function () {
+      if (oneFn.value) { oneFn = once(fn) }
+      oneFn(code, evt)
     }
-    this.stream.write(this.asJson(obj, msg, level))
+    return (code === undefined) ? onExit.passCode : onExit.insertCode
```
mcollina: Are we sure there is no better way to do this?

Maybe we can expose it to the user, something like `logger.flush()`, and a snippet module to do things like

```js
require('on-exit')(logger.flush.bind(logger))
```

And an option `flushOnExit` option?

davidmarkclements: Extreme mode is kind of an advanced mode.... At the same time though, if the user forgets to do that, they're going to lose logs...

Up to you, let me know what you want

mcollina: let's keep it this way, and maybe revise if something else is needed.

mcollina:
👍 LGTM and release as a minor?

davidmarkclements:
npm published - gh released

PR #28, opened 2016-04-13
Support extreme mode with children

mcollina:
Currently setting extreme mode with child will result in lost logs, this fixes it. Perf stays the same here, but give it a spin.

@davidmarkclements please review


davidmarkclements:
yep seems perfect

PR #29, opened 2016-04-18
Do not print 'undefined' when there is no msg in the pretty printer

mcollina:
@davidmarkclements please review :)


davidmarkclements:
good to go

PR #35, opened 2016-05-01
Add JSON to keywords to improve discoverability

satazor:
I came across with `pino` by accident. I couldn't find pino when searching for `log json`.


mcollina:
Thanks, you are right!

PR #44, opened 2016-05-07
on('level-change') and minor tweaks

davidmarkclements:
* onLevelUpdate hook - implementation, tests, readme docs
* typo corrections in readme
* `pino.levels` changed from getter to r/o value, objects are copied instead of directly exposed
* added levels to `Pino.prototype` as r/o value also (helpful for `pino-trace` to get levels from logger instance)
* expose `LOG_VERSION` on `pino` (`module.exports`) and as r/o value on `Pino.prototype` (again helpful for `pino-trace` to `LOG_VERSION` from instance)

@mcollina please review :bowtie:

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,12 +164,14 @@ Into this:
   * <a href="#trace"><code>logger.<b>trace()</b></code></a>
   * <a href="#levelVal"><code>logger.<b>levelVal</b></code></a>
   * <a href="#level-change"><code>logger.on(<b>'level-change'</b>, fn)</code></a>
-  * <a href="#levelValues"><code>pino.levels.<b>values</b></code></a>
-  * <a href="#levelLabels"><code>pino.levels.<b>labels</b></code></a>
+  * <a href="#levelValues"><code>logger.levels.<b>values</b></code> & <code>pino.levels.<b>values</b></code></a>
+  * <a href="#levelLabels"><code>logger.levels.<b>labels</b></code> & <code>pino.levels.<b>labels</b></code></a>
+  * <a href="#log_version"><code>pino.<b>LOG_VERSION</b></code> & <code>logger.<b>LOG_VERSION</b></code></a>
   * <a href="#reqSerializer"><code>pino.stdSerializers.<b>req</b></code></a>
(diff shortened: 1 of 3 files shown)


davidmarkclements:
oh yah - btw - I deliberately avoided using `EventEmitter` so it wouldn't bloat when browserified

mcollina:
I'm really 👎 on not using Node's EventEmitter. We already depend on the Streams, which are based on EE. 

> oh yah - btw - I deliberately avoided using EventEmitter so it wouldn't bloat when browserified

what's the use case? Maybe I'm missing something here.

davidmarkclements:
There is a *potential* use case - pino accepts a streams, but doesn't require them. Which means you can wrap websockets in a really light object that was just the methods pino calls. This allows you to log over websockets in the browser without adding 200kb to your payload. If we use an EE at this point we'll be introduing a lot more code to the browser. I guess it depends on whether we think supporting that case is worth it since it's primarily for server logging

mcollina:
I'm supportive for this use case, but I'm not ok in having a little pub-sub logic bundled into pino (for DRYness). So, I propose to:

1. drop the `onLevelUpdate` thing
2. inherit from EventEmitter if it is already there (`require('stre' + 'am')` in browserify).  
3. avoid inheriting from EE if EE was not bundled
4. call emit if an `emit` function is there

Basically, if you want the extension point, you need to have EE.

davidmarkclements:
yeah fair enough - *but* would that affect creation child loggers? Because we'll be instantiating an EE every time

mcollina:
@davidmarkclements that is a nice question, probably better run benchmarks. Given how the prototype chain work, it is probably a no-op (as you don't need to call `EE.call` for EE).

Another option is to expose the event as a function that you can overwrite.

davidmarkclements:
if it's a function that you overwrite that makes it hard to attach multiple listeners

If you don't have to call super then it's probably okay - I'll check it out

davidmarkclements:
ok converted to event emitter, doesn't bundle events via browserify, does provide functionality if browserify loads events via some other method, doesn't seemt to affect creation speed or logging speed. hooray

mcollina:
LGTM with minor bump

davidmarkclements:
published and released

PR #47, opened 2016-05-11
trailing comma and --mute

nkint:
Used with [mocha](https://mochajs.org/).

### Trailing comma 

to avoid this:
```
    ✓ should raise error if file does not exist[2016-05-11T09:03:34.958Z] INFO (app/: Send the file
```
and instead have this:
```
    ✓ should raise error if file does not exist
[2016-05-11T09:03:34.958Z] INFO (app/: Send the file
```

### CLI option `--mute`
to have test results without logs

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -81,7 +81,7 @@ function pretty (opts) {
 
     if (parsed.err) {
       // pass through
-      return line + '/n'
+      return line
     }
 
     if (mute) {
--- a/pretty.js
+++ b/pretty.js
@@ -48,6 +48,7 @@ function filter (value) {
 
 function pretty (opts) {
   var timeTransOnly = opts && opts.timeTransOnly
+  var mute = opts && opts.mute
 
   var stream = split(mapLine)
   var ctx
@@ -80,7 +81,11 @@ function pretty (opts) {
 
     if (parsed.err) {
       // pass through
-      return line
+      return line + '/n'
+    }
+
+    if (mute) {
+      return ''
     }
 
     if (timeTransOnly) {
@@ -121,7 +126,8 @@ if (require.main === module) {
     console.log(require('./package.json').version)
   } else {
     process.stdin.pipe(pretty({
-      timeTransOnly: arg('-t')
+      timeTransOnly: arg('-t'),
+      mute: arg('--mute')
     })).pipe(process.stdout)
   }
 }
--- a/README.md
+++ b/README.md
@@ -150,6 +150,12 @@ Into this:
 {"pid":14139,"hostname":"MacBook-Pro-3.home","level":30,"msg":"hello world","time":"2016-03-09T15:27:09.339Z","v":0}
 ```
 
+Can skip logs with `--mute` option:
+
+```sh
+cat log | pino --mute
+```
+
 <a name="api"></a>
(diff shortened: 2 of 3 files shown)


pretty.js
```diff
@@ -80,7 +81,11 @@ function pretty (opts) {
 
     if (parsed.err) {
       // pass through
-      return line
+      return line + '/n'
```
mcollina: why you are adding a newline here?
Can you add a comment on why?

mcollina:
Can you please add the option to the README and the CLI help?

mcollina:
Any updates on this one?

nkint:
Only mute.

mcollina:
Can you please make an example of how you would use this? I'm a bit lost, why you need to mute the pretty-printer instead of not having it in place?

nkint:
There are some cases (i.e. using [mocha reporters](https://mochajs.org/#reporters)) where I don't want logs to be shown (i.e. because they are messing up the final mocha reporter output)... 

But if you don't see the point I'll close the pr and keep a personal fork : )

My specific use case:

```js
'use strict';

var expect = require('chai').expect;

function foo() {
  var pino = require('./')()
  pino.info('foo')
  return 'bar';
}

describe('A foo function', () => {
  it('should return bar', (done) => {
    expect(foo()).to.equals('bar')
    done()
  })
})
```

```sh
mocha test_mocha.js --reporter doc | ./pretty.js
    <section class="suite">      <h1>A foo function</h1>      <dl>[2016-05-18T12:38:58.005Z] INFO (3613 on vt-macchip.valueteam.com): foo
        <dt>should return bar</dt>        <dd><pre><code>expect(foo()).to.equals('bar')done();</code></pre></dd>      </dl>    </section>
```

```sh
mocha test_mocha.js --reporter doc | ./pretty.js --mute
    <section class="suite">      <h1>A foo function</h1>      <dl>        <dt>should return bar</dt>        <dd><pre><code>expect(foo()).to.equals('bar')done();</code></pre></dd>      </dl>    </section>
```

nkint:
p.s. with bunyan I'm doing it with a workaround:

```sh
mocha test_mocha.js --reporter doc | bunyan -c '"mute" === "workaround"'
```

mcollina:
I'm 👍 on this. @davidmarkclements?

davidmarkclements:
@mcollina  @nkint ... yeeeeah I think so

Although maybe it should be called `--strip` - because we're relying on parse errors to determine whether we pass the line through the stream, or strip the line out. I think that makes the use case clearer

mcollina:
I think "mute" makes it clearer, but it might be an Italian thing. It's not stripping anything, it's removing all logs from an output, and leaving only everything that is not a recognized log.

davidmarkclements:
strip does imply remove - but I don't feel strongly against mute so either is fine

Also.. should we explain that mute skips logs in the stream by skipping **parseable JSON** because there is a difference. If you use this one stream that has other json, that will be skipped also.

mcollina:
@nkint can you please add a couple of checks so we don't mute valid JSON that are not pino logs?

mcollina:
Closing for now, feel free to reopen.

Issue #52, opened 2016-05-27
feature request: way to disable logging via environment variable 


psnider:
When I run my test suites (using mocha), the test report output is cluttered with log lines.

Is there a way to turn logging off without modifying my code or configuration?
I also don't see a way to disable pino programmatically, e.g. via pino.level.

What I'm looking for is a way to set the log level via an environment variable, including a way to disable logging entirely. For example:
```
PINO_LEVEL='disabled'
```
or
```
PINO_LEVEL='error'
```

I think this could be useful for testing generally, allowing me to change the logging level on an ad-hoc basis without changing code or configuration.

mcollina:
regarding disabling, you are right.

Try:

```js
log.levelVal = 100
```

to disable.

I'm not keen on looking into global envs, but definitely there should be an easy way to disable pino programmatically.

psnider:
I understand not wanting to use an environment variable, 
but I don't know any other way to do this in an ad-hoc manner.
I have many modules that use pino, and don't want to modify them all.

Is there some way to change the log level across modules, perhaps using child loggers?
I couldn't figure out how to do this.

mcollina:
@psnider if you change the log level in the parent, _before_ creating the children, that will be inherited.

mcollina:
You can now do:

```js
pino({
  enabled: !!process.env.NOLOG
})
```

to disable logging, that should be good :)

psnider:
Thanks, works well in 2.5.0

pajtai:
If using `dotenv` all `process.env` variables come over as strings, so it'd have to look something like:

```javascript
pino({
    enabled: ! (process.env.LOG_ENABLED === 'false')
});
```

normancarcamo:
Thanks I was about to ask the same but I read the docs and I end up using "silent" as level and all good.

koenpunt:
> If using `dotenv` all `process.env` variables come over as strings

This is true for all env variables, not only when using dotenv ;)

SeedyROM:
https://github.com/pinojs/pino/blob/master/docs/api.md#loggerlevel-string-gettersetter

There is also a level called silent!

kohtala:
The silent level came with #266 after the enabled was added. Nice to have it mentioned here for those who are looking for it.

PR #53, opened 2016-05-30
Added enable option and support the level 'disable'

mcollina:
as per #52

This should provide the groundwork for disabling the logger programmatically. I will wait for  @davidmarkclements to review this (he is on holiday).

Seems like #47,  @nkint can you have a look as well?

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -57,7 +57,7 @@ function pino (opts, stream) {
     buf: ''
   }
 
-  if (opts.enable === false) {
+  if (opts.enabled === false) {
     level = 'silent'
   }
 
--- a/pino.js
+++ b/pino.js
@@ -19,10 +19,23 @@ var levels = {
   debug: 20,
   trace: 10
 }
+
+// private property
+Object.defineProperty(levels, 'silent', {
+  value: 100,
+  enumerable: false
+})
+
 var nums = Object.keys(levels).reduce(function (o, k) {
   o[levels[k]] = k
   return o
 }, {})
+
+// private property
+Object.defineProperty(nums, '100', {
+  value: 'silent',
+  enumerable: false
+})
 
 function pino (opts, stream) {
   if (opts && opts._writableState) {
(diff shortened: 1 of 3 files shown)


pino.js
```diff
@@ -321,6 +325,10 @@ function onExit (fn) {
   }
 }
 
+// added private disable level
```
davidmarkclements: if it's private, should we prefix a _, also make it non-enumerable?

also suggest `silent` - which makes more sense in the level context

mcollina: It's private because it gets cloned and looked up later.

👍 on silent

mcollina:
@davidmarkclements any other update? I renamed this to silent.

We might want to make it non-enumerable, you are right. I will have have a look at that.

mcollina:
@davidmarkclements using silent level.

Any other feedback?

pino.js
```diff
@@ -19,11 +19,24 @@ var levels = {
...
+
 var nums = Object.keys(levels).reduce(function (o, k) {
   o[levels[k]] = k
   return o
 }, {})
 
+// private property
+Object.defineProperty(nums, '100', {
+  value: 'silent',
+  enumerable: false
```
davidmarkclements: enumerable: false is default, doesn't need to be set

pino.js
```diff
@@ -44,6 +57,10 @@ function pino (opts, stream) {
     buf: ''
   }
 
+  if (opts.enable === false) {
```
davidmarkclements: ....maybe.. opts.silent ? so then `if (opts.silent)  { }`

also - should there be `enable` and `disable` methods, so logging can be turned back on

pino.js
mcollina: I know, I've left it there for increased readability.

pino.js
mcollina: I tend to prefer positive logic, I find this easier to reason about but harder to work with in code.

Basically `enable: false` is for convenience, and just an alias over `level: 'silent'`. It's useful because you can do things like `pino({ level: 'info', enable: !!process.env.NO_LOG }),

davidmarkclements: ok sure but.. it should be `enabled` rather than `enable` since enabled is a state and enable is action and options configure state rather than trigger actions (.... I'm sooooo picky today)

mcollina: updated 👍

davidmarkclements:
👍 happy

davidmarkclements:
oh but wait - what about docs?

davidmarkclements:
perfection.

PR #56, opened 2016-06-13
Test refactoring

mcollina:
I've splitted the big `test.js` into a multiple files, and added code coverage reports with coveralls.


test/extreme.test.js
```diff
@@ -0,0 +1,131 @@
...
+  var i = 44
+  while (i--) {
+    normal.info('h')
+    extreme.info('h')
+  }
+
+  var expected2 = expected.split('\n')[0]
+  var actual2 = ''
+
+  var e = 'global.process = { __proto__: process,  pid: 123456};Date.now = function () { return 1459875739796;};require("os").hostname = function () { return "abcdefghijklmnopqr";};var pino = require("./");var extreme = pino({extreme: true});extreme.info("h")'
```
davidmarkclements: I know I actually added this in the first place, but since we're refactoring maybe we should clean this line up no the col width, i.e.

```
var e = 'global.process = {',
  '__proto__: process',
  'pid: 123456',
  '}',
  ....etc,
  ].join('')
```

davidmarkclements:
once pickie thing that was originally my fault, then good for merge

PR #60, opened 2016-06-19
Cache the pid and hostname globally, flatstr before writing

mcollina:
This put us on par with bole for basic and object benchmarks.

Inspired by https://github.com/rvagg/bole/blob/f3ba893c7f948f459615007767e2ffee519bc14c/bole.js#L13-L14.

^^ @rvagg good job!

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,52 +72,61 @@ This produces:
 <a name="benchmarks"></a>
 ## Benchmarks
 
-As far as we know, it is one of the fastest loggers in town:
+As far as we know Pino is fastest logger in town.
 
 `pino.info('hello world')`:
 
 ```
-benchBunyan*10000: 1752.137ms
-benchWinston*10000: 2642.697ms
-benchBole*10000: 349.301ms
-benchDebug*10000: 621.877ms
-benchLogLevel*10000: 392.642ms
-benchPino*10000: 365.075ms
-benchPinoExreme*10000: 112.848ms
+benchBunyan*10000: 1355.229ms
+benchWinston*10000: 2226.117ms
+benchBole*10000: 291.727ms
+benchDebug*10000: 445.291ms
+benchLogLevel*10000: 322.181ms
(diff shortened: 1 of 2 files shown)


davidmarkclements:
Just need to update readme benchmarks?,

mcollina:
Yours come out nicer (damn maxbook pro 15'')

davidmarkclements:
added benchmarks, we actually be bole in every test - but *only just*

README.md
```diff
@@ -72,52 +72,61 @@ This produces:
 <a name="benchmarks"></a>
 ## Benchmarks
 
-As far as we know, it is one of the fastest loggers in town:
+As far as we know Pino is fastest logger in town.
```
mcollina: I would leave this as before. The difference is so tiny, and the approach between bole and pino is essentially the same now.
20ms over 10000 log lines is hardly perceivable.

README.md
```diff
@@ -72,52 +72,61 @@ This produces:
...
-benchLogLevelInterpolateExtra*10000: 2150.711ms
-benchPinoInterpolateExtra*10000: 706.112ms
-benchPinoUnsafeInterpolateExtra*10000: 674.391ms
-benchPinoExtremeInterpolateExtra*10000: 359.784ms
-benchPinoUnsafeExtremeInterpolateExtra*10000: 346.463ms
+benchDebugInterpolateExtra*10000: 640.001ms
+benchBunyanInterpolateExtra*10000: 2888.825ms
+benchWinstonInterpolateExtra*10000: 2616.285ms
+benchBoleInterpolateExtra*10000: 1313.470ms
+benchLogLevelInterpolateExtra*10000: 1487.610ms
```
mcollina: @rvagg you should use quick-format as well :D.

README.md
davidmarkclements: Fair enough

PR #61, opened 2016-06-21
README docs for elasticsearch

### LINKED ISSUE #34: Upload to ElasticSearch
Similar to #33 but for the ELK stack.

mcollina:
As titled.

@davidmarkclements please review.

Fixes #34.

### DIFF
--- a/README.md
+++ b/README.md
@@ -740,9 +740,11 @@ It is extremely simple to use and setup
 $ node yourapp.js | pino-elasticsearch --host 192.168.1.42
 ```
 
-Assuming ElasticSearch is running on 192.168.1.42. Then, head to your
+Assuming ElasticSearch is running on 192.168.1.42.
+
+Then, head to your
 Kibana instance, and create an index pattern
-(https://www.elastic.co/guide/en/kibana/current/setup.html) on `pino`,
+(https://www.elastic.co/guide/en/kibana/current/setup.html) on `'pino'`,
 the default for `pino-elasticsearch`.
 
 [pino-elasticsearch]: https://github.com/mcollina/pino-elasticsearch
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -728,6 +728,25 @@ Using transports in the same process causes unnecessary load and slows down Node
...
+[pino-elasticsearch][pino-elasticsearch] uploads the log lines in bulk
+to [ElasticSearch][elasticsearch], to be displayed in [Kibana][kibana].
+
+It is extremely simple to use and setup
+
+```sh
+$ node yourapp.js | pino-elasticsearch --host 192.168.1.42
+```
+
+Assuming ElasticSearch is running on 192.168.1.42. Then, head to your
```
davidmarkclements: Would start a new para with then (ie insert two newlines between . and Then

README.md
```diff
@@ -728,6 +728,25 @@ Using transports in the same process causes unnecessary load and slows down Node
...
+Kibana instance, and create an index pattern
+(https://www.elastic.co/guide/en/kibana/current/setup.html) on `pino`,
+the default for `pino-elasticsearch`.
```
davidmarkclements: Maybe spell this out a bit clearer, something like: 

by default pino-elasticsearch uses the value 'pino' so set up an index pattern of 'pino-*' as per docs (link to docs here)

davidmarkclements:
Couple of tweaks, this is awesome btw

README.md
mcollina: The index pattern is `pino`, or `pino*` if we want a star in it. How should I rewrite?

davidmarkclements: Which ever way is the most basic and compatible with kibana - i'd put it in quote marks instead of back ticks though because backticks make it look like we're talking about a pino instance rather than a field value on kibana setup i.e.

'pino'

or maybe `'pino'`

mcollina:
@davidmarkclements please review again :)

davidmarkclements:
good to go

PR #63, opened 2016-06-22
added TypeScript DefinitelyTyped badge

psnider:
Note that the link is to the DefinitelyTyped site, and not the definition file for pino.
I think this is fine, because users of DefinitelyTyped need to install a type management tool,
described at DefinitelyTyped.


mcollina:
Thanks!

PR #64, opened 2016-06-23
Fix some typos on elasticsearch and logstash

dadoonet:
And adds info about the default port for elasticsearch.

Note that I'd use 127.0.0.1 as an example instead of a real IP address.

I'd probably add a note for people using elasticsearch running on another machine (not 127.0.0.1) and would say that they need to set `network.host` in their `elasticsearch.yml` file on elasticsearch server.

And a link to: https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#common-network-settings

### DIFF
--- a/README.md
+++ b/README.md
@@ -737,10 +737,21 @@ to [Elasticsearch][elasticsearch], to be displayed in [Kibana][kibana].
 It is extremely simple to use and setup
 
 ```sh
+$ node yourapp.js | pino-elasticsearch
+```
+
+Assuming Elasticsearch is running on localhost.
+
+If you wish to connect to an external elasticsearch instance (recommended for production):
+
+* Check that you defined `network.host` in your `elasticsearch.yml` configuration file. See [elasticsearch Network Settings documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#common-network-settings) for more details.
+* Launch:
+
+```sh
 $ node yourapp.js | pino-elasticsearch --host 192.168.1.42
 ```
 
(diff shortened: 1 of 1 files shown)


dadoonet:
Actually I looked at the documentation of pino-elasticsearch and the default server is 127.0.0.1 so I guess my first comment does not make sense...

> Note that I'd use 127.0.0.1 as an example instead of a real IP address.

mcollina:
The documentation here needs to be correct, so do you think it's better to specify the configuration of Elasticsearch here for another ip address? Or change it to 127.0.0.1?

I imagine that whoever follows the doc here knows little about Elasticsearch, so maybe adding one more link is ok (also on pino-elasticsearch).

Thanks for reviewing!

dadoonet:
I would probably write something like:

> It is extremely simple to use and setup
>
>```sh
>$ node yourapp.js | pino-elasticsearch
>```
>
>Assuming Elasticsearch is running on localhost.
>
>If you wish to connect to an external elasticsearch instance (recommended for production):
>
>* Check that you defined `network.host` in your `elasticsearch.yml` configuration file. See [elasticsearch Network Settings documentation](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-network.html#common-network-settings) for more details.
>* Launch:
>
>```sh
>$ node yourapp.js | pino-elasticsearch --host 192.168.1.42
>```
>
>Assuming Elasticsearch is running on `192.168.1.42`.

WDYT?

mcollina:
👍 update the PR accordingly (maybe also pino-elasticsearch).

Thanks again for reviewing!

mcollina:
Thx!

PR #71, opened 2016-07-15
Escape '"' in messages.

### LINKED ISSUE #70: When message contains double quote (") pretty doesn't print pretty
var pino = require('pino')
var pretty = pino.pretty()
pretty.pipe(process.stdout)
var log = pino({
    name: 'app',
    safe: true
}, pretty)

log.info('This has a double quote: " ')
log.info('This does not')

{"pid":61920,"hostname":"mpb.local","name":"app","level":30,"msg":"This has a double quote: " ","time":1468521356677,"v":1}
[2016-07-14T18:35:56.682Z] INFO (app/61920 on mpb.local): This does not

mcollina:
Fixes #70.

@davidmarkclements please review.

### DIFF
--- a/example.js
+++ b/example.js
@@ -27,3 +27,5 @@ pino.debug('this is a debug statement')
 
 pino.child({ another: 'property' }).debug('this is a debug statement via child')
 pino.trace('this is a trace statement')
+
+pino.debug('this is a "debug" statement with "')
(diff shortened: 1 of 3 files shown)


davidmarkclements:
:+1: - released v2.7.4 gh tag + npm pub

PR #75, opened 2016-07-25
FIxes #66: now the child loggers can override the serializers

### LINKED ISSUE #66: Support injecting serializers via .child()
Currently it is not possible to add serializers in child loggers.

See https://github.com/mcollina/pino/pull/57 as an example of how it could be implemented.

dgonzalez:
Child loggers can now override or inject serializers.


pino.js
```diff
@@ -248,7 +249,7 @@ Pino.prototype.child = function child (bindings) {
   var key
   for (key in bindings) {
     value = bindings[key]
-    if (key !== 'level' && bindings.hasOwnProperty(key) && value !== undefined) {
+    if (['level', 'serializers'].indexOf(key) === -1 && bindings.hasOwnProperty(key) && value !== undefined) {
```
mcollina: I would prefer to avoid `indexOf`, an `||` will probably be faster.

pino.js
```diff
@@ -9,6 +9,7 @@ var noop = require('./noop')
 var pid = process.pid
 var hostname = os.hostname()
 var baseLog = flatstr('{"pid":' + pid + ',"hostname":"' + hostname + '",')
+var extend = require('util')._extend
```
mcollina: I would not depend on a private property.

pino.js
```diff
@@ -258,7 +259,7 @@ Pino.prototype.child = function child (bindings) {
   return new Pino(
     bindings.level || this.level,
     this.stream,
-    this.serializers,
+    extend(this.serializers, bindings.serializers),
```
mcollina: I think we should only allocate a new object if `bindings.serializers` is present.

pino.js
mcollina: We should probably use http://npm.im/xtend

mcollina:
Thanks @dgonzalez! I've made some notes, let me know what you think.

It's generally good, I would just change some nits.

pino.js
davidmarkclements: Wait isn't this the same as `Object.assign` ?

pino.js
davidmarkclements: what effect does this have on perf, is there a less obvious but more performant way to achieve this?

dgonzalez:
Ill work on it and push the changes as soon as I can. Cheers!

pino.js
davidmarkclements: I guess it's *less* of a deal but we do benchmark child logger creation

pino.js
dgonzalez: There are diferences that I cannot remember right now. Ill have a look
again just to be sure.

pino.js
mcollina: My goal is, there should be no perf degradation _if `bindings.serializers` is not passed in_. If there are by setting `serializers`, it's fine as it is an edge case anyway.

dgonzalez: Lets aim for it!

pino.js
dgonzalez: I think there is margin here to write a fast-all-natural object merger (just saying).

dgonzalez: Looks like `Object.assign` will do it.

mcollina:
@dgonzalez any update?

dgonzalez:
Yes. Pushing today.

2016-08-01 11:03 GMT+01:00 Matteo Collina <notifications@github.com>:

dgonzalez:
@mcollina @davidmarkclements off you go. I have used `xtend` as `Object.assign` was not available in Node 0.12 or earlier (thanks travis-ci!).

davidmarkclements:
can we use https://www.npmjs.com/package/object.assign and then remove when we drop support for 0.12

mcollina:
I'm ok with object.assign, it seems a much better deal.

dgonzalez:
Updated. Out of curiosity... why `object.assign` is preferred over xtend? For me seems way simpler xtend but I don't have any strong feelings...

davidmarkclements:
Good to go far as I can see - great work @dgonzalez

dgonzalez:
Thanks! still unsure if we should just use `shim`. Seems that `polyfill` is better as it won't modify the current objects. But I am easy on that.

davidmarkclements:
yah because we're a library we shouldn't modify environment - should just use a thing which returns `Object.assign` or else substitutes if not available. For me it's about using standard language features in favour of deps as a best practice, and then polyfilling for older. Other than the shim part I don't see any extra complexity in `Object.assign` vs `xtend`, they're essentially the same. The only place where on would win over the other is in performance, and since child logger creation is not a critical performance path it's hardly worth the developer effort of benchmarking them

jstewmon:
Currently, the project README does cite the performance characteristics of child logger creation:

> Child logger creation is fast:
> 
> ```
> benchBunyanCreation*10000: 1291.332ms
> benchBoleCreation*10000: 1630.542ms
> benchPinoCreation*10000: 352.330ms
> benchPinoExtremeCreation*10000: 102.282ms
> ```

Some applications may use child loggers to adjust log settings for tracing requests. (I have such a project.)

So, I wouldn't agree that chile logger creation is not a critical performance path.

That said, these changes appear to only affect child loggers that override serializers, so there shouldn't be any expected performance regression, so the discussion is somewhat academic. However, it seems performance may vary depending on the options passed to `child`...

dgonzalez:
I think a benchmark against child creation without serializers would be handy. I'd say, merge this PR if we are happy with and I will submit another one benchmarking the child creation with and without serializers.

mcollina:
I've spent a fair amount of time to make child logger creation fast :). Let's not waste that!

A condition for this to be LGTM: child logger creation without serializers should have no slowdown.

Does this hold true?

Then the second step is to add a bench for serializers (in general), but that can happen on another PR.

davidmarkclements:
Fair enough y'all, fair enough, forgive my lapse in standards

dgonzalez:
Ill benchmark the child creation with serializers and see how it goes. The impacto should be minimal (if any)

mcollina:
This is good to go, I've checked it and the perf is the same without serializers.

I've just noticed that `object.assign` has a lot of dependencies:

```
pino@2.7.5 /Users/matteo/Repositories/pino
└─┬ object.assign@4.0.4
  ├─┬ define-properties@1.1.2
  │ └── foreach@2.0.5
  ├── function-bind@1.1.0
  └── object-keys@1.0.11
```

I'm really pending in favor of `xtend` now.

dgonzalez:
Replacing it with xtend is trivial. I will do it today if it is not done yet.

mcollina:
Don't worry about it. Let's see how this work in practice, and reasses. I hope to get rid of it in 6 months.

PR #77, opened 2016-07-28
allow serializers to strip keys

davidmarkclements:
If a serializer returns undefined, the key/value shouldn't be added to the JSON

This will allow a serlializer to strip both a key and a value (by returning undefined)

Also, if an object has an undefined value, it gets excluded from stringification, so we should follow the same behaviour

cc @mcollina @emilyrose


mcollina:
LGTM. Does this affect perf anyhow?

PR #82, opened 2016-09-02
Add level first flag to pretty options.

### LINKED ISSUE #81: Modifying the format (or pattern) of the logs
Is it possible to modify the format or pattern of the logs? For example, instead of having the date first, I would like it to print the log level first.

colm-mcqs:
Users can add `levelFirst: true` to the options passed in to pretty() to flip time and level around on the output line. 

e.g. 
`[2016-09-02T20:58:07.827Z] INFO (app/7776 .....` 
becomes 
`INFO [2016-09-02T20:46:21.237Z] (app/6868 ....`

Closes #81

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -48,6 +48,7 @@ function filter (value) {
 
 function pretty (opts) {
   var timeTransOnly = opts && opts.timeTransOnly
+  var levelFirst = opts && opts.levelFirst
 
   var stream = split(mapLine)
   var ctx
@@ -88,7 +89,10 @@ function pretty (opts) {
       return JSON.stringify(value) + '\n'
     }
 
-    line = '[' + new Date(value.time).toISOString() + '] ' + asColoredLevel(value)
+    line = (levelFirst)
+        ? asColoredLevel(value) + ' [' + new Date(value.time).toISOString() + ']'
+        : '[' + new Date(value.time).toISOString() + '] ' + asColoredLevel(value)
+
     line += ' ('
     if (value.name) {
       line += value.name + '/'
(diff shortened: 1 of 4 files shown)


mcollina:
Nice! Can you please add an explanation into the README, a cli flag, and add that to the help?

mcdonnelldean:
I love when people add human features like this!

colm-mcqs:
Will do. 

Opinion on the flag's name? Not too keen on levelFirst but can't think of anything better....

mcollina:
I don't have a better option :/. I think this is pretty handy too!

davidmarkclements:
Unless we extend the concept to a `format` flag that can be an array describing inclusion and order `['level', 'date', 'message']` something like that?

mcollina:
We have two things here:

a. the option in the CLI tool
b. the option in the API

I'm all in for a `format` option in the API. However, I'd like to have a quick switch for this in the CLI tool.

colm-mcqs:
So leave as is for now, add the CLI flag and update readme, etc. 

New PR for format array (possibly pulling levelFirst bool out of API at the same time)?

pretty.js
```diff
@@ -88,7 +89,10 @@ function pretty (opts) {
       return JSON.stringify(value) + '\n'
     }
 
-    line = '[' + new Date(value.time).toISOString() + '] ' + asColoredLevel(value)
+    line = (levelFirst)
+        ? asColoredLevel(value) + ' [' + new Date(value.time).toISOString() + ']'
+        : '[' + new Date(value.time).toISOString() + '] ' + asColoredLevel(value)
```
mcollina: can you please put the date stuff inside a function?

mcollina:
One quick nit and LGTM for me.

mcollina:
Perfect, thanks!

colm-mcqs:
:+1:

PR #86, opened 2016-09-11
Support writing the logs to stderr when i/o is redirected

### LINKED ISSUE #85: Can't log to stderr in tricky case
Hello Matteo,

While playing with pino, I noticed a subtle bug, when stderr is redirected from bash like this:
```
node index.js 2> error.log
```
And *index.js* looks like
```
var pino = require('pino')
var log = pino(process.stderr)
log.warn('will output to stdout')
```

It seems related to the property `_writableState` which is `undefined` in this case.

So the condition `opts && opts._writableState` to assign opts to the stream, can't pass.

Note that it just happens when options are omitted!

Cheers,
Benoit

mcollina:
Fixes #85.

Please review, cc @krazylek

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -46,7 +46,7 @@ Object.defineProperty(nums, '100', {
 })
 
 function pino (opts, stream) {
-  if (opts && opts._writableState) {
+  if (opts && (opts.writable || opts._writableState)) {
     stream = opts
     opts = null
   }
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -291,3 +291,27 @@ test('correctly escape msg strings', function (t) {
 
   instance.fatal('this contains "')
 })
+
+test('correctly support node v4+ stderr', function (t) {
+  t.plan(1)
+
+  // stderr inherits from Stream, rather than Writable
+  var dest = {
+    writable: true,
+    write: function (chunk) {
+      chunk = JSON.parse(chunk)
+      delete chunk.time
+      t.deepEqual(chunk, {
+        pid: pid,
+        hostname: hostname,
+        level: 60,
+        msg: 'a message',
+        v: 1
+      })
+    }
+  }
+
+  var instance = pino(dest)
+
+  instance.fatal('a message')
+})


mcollina:
@jsumners can you have a look here?

jsumners:
@mcollina I don't think this is affected by the extreme mode buffer flushing. `_writableState` doesn't track whether or not the backing file descriptor is blockable. Indeed, I pulled this patch into my feature branch and ran the tests. Everything is checking out just fine.

mcollina:
@jsumners the problem is that `stderr` do not have a `_writableState`, because it inherits directly from `Stream`.

jsumners:
It looks to have one to me:

```js
> process.stderr
WriteStream {
  // *snip*
  _writableState:
   WritableState {
     objectMode: false,
     highWaterMark: 16384,
     needDrain: false,
     ending: false,
     ended: false,
     finished: false,
     decodeStrings: false,
     defaultEncoding: 'utf8',
     length: 0,
     writing: false,
     corked: 0,
     sync: true,
     bufferProcessing: false,
     onwrite: [Function],
     writecb: null,
     writelen: 0,
     bufferedRequest: null,
     lastBufferedRequest: null,
     pendingcb: 0,
     prefinished: false,
     errorEmitted: false,
     bufferedRequestCount: 0,
     corkedRequestsFree: CorkedRequest { next: null, entry: null, finish: [Function] } },
  writable: true,
  allowHalfOpen: false,
  destroyed: false,
  _bytesDispatched: 0,
  _sockname: null,
  _writev: null,
  _pendingData: null,
  _pendingEncoding: '',
  server: null,
  _server: null,
  columns: 130,
  rows: 50,
  _type: 'tty',
  fd: 2,
  _isStdio: true,
  destroySoon: [Function],
  destroy: [Function] }
```

I'm not clear what you want me to investigate here.

mcollina:
The original issue:

```
$ node a.js 2> out
SyncWriteStream {
  domain: null,
  _events: {},
  _eventsCount: 0,
  _maxListeners: undefined,
  fd: 2,
  writable: true,
  readable: false,
  autoClose: false,
  _type: 'fs',
  _isStdio: true,
  destroySoon: [Function],
  destroy: [Function] }
```

The content of `a.js`:

```js
console.log(process.stderr)
```

As you see, it's our old friend `SyncWriteStream`.

jsumners:
I see. I'll dig into it soon and see what I can figure out.

mcollina:
@jsumners this is already passing, just check if it's ok for you.

jsumners:
Yep. The tests pass and the following does what you'd expect:

```js
var pino = require('./')
var log = pino(process.stderr)
log.info(process.stderr)
```

Issue #90, opened 2016-09-18
Remove timestamp from logs


PatrickHeneise:
When used with Docker/k8s, the timestamp is added by the container, basically duplicating the timestamp. Would be great to have an option to disable the timestamp.

mcollina:
Definitely 👍 .

mcollina:
Fixed in v2.11.0.

PR #91, opened 2016-09-18
Add timestamp option to solve issue #90

jsumners:


pino.js
```diff
@@ -245,12 +252,16 @@ Pino.prototype.asJson = function asJson (obj, msg, num) {
...
+}
+
 function getTime () {
-  return Date.now()
+  return flatstr(',"time":' + Date.now())
 }
 
 function getSlowTime () {
-  return '"' + (new Date()).toISOString() + '"'
+  return flatstr(',"time":"' + (new Date()).toISOString() + '"')
```
mcollina: `flatstr` is probably slowing things down here, because we are already doing it later on.

pino.js
```diff
@@ -130,6 +132,11 @@ function Pino (level, stream, serializers, stringify, end, name, slowtime, chind
   this._baseLog = flatstr(baseLog +
     (this.name === undefined ? '' : '"name":' + stringify(this.name) + ','))
 
+  if (this.timestamp === false) {
+    this.time = getNoTime
+    return
```
mcollina: Why are you using `return` here? It seems not correct to me

pino.js
```diff
@@ -115,12 +116,13 @@ Object.defineProperty(pino, 'levels', {
   enumerable: true
 })
 
-function Pino (level, stream, serializers, stringify, end, name, slowtime, chindings, cache, formatOpts) {
+function Pino (level, stream, serializers, stringify, end, name, timestamp, slowtime, chindings, cache, formatOpts) {
```
mcollina: can we avoid to have both `timestamp` and `slowtime`?

pino.js
jsumners: The time string was being flattened in `lscache` so I opted to do here as well. My benchmarks aren't showing any real hit, but I can remove it.

pino.js
jsumners: If `timestamp` is `false` then there is no reason to check for the `slowtime` option. This `return` short circuits that check.

pino.js
jsumners: They are different features. How do you propose reducing the verbosity? I would use an object for a single parameter to this function, but I imagine it's done this way for speed.

pino.js
jsumners: I updated it, but for some reason this comment isn't getting flagged as being on an out-of-date commit.

pino.js
mcollina: I prefer to have it as the first check of the if-else right afterwards.

jsumners: It is amended to be an `if/else if/else`.

mcollina:
Perfect.

PR #95, opened 2016-09-26
Pass through non-pino lines

### LINKED ISSUE #94: Invalid time value on JSON.stringify()
Versions:
- "pino": "^2.9.0"
- node 6

start: `node app.js | pino`

line: `console.log(JSON.stringify({'hello': 'world'}))`

leads to:

```
myproj/node_modules/pino/pretty.js:115
    return new Date(time).toISOString()
                          ^

RangeError: Invalid time value
    at asISODate (myproj/node_modules/pino/pretty.js:115:27)
    at DestroyableTransform.mapLine [as mapper] (myproj/node_modules/pino/pretty.js:94:17)
    at DestroyableTransform.transform [as _transform] (myproj/node_modules/split2/index.js:33:21)
    at DestroyableTransform.Transform._read (myproj/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:159:10)
    at DestroyableTransform.Transform._write (myproj/node_modules/through2/node_modules/readable-stream/lib/_stream_transform.js:147:83)
    at doWrite (myproj/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:313:64)
    at writeOrBuffer (myproj/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:302:5)
    at DestroyableTransform.Writable.write (myproj/node_modules/through2/node_modules/readable-stream/lib/_stream_writable.js:241:11)
    at Socket.ondata (_stream_readable.js:556:20)
    at emitOne (events.js:96:13)

```

jsumners:
Fixes #94

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -47,7 +47,7 @@ function filter (value) {
 }
 
 function isPinoLine (line) {
-  return line.hasOwnProperty('hostname') && line.hasOwnProperty('pid')
+  return line.hasOwnProperty('hostname') && line.hasOwnProperty('pid') && (line.hasOwnProperty('v') && line.v === 1)
 }
 
 function pretty (opts) {
--- a/pretty.js
+++ b/pretty.js
@@ -47,7 +47,7 @@ function filter (value) {
 }
 
 function isPinoLine (line) {
-  var requiredProperties = [ 'name', 'hostname', 'pid' ]
+  var requiredProperties = [ 'hostname', 'pid' ]
   return Object.keys(line).filter(function (k) {
     return requiredProperties.indexOf(k) > -1
   }).length > 0
--- a/pretty.js
+++ b/pretty.js
@@ -46,6 +46,13 @@ function filter (value) {
   return result
 }
 
+function isPinoLine (line) {
+  var requiredProperties = [ 'time', 'name', 'hostname', 'pid' ]
+  return Object.keys(line).filter(function (k) {
+    return requiredProperties.indexOf(k) > -1
+  }).length > 0
+}
+
 function pretty (opts) {
   var timeTransOnly = opts && opts.timeTransOnly
   var levelFirst = opts && opts.levelFirst
@@ -79,7 +86,7 @@ function pretty (opts) {
     var parsed = new Parse(line)
     var value = parsed.value
 
-    if (parsed.err) {
+    if (parsed.err || !isPinoLine(value)) {
       // pass through
       return line + '\n'
     }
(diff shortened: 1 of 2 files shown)


pretty.js
```diff
@@ -46,6 +46,13 @@ function filter (value) {
   return result
 }
 
+function isPinoLine (line) {
+  var requiredProperties = [ 'time', 'name', 'hostname', 'pid' ]
```
mcollina: time is not required anymore https://github.com/mcollina/pino/pull/91 :).
Maybe we should adjust this further.

jsumners: Duh. I'm the one that removed that.

jsumners: There has to be a way to determine if the incoming line is a pino line. I have updated it to depend only on `hostname` and `pid` properties. They are used to build the `(pid on hostname):` prefix, which is not conditional as the code currently stands.

pretty.js
```diff
@@ -46,6 +46,13 @@ function filter (value) {
   return result
 }
 
+function isPinoLine (line) {
+  var requiredProperties = [ 'hostname', 'pid' ]
+  return Object.keys(line).filter(function (k) {
+    return requiredProperties.indexOf(k) > -1
+  }).length > 0
+}
```
mcollina: I think we can remove the Object.keys and loop, and just use a couple of conditions with &&. we should also check that v is 0.

jsumners: I have thought about that myself. Before I do it: what do you think of a new common property named `pino`, like the `v` property? It could be set to the current version. Then the check only needs to be for the presence of the `pino` property.

mcollina:
I would leave the discussion about "pino" property to another PR/issue. However, I think we should just have a chops thing that also works with bole, bunyan etc.

pretty.js
```diff
@@ -46,6 +46,10 @@ function filter (value) {
   return result
 }
 
+function isPinoLine (line) {
+  return line.hasOwnProperty('hostname') && line.hasOwnProperty('pid')
+}
```
mcollina: we should check on v to be 0 as well.

jsumners: If `v` is even present it's currently at 1.

What I am basing the conditions on is https://github.com/mcollina/pino/blob/master/pretty.js#L100 . Without those properties `pretty.js` will not construct a valid reformatted log line. The `v` property is not considered at all for reformatting.

mcollina: I know but it should :). My bad when writing this for the first time :/.

PR #98, opened 2016-09-28
handle undefined returned from toJSON

davidmarkclements:
For instance:

```js
instance.fatal({test: {toJSON: function () { return undefined }}})
```
Will give 

```json
{"pid":84573,"hostname":"x","level":30,"time":1475092820866,"test":undefined,"v":1}
```

If this log is later JSON.parsed it'll explode (unrecognized char 'u' error)


mcollina:
LGTM

davidmarkclements:
note: doesn't affect benchmarks (basic, object, deep-object checked)

davidmarkclements:
merged, released, pub'd v2.12.1

PR #99, opened 2016-09-29
Support log message filtering

bripkens:
Sometimes, especially when processing a large number of events /
requests, applications can go wild and start logging the same stuff
over and over again. It is desirable to filter messages in these
cases, so as to avoid filling up hard drives or to avoid exceeding
log aggregation limits. In other cases, sensitive information
could be filtered out.

This commit adds basic facilities for log message filtering. These
allow the registration of a log message filter which is applied
before (potentially) expensive formatting and serialization
tasks are executed. This is similar to other logging libraries
such as logback in the Java world.

In the end, I'd like to use this as the foundation for a duplicate log
message filter similar to one available in logback:

http://logback.qos.ch/manual/filters.html#DuplicateMessageFilter

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -71,12 +71,14 @@ function pino (opts, stream) {
     size: 4096,
     buf: ''
   }
+  var filter = opts.filter
 
   if (opts.enabled === false) {
     level = 'silent'
   }
 
-  var logger = new Pino(level, stream, serializers, stringify, end, name, timestamp, slowtime, '', cache, formatOpts)
+  var logger = new Pino(level, stream, serializers, stringify, end, name, timestamp, slowtime, '', cache, formatOpts,
+    filter)
   if (cache) {
     // setImmediate is causing a very weird crash:
     //    Assertion failed: (cb_v->IsFunction()), function MakeCallback...
@@ -116,7 +118,8 @@ Object.defineProperty(pino, 'levels', {
   enumerable: true
 })
 
-function Pino (level, stream, serializers, stringify, end, name, timestamp, slowtime, chindings, cache, formatOpts) {
+function Pino (level, stream, serializers, stringify, end, name, timestamp, slowtime, chindings, cache, formatOpts,
+    filter) {
   this.stream = stream
   this.serializers = serializers
   this.stringify = stringify
@@ -127,6 +130,7 @@ function Pino (level, stream, serializers, stringify, end, name, timestamp, slow
   this.chindings = chindings
   this.cache = cache
   this.formatOpts = formatOpts
+  this.filter = filter
   this._setLevel(level)
 
   this._baseLog = flatstr(baseLog +
@@ -289,7 +293,8 @@ Pino.prototype.child = function child (bindings) {
     this.slowtime,
     data,
     this.cache,
-    this.formatOpts)
+    this.formatOpts,
+    this.filter)
 }
 
 Pino.prototype.write = function (obj, msg, num) {
@@ -354,6 +359,9 @@ function asErrValue (err) {
 
 function genLog (z) {
   return function LOG (a, b, c, d, e, f, g, h, i, j, k) {
+    if (this.filter && !this.filter(a, b, c, d, e, f, g, h, i, j, k)) {
+      return
+    }
     var l = 0
     var m = null
     var n = null
@@ -363,7 +371,6 @@ function genLog (z) {
       m = a
       n = [b, c, d, e, f, g, h, i, j, k]
       l = 1
-
       if (m.method && m.headers && m.socket) {
         m = mapHttpRequest(m)
       } else if (m.statusCode) {
--- a/README.md
+++ b/README.md
@@ -225,6 +225,7 @@ Returns a new logger. Allowed options are:
 * `extreme`: Enables extreme mode, yields an additional 60% performance (from 250ms down to 100ms per 10000 ops). There are trade-off's should be understood before usage. See [Extreme mode explained](#extreme). default `false`
 * `level`: one of `'fatal'`, `'error'`, `'warn'`, `'info`', `'debug'`, `'trace'`; also `'silent'` is supported to disable logging.
 * `enabled`: enables logging, defaults to `true`.
+* `filter`: A function which retrieves arguments just as they were passed to logging functions. Filter functions decide whether log messages should be formatted, serialized and logged, i.e. filter functions return a falsy value to discard log messages.
 
 `stream` is a Writable stream, defaults to `process.stdout`.
 
--- a/test/filtering.test.js
+++ b/test/filtering.test.js
@@ -1,0 +1,72 @@
+'use strict'
+
+var test = require('tap').test
+var pino = require('../')
+var sink = require('./helper').sink
+var check = require('./helper').check
+
+test('does not test filters when level is inactive', function (t) {
+  var filter = function () {
+    t.fail('Should should not be called.')
+  }
+
+  var instance = pino({filter: filter}, sink(function () {
+    t.fail('Should not downstream messages')
+  }))
+
+  instance.level = 'warn'
+  instance.info('')
+  t.end()
+})
+
+test('filters can reject messages', function (t) {
+  t.plan(4)
+
+  var filter = function (a) {
+    return a === 'a'
+  }
+  var expected = [{
+    level: 50,
+    msg: 'a'
+  }, {
+    level: 60,
+    msg: 'a'
+  }]
+
+  var instance = pino({filter: filter}, sink(function (chunk, enc, cb) {
+    var current = expected.shift()
+    check(t, chunk, current.level, current.msg)
+    cb()
+  }))
+
+  instance.error('a')
+  instance.warn('b')
+  instance.fatal('a')
+  t.end()
+})
+
+test('child loggers inherit filter from parent', function (t) {
+  t.plan(1)
+
+  var filter = function (a) {
+    return a === 'a'
+  }
+
+  var parent = pino({filter: filter}, sink(function (chunk, enc, cb) {
+    t.deepEqual(chunk, {
+      pid: chunk.pid,
+      hostname: chunk.hostname,
+      v: chunk.v,
+      time: chunk.time,
+      level: 60,
+      msg: 'a',
+      foo: 'bar'
+    })
+    cb()
+  }))
+
+  var child = parent.child({foo: 'bar'})
+  child.warn('b')
+  child.fatal('a')
+  t.end()
+})


jsumners:
My first inclination on this is that it should be handled separately. See https://github.com/mcollina/pino#transports

However, have you benchmarked your changes relative to the master branch?

mcollina:
Thanks for your contribution!! The current implementation is definitely a no-go. This type of usage of arguments causes a deopt, and it will definitely slow things down. Pino is meant to be fast. I'm ok in anything that keeps the current code optimized and with equal (or greater) throughput. 

Regarding deduplication, I agree that a transport might be easier to work with, and be more performant.

Can you please explain a little better your usecase, without mentioning logback?

Regarding filtering of sensitive data, we are working on something nice for this.Stay tuned.

bripkens:
> However, have you benchmarked your changes relative to the master branch?

Without any filtering active:

```
$ git checkout master
$ node benchmarks/basic.js
benchBunyan*10000: 1493.528ms
benchWinston*10000: 2211.598ms
benchBole*10000: 301.056ms
benchDebug*10000: 442.542ms
benchLogLevel*10000: 322.385ms
benchPino*10000: 277.271ms
benchPinoExreme*10000: 105.704ms
benchBunyan*10000: 1499.082ms
benchWinston*10000: 2215.568ms
benchBole*10000: 281.044ms
benchDebug*10000: 432.720ms
benchLogLevel*10000: 308.749ms
benchPino*10000: 262.547ms
benchPinoExreme*10000: 106.387ms

$ git checkout filtering
$ node benchmarks/basic.js
benchBunyan*10000: 1488.006ms
benchWinston*10000: 2245.172ms
benchBole*10000: 298.442ms
benchDebug*10000: 442.922ms
benchLogLevel*10000: 326.200ms
benchPino*10000: 288.153ms
benchPinoExreme*10000: 110.538ms
benchBunyan*10000: 1445.158ms
benchWinston*10000: 2176.479ms
benchBole*10000: 273.919ms
benchDebug*10000: 420.733ms
benchLogLevel*10000: 304.727ms
benchPino*10000: 266.270ms
benchPinoExreme*10000: 107.102ms
```

> My first inclination on this is that it should be handled separately

In theory, and as far as SOC is concerned, I absolutely agree with you. Unfortunately doing it further downstream has two major drawbacks:

 1. Expensive serialization and formatting logic gets executed unnecessarily.
 2. Log message analysis becomes harder as the raw log message is no longer accessible. This makes it expensive to develop log message similarity based approach for filtering.

bripkens:
> This type of usage of arguments causes a deopt, and it will definitely slow things down. 

As far as I know, it will only do so when filtering is actually active. I assumed that this is how v8 behaved, and hence my thought was that a deopt is acceptable / comparable to the creation of a new data structure for the args. Open to suggestions here :).

> Can you please explain a little better your usecase, without mentioning logback?

As mentioned in the PR's opening message: Applications sometimes go mad. They start to log the same or highly similar stuff over and over again. This is often ugly and slows applications down, but isn't a large problem. The problem begins to show when log messages aren't stored on a hard disk, but instead send to log aggregation tools (think splunk, papertrail or kibana). Especially when you don't host log aggregators yourself, i.e. when you pay X for Y GB/month, this can become expensive or cumbersome because you reach log limits (and then you have to buy more capacity / talk to them…).

For debugging, it is often enough to see the same message appear a few times per minute.

This is what I want to achieve: A generic mechanism which logs the same (or similar) messages at most three times per minute. This is what the referenced logback filtering mechanism allows one to build and this is what this PR adds the building blocks for.

mcollina:
> > This type of usage of arguments causes a deopt, and it will definitely slow things down.
> As far as I know, it will only do so when filtering is actually active. I assumed that this is how v8 behaved, and hence my thought was that a deopt is acceptable / comparable to the creation of a new data structure for the args. Open to suggestions here :).

You are right, it does not slow down the basic cases. But using `arguments` is something that should not really be done, even if in this is case does not slow things down in the base case. You can actually avoid doing it by calling `filter(a, b, c, d, e, f, g, h, i, j, k)`. Slighly different API but similar behavior. Will that be ok with you?

> > My first inclination on this is that it should be handled separately
> In theory, and as far as SOC is concerned, I absolutely agree with you. Unfortunately doing it further downstream has two major drawbacks:

> Expensive serialization and formatting logic gets executed unnecessarily.
Log message analysis becomes harder as the raw log message is no longer accessible. This makes it expensive to develop a log message similarity approach for filtering.

It really boils down to how fast is the deduplication algorithm. Is it faster than serialization and formatting? Serialization and formatting is _fast_ in pino.

Overall it does seem a good idea, and very similar to serialization, but different enough to be worth adding.

I think you should be also trying the out-of-process approach, because it might allow to "summarize" logs that have already been written, for batch processing.

@davidmarkclements @jsumners what do you folks thinks?

davidmarkclements:
I think this shouldn't be in core, filtering/redaction is a type of serializer so it should be a separate module that's passed in as a serializer

also a tool that wraps a pino serializer for our of process use would be awesome, the any kind of serializer could be used within or outside

jsumners:
I understand the desire. However, it feels like feature creep to me. Should the logger be the one dealing with extraneous messages? Or should the log manager? In my experience this is handled by the the destination -- http://www.rsyslog.com/doc/rsconf1_repeatedmsgreduction.html

My vote is for a subsequent process to de-dupe. If that is shown to be overly taxing, and doing it within pino is not, then I'd be more accepting.

Further, how do you determine duplicates without keeping track of the previous messages? Part of pino's speed is its fire and forget nature.

bripkens:
> You can actually avoid doing it by calling `filter(a, b, c, d, e, f, g, h, i, j, k)`. Slighly different API but similar behavior. Will that be ok with you?

Certainly. I updated the PR.

> I think you should be also trying the out-of-process approach, because it might allow to "summarize" logs that have already been written, for batch processing.

The thing is, that the out of process approach makes similarity identification much harder. In our Java apps, the similarity is determined solely on format strings. When doing this out-of-process, this format string is not available.

> Further, how do you determine duplicates without keeping track of the previous messages? Part of pino's speed is its fire and forget nature.

This wouldn't be part of pino's core. As such, it is up to the user to decide which can kind of similarity detection is worth it.

davidmarkclements:
Good point @jsumners - having the user supply a function thats right in the hot path is asking for trouble

davidmarkclements:
So the scenario is, we have 30k req/s and that's generate lots and lots of log messages

They're not really duplicates, since each req is unique, but what we're saying is there's a lot of unnecessary info - what we want to do is have a summary over a given timespan (or have a single "representative" log message)

When you have a need to do this, the processing cost this is non-trivial, so I think doing it in process is the wrong move in an evented platform like Node.

While this may require a different approach to summarising/de-duping, it should be preferred because IMO the alternative is a foot-gun.

mcollina:
@bripkens another good option is to "wrap" pino. You will likely achieve the same goal, as I do not think we have a majority here. I'm conflicted, but @davidmarkclements and @jsumners are definitely 👎 .

```js
function mylog  (a, b, c) {
 if (lru.get(a) && lru.get(a) === 42) {
  return 
}
lru.set(a, (lru.get(a) || 0) + 1)
pino.info(a, b, c)
}
```

What you want to achieve is extremely similar to what tracers do. In order to reduce the amount of data (remember async-tracer @davidmarkclements? :P) that needs to be processed, you implement sampling on the logging side of things, by "removing" some log lines with a similar message.

Have you got this sampling logic implemented anywhere? It might be nice to see if we can apply it to async_wrap.

cc @paolochiodi because we were talking about zipkin and sampling before.

bripkens:
> They're not really duplicates, since each req is unique, but what we're saying is there's a lot of unnecessary info - what we want to do is have a summary over a given timespan (or have a single "representative" log message)

This might be *a* use case, but in the case that I am describing this is solely about dropping log messages, not aggregating them in any way.

> Good point @jsumners - having the user supply a function thats right in the hot path is asking for trouble

Not sure I follow. When filtering is so expensive that it is actual impeding the application, then wouldn't formatting and serialization be a problem too? If this is the case, wouldn't it be great when filtering and serialization wouldn't happen at all?

On top of that, filtering becomes much easier and much more performant when format strings are still available.

> Have you got this sampling logic implemented anywhere? It might be nice to see if we can apply it to async_wrap.

Not in JS, but I can describe to you what the Java version does:

 - remember all logged format strings (lru cache with count)
 - when log message with a format string appears more often than what is defined as the threshold, omit the log message
 - after a configurable amount of time has passed, clear the cache

Also, as far as sampling is concerned, Google's tracing policy from `cloud-trace-nodejs` is rather neat:

https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/master/lib/tracing-policy.js

> @bripkens another good option is to "wrap" pino. You will likely achieve the same goal, as I do not think we have a majority here. I'm conflicted, but @davidmarkclements and @jsumners are definitely 👎 .

Looks like it :(

pino.js
```diff
@@ -354,6 +359,9 @@ function asErrValue (err) {
 
 function genLog (z) {
   return function LOG (a, b, c, d, e, f, g, h, i, j, k) {
+    if (this.filter && !this.filter(a, b, c, d, e, f, g, h, i, j, k)) {
```
jsumners: When I read `if (this.filter(...))` I expect `true` to mean "yes, the filter applied." Returning `false` for the skip seems backward to me.

bripkens: I chose `true` to mean "yes, this log message is of relevance" as this is the way `Array.prototype.filter` works. Basically translating this in my head to `retain?` or `downstream?`.

But I don't care much about `true` or `false` for discarding. Let me know what you prefer and I'll change it :).

mcollina: @jsumners the way `Array.prototype.filter` works is unintuitive, but that's ok.

jsumners:
I'm actual rather neutral on this. If it is accepted, though, I think it should include more detail in the documentation that this feature can have a substantial impact on the efficiency of pino.

Thinking about @mcollina's suggestion of a wrapper: the wrapper could be exposed as a static method on the module. That would take the check for the filter out of the hot path and provide a very explicit opt-in to the potentially slower logging.

mcollina:
@jsumners you make a very good point. It can be a very dangerous option :(.

mcdonnelldean:
I'd like the throw in a thumbs down for this. I fully understand feature but I don't get why it doesn't belong outside. There is mention of a missing format string but I don't get what impedance that adds.

Filters are application specific and given the ability to pipe processes it makes complete sense to pipe raw output to a sampler / filter and then to a destination (via transport).

@bripkens Could you maybe clarify specifically what would be missing from the second process? I assume configuration can be loaded with the sample process. Given it's just a stream you could even attach metadata to the log stream so config is carried with the message.

Again I really dislike the idea of this being part of pino given transports / filters are as simple as a Through2 stream in a new process.

davidmarkclements:
@mcdonnelldean 

https://www.npmjs.com/package/dedupe-stream amirite?

For more vague deduping a stream around http://npm.im/bloomrun would be quite effective

bripkens:
Thank you for your comments. After some consideration, I'll write a wrapper around pino to implement the functionality as @mcollina described.

PR #106, opened 2016-10-15
Add support for multiple streams

jsumners:
This PR adds a wrapper around `pino` that allows multiple destination streams to be specified. Example (from the readme):

```js
var fs = require('fs')
var mspino = require('pino/multi-stream')
var streams = [
  {stream: fs.createWriteStream('/tmp/info.stream.out')},
  {level: 'fatal', fs.createWriteStream('/tmp/fatal.stream.out')}
]
var log = mspino({streams: streams})

log.info('this will be written to /tmp/info.stream.out')
log.fatal('this will be written to /tmp/fatal.stream.out')
```

This further increases the compatibility with the Bunyan API. It does, however, come with significant cost. On my machine, the included benchmark, which goes to 10 destination streams, run three consecutive times results in:

```
benchBunyan*10000: 1835.647ms
benchMSPino*10000: 1605.660ms

benchBunyan*10000: 1945.961ms
benchMSPino*10000: 1643.411ms

benchBunyan*10000: 1909.204ms
benchMSPino*10000: 1684.122ms
```

I'm sure the code can be improved, but I wanted to get feedback on if it should be included or not before thinking too hard on it. What do you think @mcollina and @davidmarkclements ?

### DIFF
--- a/multi-stream.js
+++ b/multi-stream.js
@@ -35,38 +35,31 @@ function pino (opts, stream) {
 
   function MSPino (_loggers) {
     this.loggers = _loggers
+    var levels = ['fatal', 'error', 'warn', 'info', 'debug', 'trace']
+    for (var i = 0, j = levels.length; i < j; i += 1) {
+      if (this.loggers.hasOwnProperty(levels[i]) === false) {
+        this[levels[i]] = noop
+      }
+    }
   }
   MSPino.prototype = Object.create(realPino.Pino.prototype)
   MSPino.constructor = MSPino
 
-  function doLog (logger, level, params) {
-    logger[level].apply(logger, params)
-  }
-  MSPino.prototype._issueLog = function issueLog (level, params) {
-    if (!this.loggers[level]) return noop()
-    for (var i = 0, j = this.loggers[level].length; i < j; i += 1) {
-      doLog(this.loggers[level][i], level, params)
+  function genLog (level) {
+    return function LOG () {
+      for (var x = 0, y = this.loggers[level].length; x < y; x += 1) {
+        var logger = this.loggers[level][x]
+        logger[level].apply(logger, arguments)
+      }
     }
   }
+  MSPino.prototype.fatal = genLog('fatal')
+  MSPino.prototype.error = genLog('error')
+  MSPino.prototype.warn = genLog('warn')
+  MSPino.prototype.info = genLog('info')
+  MSPino.prototype.debug = genLog('debug')
+  MSPino.prototype.trace = genLog('trace')
 
-  MSPino.prototype.fatal = function fatal () {
-    this._issueLog('fatal', arguments)
-  }
-  MSPino.prototype.error = function error () {
-    this._issueLog('error', arguments)
-  }
-  MSPino.prototype.warn = function warn () {
-    this._issueLog('warn', arguments)
-  }
-  MSPino.prototype.info = function info () {
-    this._issueLog('info', arguments)
-  }
-  MSPino.prototype.debug = function debug () {
-    this._issueLog('debug', arguments)
-  }
-  MSPino.prototype.trace = function trace () {
-    this._issueLog('trace', arguments)
-  }
   MSPino.prototype.child = function child (bindings) {
     var levels = Object.keys(this.loggers)
     var childLoggers = {}
--- a/multi-stream.js
+++ b/multi-stream.js
@@ -42,7 +42,7 @@ function pino (opts, stream) {
       }
     }
   }
-  MSPino.prototype = Object.create(realPino.Pino.prototype)
+  MSPino.prototype = Object.create(realPino._Pino.prototype)
   MSPino.constructor = MSPino
 
   function genLog (level) {
--- a/multi-stream.js
+++ b/multi-stream.js
@@ -46,10 +46,37 @@ function pino (opts, stream) {
   MSPino.constructor = MSPino
 
   function genLog (level) {
-    return function LOG () {
-      for (var x = 0, y = this.loggers[level].length; x < y; x += 1) {
-        var logger = this.loggers[level][x]
-        logger[level].apply(logger, arguments)
+    return function LOG (arg1, arg2, arg3, arg4, arg5, arg6) {
+      for (var i = 0, j = this.loggers[level].length; i < j; i += 1) {
+        var logger = this.loggers[level][i]
+        var x
+        var args
+        switch (arguments.length) {
+          case 1:
+            logger[level](arg1)
+            break
+          case 2:
+            logger[level](arg1, arg2)
+            break
+          case 3:
+            logger[level](arg1, arg2, arg3)
+            break
+          case 4:
+            logger[level](arg1, arg2, arg3, arg4)
+            break
+          case 5:
+            logger[level](arg1, arg2, arg3, arg4, arg5)
+            break
+          case 6:
+            logger[level](arg1, arg2, arg3, arg4, arg5, arg6)
+            break
+          default:
+            args = [arg1, arg2, arg3, arg4, arg5, arg6]
+            for (x = 7; x < arguments.length; x += 1) {
+              args[x - 1] = arguments[x]
+            }
+            logger[level].apply(logger, args)
+        }
       }
     }
   }
--- a/multi-stream.js
+++ b/multi-stream.js
@@ -1,0 +1,86 @@
+'use strict'
+
+var realPino = require('./pino')
+var noop = require('./noop')
+
+function pino (opts, stream) {
+  if (opts && (opts.writable || opts._writableState)) {
+    return realPino(null, opts)
+  }
+
+  if (opts.hasOwnProperty('stream') === true) {
+    return realPino(opts, opts.stream)
+  }
+
+  if (opts.hasOwnProperty('streams') === false) {
+    return realPino(opts, stream)
+  }
+
+  if (Array.isArray(opts.streams) === false) {
+    return realPino(opts, opts.streams)
+  }
+
+  var streams = opts.streams
+  var loggers = {}
+  for (var i = 0, j = streams.length; i < j; i += 1) {
+    var _opts = Object.create(opts)
+    var s = streams[i]
+    _opts.level = (s.level) ? s.level : 'info'
+    if (loggers[_opts.level]) {
+      loggers[_opts.level].push(realPino(_opts, s.stream))
+    } else {
+      loggers[_opts.level] = [realPino(_opts, s.stream)]
+    }
+  }
+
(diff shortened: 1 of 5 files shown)


mcollina:
Thanks @jsumners!!!!! This is absolutely awesome, because it explains __why__ this feature is not in pino: there is no way to make it fast.

I do not think this is ok to be merged in, as it will violate the "pino way".

I think the same logic can be implemented with a transport, meaning a stream that decodes the message and split it across multiple files.

davidmarkclements:
Ouuuch 🤕

even if we could make it faster (or at least as fast as bunyan) it should be a sep. module

External process transport (as @mcollina says) or just logfile post processing seems to be the way to go for perf

jsumners:
The reason this isn't a post process transport is because it is a drop in replacement for API compatibility. It will allow projects like https://github.com/verdaccio/verdaccio to migrate to pino slowly. 

I can make it a separate module, say 'pino-multi-stream', and still make it this wrapper as it is written. But I think that adds an unusual way of getting the logger. As I implemented it here, it's still an explicit opt-in with a clear explanation why it is so. 

Do you still want it separated out?

test/multi-stream.test.js
```diff
@@ -0,0 +1,63 @@
...
+
+var fs = require('fs')
+var test = require('tap').test
+var mspino = require('../multi-stream')
+
+test('sends to multiple streams', function (t) {
+  var streams = [
+    {stream: fs.createWriteStream('/tmp/info.out')},
+    {level: 'debug', stream: fs.createWriteStream('/tmp/debug.out')},
+    {level: 'fatal', stream: fs.createWriteStream('/tmp/fatal.out')}
```
mcollina: you should clean up the files afterwards.

mcollina: Even better, you might use flush-write-stream (or a Writable) and check the content with that.

test/multi-stream.test.js
```diff
@@ -0,0 +1,63 @@
...
+    t.is(res.msg, 'info stream')
+
+    res = JSON.parse(fs.readFileSync('/tmp/debug.out'))
+    t.is(res.msg, 'debug stream')
+
+    res = JSON.parse(fs.readFileSync('/tmp/fatal.out'))
+    t.is(res.msg, 'fatal stream')
+
+    t.done()
+  })
```
mcollina: `setImmediate()` do not offer any guarantee that the data has been flushed.

test/multi-stream.test.js
```diff
@@ -0,0 +1,63 @@
...
+    t.is(res.child, 'one')
+
+    t.done()
+  })
+})
+
+test('supports grandchildren', function (t) {
+  var streams = [
+    {stream: fs.createWriteStream('/tmp/multigrandchild.out')}
+  ]
```
mcollina: you should probably check multiple streams here.

mcollina:
@jsumners as it stands, there is very little advantage of moving from bunyan to pino/multi-stream. If we can increase the gain a bit more to have at least a 2x (or close) on bunyan, I think it might be worth having it in here.

multi-stream.js
```diff
@@ -0,0 +1,86 @@
...
+
+  function doLog (logger, level, params) {
+    logger[level].apply(logger, params)
+  }
+  MSPino.prototype._issueLog = function issueLog (level, params) {
+    if (!this.loggers[level]) return noop()
+    for (var i = 0, j = this.loggers[level].length; i < j; i += 1) {
+      doLog(this.loggers[level][i], level, params)
+    }
+  }
```
mcollina: Ideally we should avoid this loop, and just redirect each log message to where it should go.

mcollina: are you sending the log to _every_ logger in this way.

multi-stream.js
```diff
@@ -0,0 +1,86 @@
...
+
+  function MSPino (_loggers) {
+    this.loggers = _loggers
+  }
+  MSPino.prototype = Object.create(realPino.Pino.prototype)
+  MSPino.constructor = MSPino
+
+  function doLog (logger, level, params) {
+    logger[level].apply(logger, params)
+  }
```
mcollina: we should restrain for doing apply and passing arrays around, it is slower than normal function calls. We almost never do it in pino.js (we have written quick-format for that purpose!)

jsumners:
Then let's see how things improve with the code clean-up. If it doesn't reach that threshold then I'll break it out into a separate module. The would have to be a small patch to `pino` to expose the `Pino` prototype, though.

multi-stream.js
jsumners: I don't know how else the real `pino` log method can be invoked then. If I expand the parameters to `a, b, ..., k` then any parameter that isn't passed in to the wrapper function gets sent to the real `pino` method as `undefined`. This results in incorrect logs being generated. By directly passing the parameters via`apply` the real `pino` method operates correctly.

multi-stream.js
```diff
@@ -0,0 +1,86 @@
...
+    this._issueLog('warn', arguments)
+  }
+  MSPino.prototype.info = function info () {
+    this._issueLog('info', arguments)
+  }
+  MSPino.prototype.debug = function debug () {
+    this._issueLog('debug', arguments)
+  }
+  MSPino.prototype.trace = function trace () {
+    this._issueLog('trace', arguments)
```
mcollina: Oh, I would have bet on it.

Try:

```js
  function genLog (level) {
    return function LOG (a, b, c, d, e, f, g, h, i, j, k) {
        return logger[level](a, b, c, d, e, f, g, h, i, j, k)
    }
  }
```

multi-stream.js
jsumners: For clarity, given 3 `info` streams and 1 `debug` stream by the client, `mspino` creates an internal hash:

```js
{
  info: [pinoInstance, pinoInstance, pinoInstance],
  debug: [pinoInstance]
}
```

Thus, for one `info` message `mspino` has to invoke the `info` method on three different instances of `pino`. There isn't a "default" logger.

mcollina:
Try something like https://github.com/nodejs/node/blob/master/lib/timers.js#L669-L684.

multi-stream.js
mcollina: probably you are better off in generating a flamegraph with 0x and seeing if there is an unoptimized method around.

jsumners:
My initial benchmarks were off. I had copied and pasted the Bunyan method from the basic benchmark and forgot to add invocations to all of the streams defined for it. Fix that, and adding a more reasonable benchmark for only 4 destination streams, I get:

```
benchBunyanTen*10000: 8255.483ms
benchMSPinoTen*10000: 2010.881ms
benchBunyanFour*10000: 3461.368ms
benchMSPinoFour*10000: 782.451ms

benchBunyanTen*10000: 7716.261ms
benchMSPinoTen*10000: 1705.099ms
benchBunyanFour*10000: 2944.636ms
benchMSPinoFour*10000: 858.713ms

benchBunyanTen*10000: 8142.875ms
benchMSPinoTen*10000: 1992.805ms
benchBunyanFour*10000: 3288.004ms
benchMSPinoFour*10000: 722.918ms
```

So it's *definitely* faster than Bunyan. Still ludicrously slow, but there's not getting around that.

I commented out the Bunyan benchmarks and ran it through `0x`. Unless I just don't know how to read the graph, nothing in `mspino` is unoptimized:

davidmarkclements:
@jsumners nothing be unoptd reinforces the point that this api is generally a bad idea

I would also check inlining as that's treated separately in trace output (be interested to see)

Another thing to remember, is even if a func is optimised and inline it still may be possible to make the algorithm faster (possibly with a trade off of some kind) 

Finally, I'm very strongly against having this in core - I think we should flip it to a module that's specifically for pino-Bunyan migration (e.g gives 100% compat): pino-bunyan-api - it should also(as Matteo says) at least be 2x (or I'll even go for 1.5x) faster. It's also be good if there's some kind of encouragement towards full migration, within the transition wrapping module

mcollina:
What is the overhead that this introduces for just one stream?

Even if we decide not to include this, If you need a constructor to be exposed to turn this into a module, it won't
be a problem.

jsumners:
@davidmarkclements how do I check inlining?

If this is broken out into a separate module I would definitely expand the documentation and include a stronger warning within it. How would you implement encouragement within the wrapper itself? `util.deprecate`?

jsumners:
@mcollina for one stream:

**multi-stream**
```
benchBunyanOne*10000: 875.099ms
benchMSPinoOne*10000: 224.924ms
```

**basic** (non-relevant lines snipped)
```
benchBunyan*10000: 978.472ms
benchPino*10000: 231.176ms
benchBunyan*10000: 899.794ms
benchPino*10000: 223.952ms
```

mcollina:
I'm moving towards accepting this. The overhead over "normal" pino is negligible, but people should really not do multi-stream.

This was also asked in https://github.com/mcollina/pino/issues/80.

jsumners:
I have added some emphasis to the warning the documentation about this feature. Hopefully it will assuage @davidmarkclements's concerns.

Quoted here for reference:

> **Stern warning:** the performance of this feature being dependent on the number
of streams you supply cannot be overstated. This feature is being provided so
that you can switch to `pino` from `Bunyan` and get some immediate improvement,
but it is not meant to be a long term solution. We *strongly* suggest that you
use this feature for only as long as it will take you to overhaul the way
you handle logging in your application.

> To illustrate what we mean, here is a benchmark of `pino` and `Bunyan` using
"multiple" streams to write to a single stream:

> ```
> benchBunyanOne*10000: 875.099ms
> benchMSPinoOne*10000: 224.924ms
> ```

> Now let's look at the same benchmark but increase the number of destination
streams to four:

> ```
> benchBunyanFour*10000: 3288.004ms
> benchMSPinoFour*10000: 722.918ms
> ```

> And, finally, with ten destination streams:

> ```
> benchBunyanTen*10000: 8142.875ms
> benchMSPinoTen*10000: 1992.805ms
> ```

mcollina:
I think having a separate module for the multistream would be the best choice for the project.
I'm ok in helping maintaining that module as well, as I think it's needed by the community.
We should also write a level-splitter module, which implement the same behavior as a "transport".

Both will be separate modules, and we can keep `pino` small and easy to maintain.

Closing this for now.

jsumners:
https://github.com/jsumners/pino-multi-stream

Please proofread it.

mcollina:
@jsumners it's perfect. Can you please delete this branch if you do not need it anymore?

mcollina:
We need to land a PR here that points to pino-multi-stream.

jsumners:
I will add it to the readme once pino-multi-stream is actually published.

PR #107, opened 2016-10-15
Expose Pino constructor so that crafty things can be done with it

jsumners:
As the title says...

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -415,10 +415,17 @@ module.exports.stdSerializers = {
   res: asResValue,
   err: asErrValue
 }
-module.exports.Pino = Pino
 module.exports.pretty = require('./pretty')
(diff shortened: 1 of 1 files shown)


mcollina:
There is only one catch here. It is going to be undocumented, and we might change it within a semver-minor. It's an internal API, exposed to help integration. Can you put this in a comment?

Otherwise LGTM

davidmarkclements:
Can we underscore it to signify it's private/for integration?

module.exports._Pino = Pino

davidmarkclements:
You could even use defineProperty to make sure it's non enumerable

jsumners:
Done.

mcollina:
Great, releasing now!

PR #113, opened 2016-10-21
Correctly escape all the values a user can enter in the message

mcollina:
https://tools.ietf.org/html/rfc7159#section-7

cc @mstade @jsumners @davidmarkclements please check.

Fixes #112.


mstade:
I don't think this tests all characters that must be escaped (or I'm misreading the spec):

> All Unicode characters may be placed within the quotation marks, except for the characters that must be escaped: quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).

I updated my PR with a new test to reflect the above quote: https://github.com/mcollina/pino/pull/112/commits/924ef6ffff3e34c8c81c063acef2e2e7aa612fd6

Like I said though, maybe I'm misunderstanding the thing!

mcollina:
Given that log messages _should not_ be dynamic, I suppose all those unicode chars are unneeded.

Maybe there is an element of safety/performance here.

mstade:
Well, that's what the escaping is for, isn't it? It's pretty easy for things like these to just slip in there (kids log the darndest things!) and the side effect of not covering all bases is that any downstream consumer that relies on JSON parsing to consume records will break as soon as any of these characters pop in there.

mcollina:
@mstade do you write `'\u0014'` in your log messages? If it's a user input, it needs to go either in a format string or as part of the object. Log messages should be fixed, as otherwise they are hard to track.

jsumners:
The tests verify here, and on my ludicrously slow 3.2 GHz Quad-Core Intel Xeon the basic benchmark is:

```
benchBunyan*10000: 2584.189ms
benchWinston*10000: 4775.499ms
benchBole*10000: 613.148ms
benchDebug*10000: 954.171ms
benchLogLevel*10000: 711.739ms
benchPino*10000: 568.414ms
benchPinoExreme*10000: 214.317ms
benchBunyan*10000: 2332.487ms
benchWinston*10000: 4262.136ms
benchBole*10000: 530.107ms
benchDebug*10000: 919.134ms
benchLogLevel*10000: 637.075ms
benchPino*10000: 541.006ms
benchPinoExreme*10000: 220.117ms
```

For reference, my benchmark on the master branch:

```
benchBunyan*10000: 2388.969ms
benchWinston*10000: 4840.217ms
benchBole*10000: 692.859ms
benchDebug*10000: 979.721ms
benchLogLevel*10000: 668.728ms
benchPino*10000: 535.632ms
benchPinoExreme*10000: 227.121ms
benchBunyan*10000: 2337.364ms
benchWinston*10000: 4377.451ms
benchBole*10000: 533.390ms
benchDebug*10000: 831.935ms
benchLogLevel*10000: 599.801ms
benchPino*10000: 587.966ms
benchPinoExreme*10000: 210.844ms
```

mstade:
**I** don't, but isn't it presumptuous to say it won't happen? The spec is pretty clear about it: these things *must* be escaped. Should these characters appear unescaped, then the output is – according to the spec – *not* valid JSON. It may seem pedantic, but I'd err on the side of caution, or at the very least make the "it may not be valid JSON" output path an explicit option.

jsumners:
As for escaping unicode within log messages. I personally can't think of a case where I would ever do that, but I'm also completely satisfied with the ASCII character set for 99.9999% of my writing. I think getting a fix for the most common escapes published is more important than covering unicode immediately.

mcollina:
@jsumners would you mind verifying the performance of this?

jsumners:
@mcollina I did a couple comments above, but here it is from my much better laptop:

**newlines**
```
benchPino*10000: 278.198ms
benchPinoExreme*10000: 123.657ms
benchPino*10000: 246.355ms
benchPinoExreme*10000: 110.902ms
```

**master**
```
benchPino*10000: 241.248ms
benchPinoExreme*10000: 104.916ms
benchPino*10000: 224.621ms
benchPinoExreme*10000: 106.788ms
```

mstade:
@jsumners – the issue is whether or not escaping all control characters (U+0000 - U+001F) is meaningful. If I read the spec correctly, those (and `"` and `\`) are the only characters that *must* be encoded, whereas anything else *may* be encoded. Some of those control characters, like U+0014, may be so rare as to be non-existant, but the point is that *should* one appear and it's *not* encoded then the output is *not* JSON. Pedantic as it may be, I think providing strong guarantees on this point seems prudent.

The space of control characters is also quite small, only 31 characters. I'd be surprised if covering them all (whether through `JSON.stringify` or other means) would incur such significant costs as to be unacceptable.

jsumners:
@mstade I completely agree with you. But I think I should attempt to clarify @mcollina's position.

A `pino` log line has two parts: the information such as the timestamp, level, and version; and the message. A message can either be a simple string, an object to be serialized, or a format string.

1. simple string: this one is the problem. If what we are logging is nothing more than a simple string then `fastRep` is used to escape it for inclusion in `pino`'s `msg` log property. While unusual, it's at least likely that the basic set of escape characters will make it into the simple message. It's far less likely that the unicode characters will make it in.

2. object: if we are logging an object then associated serializers and `fast-safe-stringify` will be used be used to generated the `msg`. In this case *all* characters will be escaped properly

3. format string: this will be processed the same as point 2 -- through `fast-safe-stringify`

So @mcollina's solution for the unicode case is to note the exception in the documentation and suggest that if you need to use unicode in your log message you need to use a format string or an object. By doing this the overhead of processing simple strings for unicode is eliminated. Thus most of the performance of `pino` prior to this patch is retained while conceding that more characters need to be escaped.

mstade:
I think I understand. So in order to opt-in to a (probably) slower but safer path you'd do something like this:

```javascript
myLogger.log('%s', 'my actual log message with interesting characters like \u0014')
```

And if you're ok with a not 100% guarantee that the output will be valid JSON, you can just log strings as you'd do otherwise. *Most* cases *probably* won't have uncommon control characters in them, so you're *probably* fine. This means *any* dynamic log statements can cause surprises, if you aren't careful about the input. So something like this:

```javascript
var char = "\u0014"
myLogger.log("Here's a funny character: " + char)
```

.. would produce invalid JSON, and to be safe should be written as:

```javascript
var char = "\u0014"
myLogger.log("Here's a funny character: %s", char)
```

Maybe a non sequitur, but let's consider ES2015 template strings as well:

```javascript
let char = "\u0014"
myLogger.log(`Here's a funny character: ${char}`)
```

My gut says `sprintf` style format strings are and will continue to be less well known, less popular, and ultimately less useful than template strings. Particularly as ES2015+ continues to gain ground.

I can't help but think that this tradeoff is a footgun for little gain. It took me a while to find why things broke on `\n`, and it'd be a shame if others have to spend such time on other control characters. To me, the performance impact between 100% compliance and ~21% compliance (escaping only 7 of 33 *must* escape characters) – if indeed there is one of any significance – is worth it to guarantee valid output.

Regardless, I'm happy to see *my* use case covered by these changes.

jsumners:
@mstade you understand correctly. I am quite positive a patch that covers all cases, and is equally performant, will not be rejected. But there isn't a performant way to do it that is immediately obvious to me.

mstade:
Wait – equally performant to *what* though, what's the baseline as it were? `master`? This branch?

jsumners:
I would say equal to this branch as it is. But if there is a way to reach pre-`newlines` performance that would be superb. @mcollina and @davidmarkclements may have other opinions.

mstade:
I'm having a hard time getting consistent benchmark results still. I don't see the consistent slowdown in using `JSON.stringify`. I see the same with this branch – it's sometimes slower than master, sometimes faster, most of the time comparable. This is when running Node v6.9.1. What might I be doing wrong?

mcollina:
@jsumners I think I have made it. It should be as fast as master, but with full escaping. Check it out.

jsumners:
Over here:

**master**
```
benchPino*10000: 245.411ms
benchPinoExreme*10000: 133.677ms
benchPino*10000: 242.222ms
benchPinoExreme*10000: 105.855ms
```

**newlines**
```
benchPino*10000: 249.410ms
benchPinoExreme*10000: 105.333ms
benchPino*10000: 218.602ms
benchPinoExreme*10000: 103.165ms
```

That's quite impressive @mcollina. A very nice compromise between both solutions. 👍

mcollina:
what do you think @davidmarkclements? Should we merge and get it out?

jsumners:
Do you think a test that shows a little more complexity is worth adding?:

```js
test('correctly escape `hello \\u001F world \\n \\u0022`', function (t) {
  t.plan(1)
  var instance = pino({
    name: 'hello'
  }, sink(function (chunk, enc, cb) {
    delete chunk.time
    t.deepEqual(chunk, {
      pid: pid,
      hostname: hostname,
      level: 60,
      name: 'hello',
      msg: 'hello \u001F world \n \u0022',
      v: 1
    })
    cb()
  }))

  instance.fatal('hello \u001F world \n \u0022')
})
```

mstade:
Looking good @mcollina! I'm unable to break this, but agree with @jsumners that a fuzzier test is probably worth adding. FWIW, I've used [js-combinatorics](https://www.npmjs.com/package/js-combinatorics) in the past to generate large fuzzy data sets for these kinds of tests. Might be a fun little side project if you're into that sort of thing. :o)

README.md
```diff
@@ -408,6 +423,8 @@ object, all its properties will be included in the JSON line.
 If more args follows `msg`, these will be used to format `msg` using
 [`util.format`](https://nodejs.org/api/util.html#util_util_format_format)
 
+For best performance _and_ safety, use a fixed `msg`, and place all
+varying data either in the object or as a format.
```
mstade: Is the safety notice still relevant since you're now doing full escaping?

pino.js
```diff
@@ -201,14 +201,21 @@ Object.defineProperty(
...
-  for (var i = 0; i < l; i++) {
-    if (str[i] === '"') {
-      result += str.slice(last, i) + '\\"'
+  var point = 255
+  for (var i = 0; i < l && point >= 32; i++) {
+    point = str.charCodeAt(i)
+    if (point === 34 || point === 92) {
+      result += str.slice(last, i) + '\\' + str[i]
       last = i + 1
     }
```
mstade: Can the return conditional be moved here, to exit the moment a code point below 32 is found? Something like:

```javascript
} else if (point < 32) {
  return JSON.stringify(str)
}
```

pino.js
```diff
@@ -217,7 +224,7 @@ function fastRep (s) {
   } else {
     result += str.slice(last)
   }
-  return result
+  return point < 32 ? JSON.stringify(str) : '"' + result + '"'
```
mstade: If the conditional is moved to break the loop whenever a code point below 32 is found, this just becomes:

```javascript
return '"' + result + '"'
```

pino.js
mstade: Nevermind, I missed the condition in the for loop!! :o)

pino.js
mcollina: You are not correct. If a code point below 32 is found, it becomes `JSON.stringify(str)`, which will yield the correct result.

The unit tests to be compliant with the spec are passing.

pino.js
mcollina: that's what I did in the beginning, but it turned out to be slower, because of the two return statements.

pino.js
mstade: I meant if you moved the return statement into the loop, this return statement would be simplified. I just didn't see the early break in the for loop condition – all just a misunderstanding, my bad!

README.md
mcollina: thanks for noticing this!!!!

davidmarkclements:
wow - outstanding work y'all

mstade:
Thanks @mcollina – much obliged!

PR #114, opened 2016-10-24
minimal browser implementation

davidmarkclements:
currently pino isn't working in the browser

this simply uses the corresponding `console` methods, and ensures a light payload

cc @mcollina

### DIFF
--- a/.zuul.yml
+++ b/.zuul.yml
@@ -1,0 +1,4 @@
+name: pino
+ui: tape
+browserify:
+  - transform: fresh-require/transform
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ It also includes a shell utility to pretty-print its log files.
 * [How do I rotate log files?](#rotate)
 * [How do I redact sensitive information?](#redact)
 * [How to use Transports with Pino](#transports)
+* [Pino in the browser](#browser)
 * [Caveats](#caveats)
 * [Team](#team)
 * [Acknowledgements](#acknowledgements)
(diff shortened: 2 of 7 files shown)


davidmarkclements:
currently pino isn't working in the browser

this simply uses the corresponding `console` methods, and ensures a light payload

cc @mcollina

jsumners:
If I'm not mistaken, this would remove the need for things like https://github.com/mcollina/pino/blob/eaaa41ff53c7ca3283fe273fa3fe8658fe286313/pino.js#L144

mcollina:
We need a bit of child logging here to be complete.
And also a test for the browser, with tap and zuul.

davidmarkclements:
@jsumners - true
@mcollina - right

davidmarkclements:
ok child loggers are done, and I've stubbed every public part of the pino API

except asJson

Any objects to making asJson private - as far as I'm aware we only use it internally (plus using it externally is actually pretty difficult, it relies on internal pino state)

davidmarkclements:
It's written with only ES3 so should work in all browsers but regarding testing...
What if we just unit test in node - my thought here is that if we test in chrome it's essentially no different to testing in node - the only way to be sure regarding browsers is to run the browser tests in every browser. 

Also - browser pino won't support dynamically setting level (since Object.defineProperty is ES5)

For browser logging, it would actually be more useful to read a global (PINO_LEVEL) anyway - which we could also discuss

@mcollina @jsumners - thoughts?

davidmarkclements:
ok tests added - for me this is ready for first release ( unless there's a strong consensus for browser tests )

jsumners:
I don't think the browser implementation needs to have 1:1 feature parity with the Node version. As long as the interface remains the same and logs get written to the console I would be happy.

davidmarkclements:
for sure it doesn't - there's no wrapping logs with timestamps etc. 

but the reason I've stubbed all the things is to avoid exceptions in isomorphic cases

jsumners:
Also, if you look at http://caniuse.com/#search=console I think supporting the minimal case where everything is 100% green is acceptable. Which means ES5 is also possible if it would make implementation easier.

davidmarkclements:
I'm happy with ES3 because it means we're totally safe - unless there's a strong case for dynamic level setting in the browser

jsumners:
LGTM

mcollina:
I think we should add support for `level = ` with  `defineProperty`. I do not care about anything < IE 10 anyway.

http://caniuse.com/#search=defineProperty

I _do_ care for browser tests, especially if we can integrate them with `zuul`. 
However, I can do them later on myself, if we really need them: levelgraph has some good examples, https://github.com/mcollina/levelgraph.

mcollina:
I think we can add `defineProperty` later on. No worries now, let's get this merged and released first.

davidmarkclements:
I'll convert browser tests to tape though - for later integration

davidmarkclements:
ok - tests converted to tape, zuul added - works in browser 100% coverage

PR #119, opened 2016-10-27
String escape performances

delvedor:
I've made some experiments with the string stringify to find the fast way to manage unicode characters.
I wrote a benchmark that tests `JSON.stringify` as is, `escape` and `asStr` (the function that I wrote).
Comparing my results with the Matteo's one, we found some differences.

Here is the result of mine benchmarks:

``` bash
short 33
long 541
freakingLong 541000

long-stringify: 2.625ms
long-asString: 2.460ms
long-asStr: 2.166ms

short-stringify: 0.006ms
short-asString: 0.050ms
short-asStr: 0.010ms

JSON.stringify - freaking long x 524 ops/sec ±1.07% (88 runs sampled)
asStringSmall - freaking long x 506 ops/sec ±1.05% (84 runs sampled)
asStr - freaking long x 509 ops/sec ±1.03% (89 runs sampled)

JSON.stringify - long x 433,415 ops/sec ±1.01% (91 runs sampled)
asStringSmall - long x 437,293 ops/sec ±1.03% (86 runs sampled)
asStr - long x 441,310 ops/sec ±1.02% (87 runs sampled)

JSON.stringify - short x 3,546,163 ops/sec ±0.95% (90 runs sampled)
asStringSmall - short x 2,312,364 ops/sec ±0.85% (91 runs sampled)
asStr - short x 1,914,376 ops/sec ±0.90% (91 runs sampled)
```

Here there are the gist of the two [escape functions](https://gist.github.com/delvedor/7d747487dc18ad3e52feb555969ee97a) and the [benchmark](https://gist.github.com/delvedor/23111168392d83a4f994a26bdf7822b6).
As you can see `JSON.stringify` is faster in the _freaking long_ and _short_ strings, while is slower in the _long_ case.
If you run the pino bench with `asStr` pino is slower, but in the raw bench `asStr` is faster than `escape`.
Am I missing something?

`asStr` code:

``` js
function asStr (str) {
  var point = 255
  for (var i = 0, l = str.length; i < l && point > 31; i++) {
    point = str.charCodeAt(i)
    if (point === 34 || point === 92) {
      str = str.slice(0, i) + '\\' + str.slice(i++)
      l++
    }
  }
  return point < 32 ? JSON.stringify(str) : '"' + str + '"'
}
```

Related: 
https://github.com/mcollina/fast-json-stringify/pull/14

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -205,6 +205,17 @@ Object.defineProperty(
 // 34 and 92 happens all the time, so we
 // have a fast case for them
 function escape (s) {
+  var point = 255
+  for (var i = 0, l = s.length; i < l && point > 31; i++) {
+    point = s.charCodeAt(i)
+    if (point === 34 || point === 92) {
+      s = s.slice(0, i) + '\\' + s.slice(i++)
+      l++
+    }
+  }
+  return point < 32 ? JSON.stringify(s) : '"' + s + '"'
+}
+/* function escape (s) {
   var str = s.toString()
   var result = ''
   var last = 0
@@ -223,7 +234,7 @@ function escape (s) {
     result += str.slice(last)
   }
   return point < 32 ? JSON.stringify(str) : '"' + result + '"'
-}
+} */
 
 Pino.prototype.asJson = function asJson (obj, msg, num) {
   if (!msg && obj instanceof Error) {


mcollina:
I am a bit puzzled with these results.
I will need some time to digest what is the best course of action for this.

davidmarkclements:
Hey @delvedor can you show the benchmarking script, I'd like to see the strings and how they're constructed

delvedor:
@davidmarkclements I've added the Gist link above, here it is :)
https://gist.github.com/delvedor/23111168392d83a4f994a26bdf7822b6

davidmarkclements:
Thanks :)

What if the ratio of Unicode chars to ascii chars has an impact - in the short str is there a higher Unicode density? What if we change the position of the Unicode chars (say shift them all right by one by adding a char to the beginning, and popping off the last string then use flatstr to avoid concat having impact) 

I'd also like to know if a particular range of Unicode chars has a higher impact, would a higher plane take longer to parse? Are all the Unicode chars close together as far as code points go? 

Finally I'd like to see a test on lone surrogates, what effect do they have on perf and does the custom function work with lone surrogates

delvedor:
@davidmarkclements I've made this script to generate the string with the unicode chars frequency. Maybe can help! :)
https://gist.github.com/delvedor/fc3e29224f1410dcd584c00811c18342

davidmarkclements:
oh wait - these are *control* characters not unicode characters

everything I said about code points/code planes and lone surrogates doesn't apply

davidmarkclements:
closed via #129

PR #124, opened 2016-11-08
Add support for custom log levels

jsumners:
This PR defines a new static method that allows a user to create their own log level. Example:

```js
var pino = require('pino')
pino.addLevel('myLevel', 35)
var log = pino({level: 'myLevel'})
log.myLevel('a message')
log.warn('another message') // gets logged as well
```

This PR also ties level changes to the `pino.levels` property. Otherwise, adding a new level would result in pino instances not knowing about the new level.

## Benchmarks

**addlevel**
```
benchPino*10000: 231.730ms
benchPinoExreme*10000: 105.404ms
benchPino*10000: 225.207ms
benchPinoExreme*10000: 99.640ms
```

**master**
```
benchPino*10000: 231.722ms
benchPinoExreme*10000: 108.249ms
benchPino*10000: 221.465ms
benchPinoExreme*10000: 103.063ms
```


mcollina:
This is a nice feature that makes sense with the log file format, I wondered a couple of times how to log with level 35, for example. The only thing I am not happy with is the tweaking of the module-global, it can result in brittle code, and hard to debug errors.

jsumners:
I don't see the change as being to the module global. The module global is the `levels` and `nums` variables that are copied onto the `pino` function. In my opinion, it was wrong to directly reference `levels` and `nums` from the `Pino.prototype` methods (yes, I know I was last to re-factor them). But, to directly address the concern about brittleness, the `addLevel` function explicitly prevents the user from changing the standard levels. Thus, the shipped state of pino cannot be changed via this method other than to add to it.

mcollina:
Because of how node works there can be different versions of the same module around. Tweaking the global values might cause problems if a module is using a custom level, as it might depend on a different version of pino than the parent one.

jsumners:
Good point. I'll think about how to solve that problem, but I'm open to ideas.

davidmarkclements:
I can see three options

1. add something to the global scope, or to a global object (like `process`) - this is generally ill-advised
2. a custom build process
3. use env vars to set levels - this breaks the top down config structure (too magical) 
4. a package.json field (roughly the same drawbacks as 3, but additionally violates principle of least surprise - too esoteric) 

our prevailing philosophy for pino has been, env vars should only have an effect at the app level, and then be passed down and through to a pino instance

We could simply be explicit - by passing levels to pino function, we can make it clear that levels only apply on an instance by instance level - and then note that a good strategy for doing this across a process (and indeed across many processes) is setting env vars which are then passed to the instantiation functions

jsumners:
I'm having a difficult time understanding your last paragraph @davidmarkclements, but I know that I don't like any of the 1 - 4 options.

The solution may be to make the method an instance method. But the project I'm working on expects the operation to be a static. And if the level doesn't exist prior to an instance being created, and the level is set via the constructor options, then it won't work.

jsumners:
The `addLevel` method is now an instance method instead of a static method. Also, the `levels` property gets assigned to the `Pino` instance; once everything is initialized, the global `levels` and `nums` are not referenced by `Pino` instances. This means adding levels does not interfere with the global state of the module.

**addLevel**
```
Running BASIC benchmark

benchBunyan*10000: 923.717ms
benchWinston*10000: 1853.523ms
benchBole*10000: 267.104ms
benchDebug*10000: 418.852ms
benchLogLevel*10000: 275.071ms
benchPino*10000: 228.344ms
benchPinoExreme*10000: 110.595ms
benchBunyan*10000: 870.290ms
benchWinston*10000: 1754.976ms
benchBole*10000: 242.574ms
benchDebug*10000: 393.464ms
benchLogLevel*10000: 256.559ms
benchPino*10000: 226.061ms
benchPinoExreme*10000: 100.286ms

==========
basic averages
Bunyan average: 897.0035
Winston average: 1804.2495
Bole average: 254.839
Debug average: 406.158
LogLevel average: 265.81500000000005
Pino average: 227.2025
PinoExreme average: 105.4405
==========

Running CHILD benchmark

benchBunyanChild*10000: 1071.734ms
benchBoleChild*10000: 294.188ms
benchPinoChild*10000: 268.986ms
benchPinoExtremeChild*10000: 131.967ms
benchBunyanChild*10000: 1065.870ms
benchBoleChild*10000: 261.570ms
benchPinoChild*10000: 244.487ms
benchPinoExtremeChild*10000: 118.525ms

==========
child averages
BunyanChild average: 1068.802
BoleChild average: 277.879
PinoChild average: 256.7365
PinoExtremeChild average: 125.24600000000001
==========
```

**master**
```
Running BASIC benchmark

benchBunyan*10000: 940.189ms
benchWinston*10000: 1859.667ms
benchBole*10000: 258.799ms
benchDebug*10000: 463.724ms
benchLogLevel*10000: 272.053ms
benchPino*10000: 227.422ms
benchPinoExreme*10000: 103.318ms
benchBunyan*10000: 882.152ms
benchWinston*10000: 1773.495ms
benchBole*10000: 232.979ms
benchDebug*10000: 401.959ms
benchLogLevel*10000: 258.316ms
benchPino*10000: 225.177ms
benchPinoExreme*10000: 99.087ms

==========
basic averages
Bunyan average: 911.1705
Winston average: 1816.581
Bole average: 245.889
Debug average: 432.8415
LogLevel average: 265.18449999999996
Pino average: 226.2995
PinoExreme average: 101.2025
==========

Running CHILD benchmark

benchBunyanChild*10000: 1144.688ms
benchBoleChild*10000: 307.021ms
benchPinoChild*10000: 280.006ms
benchPinoExtremeChild*10000: 128.241ms
benchBunyanChild*10000: 1089.686ms
benchBoleChild*10000: 272.223ms
benchPinoChild*10000: 251.781ms
benchPinoExtremeChild*10000: 122.741ms

==========
child averages
BunyanChild average: 1117.187
BoleChild average: 289.622
PinoChild average: 265.8935
PinoExtremeChild average: 125.49100000000001
==========
```

jsumners:
@davidmarkclements any opinion on inclusion?

README.md
```diff
@@ -211,6 +211,7 @@ INFO [2016-03-09T15:27:09.339Z] (14139 on MacBook-Pro-3.home): hello world
   * <a href="#resSerializer"><code>pino.stdSerializers.<b>res</b></code></a>
   * <a href="#errSerializer"><code>pino.stdSerializers.<b>err</b></code></a>
   * <a href="#pretty"><code>pino.<b>pretty()</b></code></a>
+  * <a href="#addLevel"><code>pino.<b>addLevel(name, lvl)</b></code></a>
```
mcollina: This still references the global way of doing things, it needs updating.

davidmarkclements:
yeah @jsumners once readme fix is in, let's totally get this merged great job James!

jsumners:
Yay for doing things late at night. Readme is fixed.

README.md
```diff
@@ -415,6 +417,34 @@ If more args follows `msg`, these will be used to format `msg` using
 Flushes the content of the buffer in [extreme mode](#extreme). It has no effect if
 extreme mode is not enabled.
 
+<a name="addLevel"></a>
+### pino.addLevel(name, lvl)
```
mcollina: this should probably be `logger.addLevel`.

jsumners: *face palm*

mcollina:
LGTM!

Do we want to release this as part of the v2 line, and wait for v3, or release it as part of v3?

jsumners:
I believe both this and `native-assign` are ready to go. I don't see any need to merge this in, release a v2, merge in `native-assign`, and then release a v3 all in short succession.

PR #125, opened 2016-11-11
Refactor internal Pino to use an options object

jsumners:
In order to solve #124 I intend to associate the levels with the actual `Pino` instance returned by `pino`. To do that, I will need to add another parameter to the `Pino` constructor. I think the constructor already has way too many parameters, so I refactored it to accept only two such that the signature is `Pino(opts, stream)`.

Running the basic benchmark on the master branch and this branch 3 times each and averaging the results, I get:

+ refactored bench average: 235.316333
+ master bench average: 237.611667

Thus, this refactor is equally performant.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -13,6 +13,17 @@ var extend = require('object.assign').getPolyfill()
 
 var LOG_VERSION = 1
 
+var defaultOptions = {
+  safe: true,
+  name: undefined,
+  serializers: {},
+  timestamp: true,
+  slowtime: false,
+  extreme: false,
+  level: 'info',
+  enabled: true
+}
+
 var levels = {
   fatal: 60,
   error: 50,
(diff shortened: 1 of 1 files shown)


pino.js
```diff
@@ -52,48 +63,48 @@ function streamIsBlockable (s) {
...
 
-  if (opts.enabled === false) {
-    level = 'silent'
+  if (iopts.enabled === false) {
+    iopts.level = 'silent'
   }
 
-  var logger = new Pino(level, stream, serializers, stringify, end, name, timestamp, slowtime, '', cache, formatOpts)
-  if (cache) {
+  var logger = new Pino(iopts, istream)
```
davidmarkclements: nice, much cleaner

pino.js
```diff
@@ -52,48 +63,48 @@ function streamIsBlockable (s) {
...
-      if (cache.buf) {
+      var buf = iopts.cache.buf
+      if (buf) {
         // We need to block the process exit long enough to flush the buffer
         // to the destination stream. We do that by forcing a synchronous
         // write directly to the stream's file descriptor.
-        var fd = (stream.fd) ? stream.fd : stream._handle.fd
-        require('fs').writeSync(fd, cache.buf)
+        var fd = (istream.fd) ? istream.fd : istream._handle.fd
+        require('fs').writeSync(fd, buf)
```
davidmarkclements: I don't know where this came from the first time, but we shouldn't be requiring inline as a matter of principle, would you mind moving this out

davidmarkclements:
in this particular step, from a user perspective, does this change or add to the API at all - or is it purely an internal stepping stone to the next thing - if the former can you add something to readme about usage change/addition

regarding perf, just to be absolutely sure can we check all benchmarks, in particular the child loggers

pino.js
jsumners: It's possible that I did it when fixing the "could be a non-sync stream" bug. I'll fix it.

jsumners:
I'm running benchmarks right now; it's going to take a while. But my preliminary check is that everything remains the same.

This PR does not change anything the user needs to be aware of. Everything is internal to the module.

jsumners:
Attached are my logs for all of the benchmarks. I don't see any differences that aren't accounted for by other processes taking time away from the work.

[master.bench.txt](https://github.com/pinojs/pino/files/586016/master.bench.txt)
[refactor.bench.txt](https://github.com/pinojs/pino/files/586015/refactor.bench.txt)

PR #129, opened 2016-11-15
deep interpolation perf + double escape bug fix

davidmarkclements:
deep interpolation perf (in multiarg benchmark) is the slowest performing

```
benchDebugInterpolateDeep*10000: 456.411ms
benchBunyanInterpolateDeep*10000: 6811.007ms
benchWinstonInterpolateDeep*10000: 8027.710ms
benchBoleInterpolateDeep*10000: 4930.107ms
benchLogLevelInterpolateDeep*10000: 3078.195ms
benchPinoInterpolateDeep*10000: 18080.267ms
benchPinoUnsafeInterpolateDeep*10000: 17414.391ms
benchPinoExtremeInterpolateDeep*10000: 18669.065ms
```

the deep object in multiarg is circular, when we make this genuinely deep, the number get worse: 

```
benchBunyanInterpolateDeep*10000: 11986.442ms
benchWinstonInterpolateDeep*10000: 12761.023ms
benchBoleInterpolateDeep*10000: 8532.904ms
benchLogLevelInterpolateDeep*10000: 5599.804ms
benchPinoInterpolateDeep*10000: 34322.165ms
benchPinoUnsafeInterpolateDeep*10000: 31390.403ms
```

with this commit pino becomes the fastest for this case

deep-interp-perf branch:

```
benchBunyanInterpolateDeep*10000: 10815.396ms
benchWinstonInterpolateDeep*10000: 12499.987ms
benchBoleInterpolateDeep*10000: 8399.518ms
benchPinoInterpolateDeep*10000: 8380.004ms
benchPinoUnsafeInterpolateDeep*10000: 8171.609ms
benchPinoExtremeInterpolateDeep*10000: 8138.440ms
```

This also replaces quick-format with quick-format-unescaped and removes the escape function (`JSON.stringify` on a string is faster in every case - v6)

Also, I've remove debug and loglevel from interpolation benchmarks - they don't interpolate

We have to count interpolators because of a nasty perf regression in util.format where if the args exceed interpolators speed drops drastically  (which quick-format handles, but util.format is much faster at deep obj interpolation)

The thought "why don't we just use util.format for everything" may occur - and it's because the only case where util.format JSON.stringifies an object is the interpolation scenario, ever where else it presents object like `{foo: 'bar'}` whereas we (via quick-format) return `{"foo": "bar"}` which is right and good for JSON logging

also we use a template string on line 412, as it's consistently faster to embed util format return in a template string - see https://github.com/nodejs/node/blob/master/lib/console.js#L39-L44


davidmarkclements:
Closes #119 - nomlonger relevant

davidmarkclements:
just added a long string benchmark - also have to always pass {lowres: true} to quick-format-unescaped so that it behaves the same as util.format

Issue #131, opened 2016-11-23
cannot read property fd of undefined


ronag:
```
foo/node_modules/pino/pino.js:123
        var fd = (istream.fd) ? istream.fd : istream._handle.fd
                                                            ^

TypeError: Cannot read property 'fd' of undefined
    at foo/node_modules/pino/pino.js:123:61
    at f (foo/node_modules/once/once.js:25:25)
    at process.onExit.insertCode (foo/node_modules/pino/pino.js:444:7)
    at emitOne (events.js:96:13)
    at process.emit (events.js:188:7)
    at processEmit [as emit] (foo/node_modules/fsevents/node_modules/signal-exit/index.js:146:32)
    at process._fatalException (bootstrap_node.js:284:26)
```

```javascript
var pino = require('pino')
var pretty = pino.pretty()
pretty.pipe(process.stdout)
var log = pino({
  name: 'app',
  safe: true
}, pretty)

log.child({ widget: 'foo' }).info('hello')
log.child({ widget: 'bar' }).warn('hello 2')
```

Node v7

jsumners:
This should only be possible if you are using "extreme" mode. Indeed, using your test script, I do not see the issue:

```bash
% node test.js                                                                                                                                     
[2016-11-23T13:30:22.891Z] INFO (app/59494 on sumners-osx.int.clayton.edu): hello
    widget: "foo"
[2016-11-23T13:30:22.897Z] WARN (app/59494 on sumners-osx.int.clayton.edu): hello 2
    widget: "bar"

[jsumners@sumners-osx] /tmp/23
% node --version                                                                                                                                   
v7.2.0
```

davidmarkclements:
sidepoint - `safe` is true by default - @ronag could you give us exact node version, also what OS are you on?

ronag:
```javascript
const pretty = pino.pretty()
pretty.pipe(process.stdout)
const logger = pino({
  name: 'nxt-file',
  extreme: true
}, pretty)
```

OSX 10.12, v7.0.0

mcollina:
which version of pino? We should have adressed that in the past. You can't use extreme mode with anything that is not a socket or a file stream.

mcollina:
(you should see a better error than this one)

ronag:
3.0.1

Could we have a wrapper around the destination if it does not support extreme mode?

jsumners:
[...]

Issue #92, opened 2016-09-23
pino issues `process.exit` when it isn't the controlling process


jsumners:
When pino is in "extreme" mode it has to register handlers for `process.on('exit')` (https://github.com/mcollina/pino/blob/master/pino.js#L390). These handlers are necessary so that pino can flush the extreme mode cache prior to the process exiting, or else there is the potential for data loss. After flushing the cache, pino then checks to see if its events are in the event queue (https://github.com/mcollina/pino/blob/master/pino.js#L98). If so, pino issues a `process.exit()`.

It is my opinion that a logger should not be invoking `process.exit`. That should be left up to the application that included pino. However, once event handlers are registered on `process` the process doesn't respond to signals like `SIGTERM` normally; at this point it is up to the application to deal with those signals.

I think there are a couple possible solutions:
1. instead of issuing `process.exit` pino can issue its own "exit" event so that the actual application can do what it needs to do for shutdown
2. pino can accept a callback during initialization that will be invoked after the cache flushing

In either case, this moves the responsibility of terminating the process to the actual application. And it removes the surprise `process.exit` within pino.

Thoughts, @mcollina @davidmarkclements ?

davidmarkclements:
We did discuss this very thing (point 2) early on with a view to reviewing current impl. later

I think 2 is best, extreme mode is for advanced users, we just need to explain in docs

@mcollina what do you think?

mcollina:
We need to change the current behavior, as accessing _events will become unsupported in the near future (there is a core issue on this).

IMHO the current behavior is ok. If extreme mode is enabled, the fact that pino listens to some global event should be transparent: if there is no one listening, we should 

The other option is letting the user decide, and not opt them in. We already expose flush(), it is just a refactoring to make it sync, and document a way on how to call flush() correctly. Or maybe offer an handy method that does what we do now.

I disagree on offering the user a callback on creation.

jsumners:
I have been working on the callback method and it is proving very difficult to test. I agree with @mcollina that it is not an optimal solution.

I like the idea of requiring `flush()` to be called for a proper shutdown in extreme mode.

mcollina:
I think we may want an extreme mode and a flushBeforeExit thing, that enables the current behavior.

davidmarkclements:
Like `pino({flushBeforeExit: true})` ?- but who's not going to want that by default? 

It's tricky because any of the currently discussed options require explanation,

[...]

PR #132, opened 2016-11-24
Add a ready event for extreme mode

jsumners:
This is to make #131 less of an issue. We can't really fail gracefully at that point due to the asynchronous nature of waiting for file descriptors to settle. So the next best thing is to give the extreme mode user a `ready` event so that they know when the logger is fully vetted and ready to go. But they *really* need to listen for the `error` event to know if the stream they have supplied is valid.


davidmarkclements:
I'm... not sure about this

1. if they really need to listen to error, then isn't ready a red herring
2. the *only* reason we need ready is because of a hack, I wonder if there's another way

add to austin discussion?

davidmarkclements:
discussed - alt solution

jsumners:
The alt solution is covered by -- https://github.com/pinojs/pino/issues/92#issuecomment-264035843

PR #134, opened 2016-11-29
Add new err-serializer that serializes all error properties

satazor:
This replaces the core serializer with the one I made in https://github.com/IndigoUnited/pino-err-serializer.

Before merging, I need to port the tests to this repo. Any guidance to were these should be added?

There's https://github.com/pinojs/pino/blob/master/test/serializers.test.js but it doesn't actually test the core serializers.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -364,11 +364,17 @@ function asResValue (res) {
 }
 
 function asErrValue (err) {
-  return {
+  var obj = {
     type: err.constructor.name,
     message: err.message,
     stack: err.stack
   }
+  for (var key in err) {
+    if (obj[key] === undefined) {
+      obj[key] = err[key]
+    }
+  }
+  return obj
 }
 
 function countInterp (s, i) {


mcollina:
Here it is https://github.com/pinojs/pino/blob/master/test/basic.test.js#L70-L90.

We can probably move that in its own test file.

Be careful that we need to handle also this case https://github.com/pinojs/pino/blob/master/pino.js#L242-L244.

satazor:
@mcollina thank for the hint, still haven't got time to port the tests. I'm currently busy at work, but will try to find some time this weekend. Alternatively, if someone have some time to help I would be much grateful.

mcollina:
Take your time ;). We are at Node Interactive and we won't have much time either!

satazor:
@mcollina I'm wondering if `serializers.test.js` could be renamed to `serializers-behavior.test.js` (I tought initially that this would be the place to add the err-serializer tests) and if I could add a `err-serializer.test.js`.

mcollina:
Just add a err-serializer.test.js. It's a complex enough issue to have its own test file.
Don't worry too much on where the tests are, they can be moved easily.

mcollina:
hey @satazor, any updates on this?

satazor:
Hey @mcollina, unfortunately no. I'm starting my own company and it drains all my time. This is not forgotten but I understand that you want to move on with this. What do you suggest?

mcollina:
@davidmarkclements would you have some time to work on this? I'm currently blocked for most of this week.

mcollina:
Released as v3.1.0.

Issue #136, opened 2016-11-29
employ an .npmignore file


osher:
the published package is pacaked with all the sources in it, including tests, examples, demo.png and all.

published package should be slim and elegant.

proposed file:
`.npmignore`
```
benchmarks
test
example
CONTRIBUTING.md
*.png
*.yml
*.txt
```

jsumners:
I agree. I just wish adding a npmignore file didn't cause npm to then ignore the gitignore file and make keeping them synced a problem.

osher:
rich mans problem.

osher:
mm. thats a nice idea for a CLI tool...
maybe I'll work on it later

davidmarkclements:
@mcollina and I have some pretty strong views on keeping `test`, and to some extent example, folder in published module - possibly by extensions benchmarks too

*but* if we were to to do this, we might consider whitelisting instead via the `files` array in `package.json`

jsumners:
For our code size, I'm really not concerned either way. I do like a production install only having what is absolutely necessary. But during development it can be useful to have these things delivered with the module so that you can easily peruse it if needed. So for me, ^shrug^.

mcollina:
IMHO we should do this to remove the `*.png` files at a bare minimum. I really think we should retain the tests, examples and CONTRIBUTING.md. I actually prefer the `files` approach.

osher:
I never thought this will become such a discussion.

https://github.com/pinojs/pino/pull/137

osher:
@jsumners - when a package is installed as a dependency, there is no distinction between dependency on production and dependency on a developer machine.
The idea is that develoeprs that work on a package work with GIT and get everything from there.
All the rest - production, stage, QA, and developers that just use the package - get it from NPM.

To sum up:
- publish to NPM a slim package 
- keep in GIT all the dev-utils, docs, examples and design (excluded from published pacakges)

jsumners:
I understand that there is no distinction. But I *do* navigate to modules within a project's `node_modules` to view the docs, code, and tests (since most people think they are documentation). So there is a reason to keep such things in the package. At the same time, I understand the desire to *not* have those things delivered when installing as a dependency for production. Hence my stance of accepting either solution.

mcollina:
I think we are all in for a `files` section, and removing the not needed things: png, yml, benchmarks.

jsumners:
[...]

PR #137, opened 2016-11-29
add .npmignore file

osher:
refers to #136

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,13 @@
   "bin": {
     "pino": "./pretty.js"
   },
+  "files" : [
+    "pino.js", 
+    "browser.js",
+    "pretty.js",
+    "LICENEE",
+    "usage.txt"
+  ],
   "scripts": {
     "browser-test": "zuul tape test/browser.test.js --local",
     "test": "standard && tap --no-cov test/*test.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -7,6 +7,13 @@
   "bin": {
     "pino": "./pretty.js"
   },
+  "files" : [
+    "pino.js", 
+    "browser.js",
+    "pretty.js",
+    "LICENEE",
```
mcollina: this should be LICENSE

osher: that's against the standard
besides - tests are 40K and that doubles the size of the package.
FYI.

osher: I adhered.
let me know if you change your mind again, I'll be happy to change it.

osher: oups. almost missed. I had a typo and it still included LISENSE?
So we don't need to mention it. npm granbs it anyway.

osher:
Here's what we got so far:
image

mcollina:
this brings down the size of tgz from 460K to 28K, pretty awesome. Thanks!

osher:
peasure working with you.
You're more responsive than I am!

davidmarkclements:
Thanks @osher!

PR #147, opened 2016-12-13
Added pino-tee to the README.

### LINKED ISSUE #145: Passing multiple streams into pino
Hi,

Is it possible to pass multiple streams into `pino`? Something like what we do in `bunyan` lib?

```javascript
var bunyan = require('bunyan');
var log = bunyan.createLogger({
    name: <string>,                     // Required
    level: <level name or number>,      // Optional, see "Levels" section
    stream: <node.js stream>,           // Optional, see "Streams" section
    streams: [<bunyan streams>, ...],   // Optional, see "Streams" section
    serializers: <serializers mapping>, // Optional, see "Serializers" section
    src: <boolean>,                     // Optional, see "src" section

    // Any other fields are added to all log records as is.
    foo: 'bar',
    ...
});
```

mcollina:
Fixes #145.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ It also includes a shell utility to pretty-print its log files.
 * [How to use Pino with Koa](#koa)
 * [How to use Pino with `debug`](#debug)
 * [How do I rotate log files?](#rotate)
+* [How to save to multiple files?](#multiple)
 * [How do I redact sensitive information?](#redact)
 * [How to use Transports with Pino](#transports)
 * [Pino in the browser](#browser)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -838,6 +839,26 @@ We would rotate our log files with logrotate, by adding the following to `/etc/l
...
+
+```bash
+npm i pino-tee -g
+```
+
+The following writes the log output of `app.js` to `./all-logs`, while
+writing only warnings and errors to `./warn-log:
+
+```bash
+node app.js || pino-tee warn ./warn-logs > ./all-logs
```
jsumners: This should be a single `|`.

mcollina:
@jsumners updated.

PR #148, opened 2016-12-14
Err serializer tests

colmose:
I've ported across the tests for err-serializer.

Not sure if you'd prefer it in a separate branch or within the other PR. I can work in a fork of the PR if that makes more sense. 

I had to make a change to pino.js to remove the else in the if(obj.stack) check on line 244, not sure if that was the right approach?


colmose:
I'll request the merge to the err-serializer PR.

PR #149, opened 2016-12-14
Add tests for err-serializer

colmose:
I've ported across the tests for err-serializer.

Not sure if you'd prefer it in a separate branch or within the other PR. I can work in a fork of the PR if that makes more sense.

I had to make a change to pino.js to remove the else in the if(obj.stack) check on line 244, not sure if that was the right approach?

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -157,7 +157,6 @@ function Pino (opts, stream) {
     if (levelIsStandard === false && valIsStandard === false) this.addLevel(opts.level, opts.levelVal)
   }
   this._setLevel(opts.level)
-
   this._baseLog = flatstr(baseLog +
     (this.name === undefined ? '' : '"name":' + this.stringify(this.name) + ','))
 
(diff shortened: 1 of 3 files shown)


jsumners:
I don't think removing the `else` is completely correct. The first condition was for avoiding the loop in the second condition. I think a better solution would be to track if an error serializer was added and then skip the first condition if so.

pino.js
```diff
@@ -241,14 +240,13 @@ Pino.prototype.asJson = function asJson (obj, msg, num) {
...
-          if (value !== undefined) {
-            data += ',"' + key + '":' + value
-          }
+    }
+    for (var key in obj) {
+      value = obj[key]
+      if (obj.hasOwnProperty(key) && value !== undefined) {
+        value = this.stringify(this.serializers[key] ? this.serializers[key](value) : value)
+        if (value !== undefined) {
+          data += ',"' + key + '":' + value
```
mcollina: this is not ok. we are probably including also the properties for `obj.stack` and `obj.type`.

mcollina:
Can you please also include the commit on #134?

colmose:
@mcollina sort out the if/else first or move it to #134 as it is now and fix in there?

mcollina:
same thing to me, as you prefer.

colmose:
Hey @mcollina @davidmarkclements I won't get a chance to work on this until at least the middle of next week. Happy to do it then but if someone else wants to take it on, happy for the start made here to be used.

davidmarkclements:
hey @colmose still happy with your time line - it's mostly perf improvement so whenever you can :grin:

jsumners:
Any updates on this one?

mcollina:
Released as v3.1.0.

colmose:
Ah, apologies I didn't get to this over the hols. 👍

Issue #154, opened 2016-12-30
pino-pretty crashes


mzehrer:
This happens while logging some random output from a shell command:

/usr/lib/node_modules/pino/pretty.js:50
  return line.hasOwnProperty('hostname') && line.hasOwnProperty('pid') && (line.hasOwnProperty('v') && line.v === 1)
             ^

TypeError: Cannot read property 'hasOwnProperty' of null
    at isPinoLine (/usr/lib/node_modules/pino/pretty.js:50:14)
    at DestroyableTransform.mapLine [as mapper] (/usr/lib/node_modules/pino/pretty.js:87:24)
    at DestroyableTransform.transform [as _transform] (/usr/lib/node_modules/pino/node_modules/split2/index.js:33:21)
    at DestroyableTransform.Transform._read (/usr/lib/node_modules/pino/node_modules/readable-stream/lib/_stream_transform.js:159:10)
    at DestroyableTransform.Transform._write (/usr/lib/node_modules/pino/node_modules/readable-stream/lib/_stream_transform.js:147:83)
    at doWrite (/usr/lib/node_modules/pino/node_modules/readable-stream/lib/_stream_writable.js:347:64)
    at writeOrBuffer (/usr/lib/node_modules/pino/node_modules/readable-stream/lib/_stream_writable.js:336:5)
    at DestroyableTransform.Writable.write (/usr/lib/node_modules/pino/node_modules/readable-stream/lib/_stream_writable.js:274:11)
    at Socket.ondata (_stream_readable.js:528:20)
    at emitOne (events.js:77:13)
events.js:141
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at Object.exports._errnoException (util.js:856:11)
    at exports._exceptionWithHostPort (util.js:879:20)
    at WriteWrap.afterWrite (net.js:763:14)

davidmarkclements:
hey @mzehrer can you add the output being logged?

mzehrer:
Sorry, my fault, there was some direct console.log to stdout in one module. This is maybe the downside of the external transport idea?

davidmarkclements:
Good time as any to discuss this

I would say, that if you're using a logger to stdout then a console.log would be stray,
so you that transport (or prettifier) to crash

Maybe a more helpful message in the prettifier? 

Maybe a recommendation in docs to noop console.log etc in production? 

```js
if (process.NODE_ENV === 'production') {
[...]

PR #155, opened 2016-12-30
Fix prettifier crashing when receiving the line: `null`

jsumners:
Issue #154 shows that if the prettifier receives a line value that is the literal text `null\n`, then `isPinoLine` will fail. The issue is that the spec (https://tools.ietf.org/html/rfc7159#section-3) allows literal values `null`, `true`, and `false` as valid JSON in addition to `[]` and `{}`. But `null` is not an object, so it doesn't have the `hasOwnProperty` method; thus the crash.

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -47,6 +47,7 @@ function filter (value) {
 }
 
 function isPinoLine (line) {
+  if (line === null || line === undefined) return false
   return line.hasOwnProperty('hostname') && line.hasOwnProperty('pid') && (line.hasOwnProperty('v') && line.v === 1)
 }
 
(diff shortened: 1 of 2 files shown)


pretty.js
```diff
@@ -47,6 +47,7 @@ function filter (value) {
 }
 
 function isPinoLine (line) {
+  if (line === null) return false
   return line.hasOwnProperty('hostname') && line.hasOwnProperty('pid') && (line.hasOwnProperty('v') && line.v === 1)
```
mcollina: I think we can go multiline with `return line &&`, if `line` is falsy there is no chance for it to have the other properties.

jsumners: Done.

jsumners:
3.1.0 or 3.0.7?

mcollina:
I'd say 3.0.7. it's a bugfix.

PR #156, opened 2017-01-02
Updated license

mcollina:


package.json
```diff
@@ -42,6 +42,10 @@
     "json"
   ],
   "author": "Matteo Collina <hello@matteocollina.com>",
+  "contributors": [
+    "David Mark Clements <huperekchuno@googlemail.com>",
```
davidmarkclements: should this actually be nearform addr?

davidmarkclements: (same q. for you)

mcollina: not really.

marcbachmann: please use the email you're using for the commits

mcollina:
@davidmarkclements are you good to go?

davidmarkclements:
yep

PR #157, opened 2017-01-02
New error serializer

### LINKED ISSUE #146: An error with Errors?
Very minor issue I came across, but it suddenly crashes node when it happens.

When I failed to wrap my Error object (which may be null) in an object, pino threw this message:

Code:

    logger.info(err, 'End of playback');

Error:

    C:\Users\<me>\Projects\<projectname>\node_modules\pino\pino.js:378
        pos = s.indexOf(i, pos);
               ^

    TypeError: Cannot read property 'indexOf' of null
        at countInterp (C:\Users\<me>\Projects\<projectname>\node_modules\pino\pino.js:378:12)
 
I tried doing this in the indicated function:

    pos = s ? s.indexOf(i, pos) : -1

And it revealed that I did not enclose my `err` in an object.

    { ... "msg":"null \"End of playback\"" ... }

However, I usually choose not to wrap `err`, because it results in the following log:

    logger.error({ error: err }, err.message);

    [2016-12-12T09:13:37.051Z] ERROR(appname/### on computer): <The contents of err.message -- this string has content>
       error: {}

The error is rendered as empty, but it does in fact have an message. The only way to get the full stack trace is to use `logger.error(err)`. This scenario works well with Promises (where err is guaranteed in the catch block); in this instance, `err` was a parameter in a callback. Thus, the above.

To wrap up:

1. It boils down to Errors being a special case, for some reason, and it is rendered as an empty object -- is this a bug?

2. Could the if-else above (`pos = s ? s.indexOf(i, s) : -1`) be incorporated into the `countInterp` function? It would at least prevent unexpected crashing.

mcollina:
Includes both #134 and #149.

Fixes #146.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -364,11 +364,17 @@ function asResValue (res) {
 }
 
 function asErrValue (err) {
-  return {
+  var obj = {
     type: err.constructor.name,
     message: err.message,
     stack: err.stack
   }
+  for (var key in err) {
+    if (obj[key] === undefined) {
+      obj[key] = err[key]
+    }
+  }
+  return obj
 }
 
 function countInterp (s, i) {
--- a/pino.js
+++ b/pino.js
@@ -157,7 +157,6 @@ function Pino (opts, stream) {
     if (levelIsStandard === false && valIsStandard === false) this.addLevel(opts.level, opts.levelVal)
   }
   this._setLevel(opts.level)
-
   this._baseLog = flatstr(baseLog +
     (this.name === undefined ? '' : '"name":' + this.stringify(this.name) + ','))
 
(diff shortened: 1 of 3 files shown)


pino.js
```diff
@@ -241,14 +240,13 @@ Pino.prototype.asJson = function asJson (obj, msg, num) {
...
       data += ',"type":"Error","stack":' + this.stringify(obj.stack)
-    } else {
-      for (var key in obj) {
-        value = obj[key]
-        if (obj.hasOwnProperty(key) && value !== undefined) {
-          value = this.stringify(this.serializers[key] ? this.serializers[key](value) : value)
-          if (value !== undefined) {
-            data += ',"' + key + '":' + value
-          }
+    }
```
mcollina: we are not doing it twice, `stack`  is not enumerable, and it won't pop up within the for loop below.

jsumners:
LGTM

PR #164, opened 2017-01-10
.pretty isn't an instance method

watson:
Move the `.pretty` function up to under the constructor docs to highlight that it isn't an instance function on the `logger` object.

I found this discrepancy while working on the `v2.x.x` branch. I'm not sure if this is preferred way to document it, but I at least wanted to make a PR where it could be discussed.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,6 +1,7 @@
 # Table of Contents
 
 + [Constructor](#constructor)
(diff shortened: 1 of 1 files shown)


jsumners:
LGTM

I just maintained the original ordering when re-organizing the docs.

Issue #167, opened 2017-01-15
addLevel is not working


HCanber:
__pino 3.1.1__

This fails with: _TypeError: log.foo is not a function_

``` js
const pino = require('pino')
const log = pino()
log.addLevel('foo', 35)
log.foo('bar')
```

The tests does not show this because:
1. Levels added to one instance via the constructor will be added to new instances
2. Adding a level on the constructor works and there exists a test that verifies this. The test creates a new level with the options: `{level:'foo', levelVal: 35}`
3. The test that verifies `addLevel('foo', 35)` is working because it's executed __after__ the test in _(2)_, and because of _(1)_ it uses that function.

If the test in _(3)_ is modified to `addLevel('foo2', 35)`,  the test fails.

Here is some code to show the problem:

``` js
const log = pino({level:'foo', levelVal:35})
const log2 = pino()
log2.foo('should fail, but will work')
```

Three problems exists in the current solution
- `levels` object is shared
- `lscache` object is shared
- `addLevel` does not add a new property

I have a fix for this, which copies `levels` and `lscache` to the instances when new custom levels are added to them. If no custom levels are added the standard values on `Pino.prototype` is used.

Let me know if you want me to submit a PR.

davidmarkclements:
Hi @HCanber thanks for catching this, we may need to tweak depending on how it affects benchmarks but a PR would be much appreciated!

itavy:
Hi,

i have found a few issues which are related to this one:
1. silent level is not documented at all in docs (it has level 100) and if i try to a loglevel with that value if fails with "unknown log level" 
2. if i add a level via constructor it set logLevel it's value: 
  i think the lovLevel(s) added at construct time should not influence the behavior of the logger;
  
3. if i am using a custom logLevel for debugging purpose pretty.js will fail because it doesn't know corespondent color for that level;
  this can be fixed easy with a default color for unknown logLevels.

i think in the evening i will be able to look into this and submit a PR if it is ok.

mcollina:
> silent level is not documented at all in docs (it has level 100) and if i try to a loglevel with that value if fails with "unknown log level" 

It's hidden because it's a maximum, you can't log at that level.

> if i add a level via constructor it set logLevel it's value: 
i think the lovLevel(s) added at construct time should not influence the behavior of the logger;

#168 and #169 should address this?

[...]

PR #168, opened 2017-01-16
Adds `lscache` as an instance property and fixes add level tests

jsumners:
This PR addresses issue #167 . The problem was that adding a new level to an existing instance relied on `instance.level = 42` to be invoked before the new level function would be present. This PR adds the function when `addLevel()` is invoked, adds `lscache` as an instance property, and fixes the tests so that each test uses a clean instance of Pino.


jsumners:
**issue-167:**
```
==========
basic averages
Bunyan average: 1583.399
Winston average: 3249.66
Bole average: 471.3795
Debug average: 852.1855
LogLevel average: 538.803
Pino average: 526.1445
PinoExreme average: 187.9665
==========
```

**master:**
```
==========
basic averages
Bunyan average: 1578.0115
Winston average: 3304.4545
Bole average: 487.433
Debug average: 880.4190000000001
LogLevel average: 531.9645
Pino average: 520.3565
PinoExreme average: 186.252
==========
```

pino.js
```diff
@@ -220,6 +220,10 @@ Object.defineProperty(Pino.prototype, 'level', {
   set: Pino.prototype._setLevel
 })
 
+Object.defineProperty(Pino.prototype, 'lscache', {
+  value: copy({}, lscache)
```
marcbachmann: We don't use `lscache` anywhere else, so there's no need to copy it. This has exactly the same behavior like it was before.

Your tests only work because you delete the cache, which forces node to reinitialize the complete file.
`lscache` is still shared between multiple `pino` instances.

marcbachmann: Ok. Except if you keep a reference to the old instance. So your fix would work. 👍

mcollina:
Can you please add a quick test for:

```js
const log = pino({level:'foo', levelVal:35})
const log2 = pino()
log2.foo('should fail, but will work')
```

HCanber:
See #169

jsumners:
The explicit test was a better test than removing Pino from the require cache. It highlighted that the shared state bug still existed.

The fix to the shared state does not impact performance as far as I can tell. I got an average of 506ms on one basic benchmark and an average of 521ms on another. This is consistent with other runs, and against `master`, on my machine.

marcbachmann:
You get the same results on the childCreation benchmark?
https://github.com/pinojs/pino/blob/master/benchmarks/childCreation.bench.js

jsumners:
Here's my results from `npm run bench-conception` three times each on `master` and `issue-167`:

**master:**
+ min avg: 303.2115
+ max avg: 320.9715

**issue-167:**
+ min avg: 365.301
+ max avg: 400.97

davidmarkclements:
ok so that's adding 20-25% to child creation...

For #168 vs #169 - in terms of simplicity #168 is the simplest, but in terms of "no simpler" - I'd accept more changes (#169) if we can show better numbers.

@HCanber and/or @jsumners can you generate some comparison benchmarks between the two PR's - for basic logging and for child creation and for child logging

jsumners:
I have recovered the speed of child creation. Here are my results on 3 runs of `bench-conception` with my latest commits:

**issue-167:**
min avg: 315.4315
max avg: 317.5745

**master:**
min avg: 309.0435
max avg: 316.1945

And here's the same for `bench-basic`:

**issue-167:**
min avg:  245.7525
max avg: 281.0665

**master:**
min avg: 246.8005
max avg: 250.5725

HCanber:
I also ran the test three times. I'm showing the min and max averages below.

I don't think you really can draw any other conclusions from either of our numbers other than that they do not significantly change anything.

### basic
__#169__
Pino average: 327.8325 - 344.25
PinoExreme average: 137.645 - 142.3645

__v3.1.1__
Pino average: 311.8325 - 385.3055
PinoExreme average: 134.2765 - 157.55200000000002

### child

__#169__
PinoChild average: 343.22 - 361.780499
PinoExtremeChild average: 147.84550 - 155.6855

__v3.1.1__
PinoChild average: 338.4435 - 385.491
PinoExtremeChild average: 148.3034999 - 163.6855

### conception

__#169__
PinoCreation average: 389.947 - 395.8165
PinoExtremeCreation average: 190.651 - 197.143

__v3.1.1__
PinoCreation average: 371.711 - 420.5085
PinoExtremeCreation average: 184.1485 - 203.723

### Object creations and GC
I don't know if you might want to look at object creations and GC, for example creating a lot of children.

### Change the api?
What do you think about removing `addLevel` and instead only support level creations at constructor time. This would simplify things a lot.
Or is there a need to add levels at a later time?

``` js
const standardLevels = pino.levels
const levels = Object.assign({ foo: 35, foo2: 37}, pino.levels) //Add two new levels
const log = pino({levels})
log.foo('logging at foo level')
```

OT: Why are copies sometimes made with `copy({}, src)` and sometimes with `Object.assign({},src)` ?

jsumners:
> I don't know if you might want to look at object creations and GC, for example creating a lot of children.

My personal opinion is that creating a child for every request is a bad idea. But it is something people do.

> What do you think about removing addLevel and instead only support level creations at constructor time. This would simplify things a lot.

I added the custom level support because Verdaccio patches it into Bunyan, and I want to submit a PR to replace Bunyan with Pino to that project. They patch the feature in at construction, but since I was implementing the feature I figured I'd make it more robust.

> Why are copies sometimes made with copy({}, src) and sometimes with Object.assign({},src) ?

I do not know.

davidmarkclements:
So.. what would be useful is like for like comparisons on the same machine - @jsumners / @HCanber  could you run one branch then the other and share results for comparison?

jsumners:
@HCanber's fork:

**child:**
min avg: 256.306
max avg: 274.5445

**conception:**
min avg: 309.017
max avg: 319.6645

**basic:**
min avg: 250.2905
max avg: 271.7505

My branch:

**child:**
min avg: 257.069
max avg: 263.813

**conception:**
min avg: 319.217
max avg: 321.2899

**basic:**
min avg: 242.18
max avg: 252.851

pino.js
```diff
@@ -138,8 +138,9 @@ function pino (opts, stream) {
 
 defineLevelsProperty(pino)
 
-function Pino (opts, stream) {
-  this.stream = stream
+// Must be invoked via .call() to retain speed of inlining into the
+// Pino constructor.
+function applyOptions (opts) {
```
davidmarkclements: but why not just pass this in as ctx? 

```js
applyOptions(ctx, opts) { 
  ctx.serializers = opts.serializers 
  // ..etc
}
```

davidmarkclements:
ok cool - since #168 has been sped up and still makes fewer alterations, I think we should go with #168 - @jsumners if you could just come back to me on the comment

Also at brief glance, it seems like #169 tests are more thorough, if this is the case could you make the tests for this PR equivalent

@HCanber thanks so much for reporting this and putting you time into this, it's very much appreciated

pino.js
jsumners: It is significantly slower. I tried that route first.

jsumners:
Maybe @hcanber would like to add a PR against this beach to add his tests? That way I'm not stealing his work.

jsumners:
I have added a new commit which addresses issue #177 to this PR. While developing the resolution to that issue the shared state bug affected the added test. Thus, I decided that both issues are intertwined and opted to resolve both in this one PR.

pino.js
mcollina: wow, that's unexpected.

jsumners:
@HCanber do you want to add your tests to this branch? Or do you want me to do it?

mcollina:
@jsumners you can also merge his branch in here, but not edit the `pino.js` file.

jsumners:
@mcollina good point. It is done.

HCanber:
@jsumners Thanks. Been busy at work.

jsumners:
@HCanber no problem. Thank you for finding the issue and collaborating on getting it resolved.

PR #169, opened 2017-01-16
addLevel bugfixes

### LINKED ISSUE #167: addLevel is not working

HCanber:
This PR fixes #167

I've split it into three commits.

- a18f09244a28af865bdda1c96f501812b5504f3e Modifies the existing test to show that the current code fails
- dbdd81956087a54a2efbc3f203f16ea7bff1ea10 Adds a new test, that fails, to show that levels created via the constructor exists on other instances as well
- 886be9cce333ec8f230673bc456317cac379d05b Fixes the problems by not sharing levels and lscache between instances.

## Solution
Default `levels` and `_lscache` properties have been added to `Pino.prototype`. These are used when no custom levels have been added. 

##### addLevel()
The first time `addLevel()` is called on a `Pino` instance, new copies of `levels` and `_lscache` are created on the instance and these are then modified to include the new level.

##### Children of parents with custom levels
If a logger instance has custom levels, these are copied over to children as well, when `child()` is called. 

#### NOTE: Parent and children share `levels` and `_lscache`
The objects in `levels` and `_lscache` are shared between the parent and children. This means that if `addLevel` is called on one instance, all others `levels` and `_lscache` properties will contain the new level as well.

I choose to do it like this, in order to not make copies for every child. I think this is a fair tradeoff. 
Custom levels should be added to the first logger instance, before any child logs have been created, and no new levels should be added after that. `API.md` should probably include a comment about this.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -34,22 +34,38 @@ var levels = {
   debug: 20,
   trace: 10
 }
+Object.freeze(levels)
 
 var nums = Object.keys(levels).reduce(function (o, k) {
   o[levels[k]] = k
   return o
 }, {})
+Object.freeze(nums)
 
-function defineLevelsProperty (onObject) {
+var defaultLevelsValues = {
+  values: levels,
+  labels: nums
+}
+Object.freeze(defaultLevelsValues)
+
+function defineLevelsProperty (onObject, existing, noCopy) {
+  var value = noCopy ? existing : {
+    values: copy({}, existing.values),
+    labels: copy({}, existing.labels)
+  }
   Object.defineProperty(onObject, 'levels', {
-    value: {
-      values: copy({}, levels),
-      labels: copy({}, nums)
-    },
+    value: value,
     enumerable: true
   })
   Object.defineProperty(onObject.levels.values, 'silent', {value: 100})
   Object.defineProperty(onObject.levels.labels, '100', {value: 'silent'})
+}
+
+function defineLsCacheProperty (onObject, value, noCopy) {
+  Object.defineProperty(onObject, '_lscache', {
+    value: noCopy ? value : Object.assign({}, value),
+    enumerable: false
+  })
 }
 
 // IIFE so the keys are cached at module load
@@ -66,12 +82,6 @@ var isStandardLevelVal = (function () {
     return keys.indexOf(val + '') > -1
   }
 }())
-
-// level string catch
-var lscache = Object.keys(nums).reduce(function (o, k) {
-  o[k] = flatstr('"level":' + Number(k))
-  return o
-}, {})
 
 function streamIsBlockable (s) {
   if (s.hasOwnProperty('_handle') && s._handle.hasOwnProperty('fd') && s._handle.fd) return true
@@ -136,7 +146,7 @@ function pino (opts, stream) {
   return logger
 }
 
-defineLevelsProperty(pino)
+defineLevelsProperty(pino, defaultLevelsValues)
 
 function Pino (opts, stream) {
   this.stream = stream
@@ -149,6 +159,10 @@ function Pino (opts, stream) {
   this.chindings = opts.chindings
   this.cache = opts.cache
   this.formatOpts = opts.formatOpts
+  if (opts.customizedLevels) {
+    defineLsCacheProperty(this, opts.customizedLevels._lscache, true)
+    defineLevelsProperty(this, opts.customizedLevels.levels, true)
+  }
 
   if (opts.level && opts.levelVal) {
     var levelIsStandard = isStandardLevel(opts.level)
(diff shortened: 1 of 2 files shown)


pino.js
```diff
@@ -34,24 +34,40 @@ var levels = {
   debug: 20,
   trace: 10
 }
+Object.freeze(levels)
 
 var nums = Object.keys(levels).reduce(function (o, k) {
   o[levels[k]] = k
   return o
 }, {})
+Object.freeze(nums)
```
mcollina: `Object.freeze` has a very bad performance reputation. Given that all those globals are done to speed things up, we cannot probably afford this.

mcollina:
Does this have any effect on performance?

pino.js
HCanber: Yes, my thoughts as well. However they can be removed as well, I just kept them to prevent future bugs like this from happening.

HCanber:
I ran basic, child, grandchild & conception a few times and couldn't see any performance changes. Sometimes the tests were slower, sometimes faster.
I haven't looked at the perf test code though. Do they add custom levels?

mcollina:
No, I'm more concerned about the basic/object performance.

HCanber:
The only thing that would affect that is the changes made to `asJson` where `lscache` no longer is a shared value, but exists on the instance, or `Pino.prototype`:

> ````
> var data = this._baseLog + this._lscache[num] + this.time()
> ````
> https://github.com/pinojs/pino/pull/169/files#diff-04e6fb630166e65e0231f8c63a731aa0R254

All other changes affects `addLevel()` and `child()`

pino.js
```diff
@@ -170,7 +184,7 @@ function Pino (opts, stream) {
 }
 
 Pino.prototype = new EventEmitter()
-defineLevelsProperty(Pino.prototype)
+defineLevelsProperty(Pino.prototype, defaultLevelsValues)
```
jsumners: See my latest update to #168 . This is accomplished with less changes by simply adding the `levels` property at construction instead of on the prototype.

jsumners:
I like the addition of cascading the custom level to the child loggers. If #168 gets accepted as the solution to #167, I'd like to see a PR submitted against the new revision that adds the feature.

pino.js
HCanber: Yes, less code changes, _but_ with your solution you __always__ have to call `defineLevelsProperty` in the `Pino` constructor. For every child. Even if you only use the standard levels. In my solution it's only done when needed, and instead I rely on prototype chaining.

HCanber:
@jsumners This works:

```js
logger.addLevel('foo', 35)
const child = logger.child({})
child.foo('fooo')
```

This does not work:

```js
const child = logger.child({})
logger.addLevel('foo', 35)
child.foo('fooo')  //foo not defined
```

pino.js
davidmarkclements: Tough call, we need numbers to see the effects (prototype lookups also have cost -  but then again v8 is good at amortising)

jsumners: The rest of @pinojs/core may disagree with me, but I think the constructor is the one place Pino can afford a little time.

jsumners:
@HCanber right. Children are not currently included in `addLevel` my patch in #168 merely addresses the bug present in the current implementation -- state being shared when it shouldn't be.

For the record, whichever solution, mine or yours, the community chooses to use is fine with me. I had a few extra minutes to work on it last night, and I'm responsible for adding the feature in the first place, so I submitted my PR.

pino.js
davidmarkclements: yeah child creation isn't a big deal but I'd rather not slow down the benchmarks either hmmmmm

davidmarkclements:
I agree levels cascading is a good thing - and agree that can also be a separate PR in the case of #168 being accepted. What we need is numbers.

Since we're talking about the same thing, can we move any further discussion to #168 (because it predates #169) - see my comment on #168 regarding benchmark comparisons

One final thing - @HCanber & @jsumners  just to say thanks for putting the effort in here, it's much appreciated - I want to make sure we get the best solution and competing solutions is actually an awesome way to do that. Pino is actually in a very fortunate situation here.

mcollina:
The real thing is to get something that does not slow down child creation at this point. That is done quite a bit within pino-http and others.

Issue #172, opened 2017-01-16
Pretty formatting fails when a custom level is added


jsumners:
```js
'use strict'

const pino = require('pino')
const pretty = pino.pretty()
pretty.pipe(process.stdout)
const log = pino({level: 'foo', levelVal: 35}, pretty)

log.foo('hello')
```

Yields:
```
% node foo.js
/private/tmp/16/node_modules/pino/pretty.js:131
    return levelColors[value.level](levels[value.level])
                                   ^

TypeError: levelColors[value.level] is not a function
    at asColoredLevel (/private/tmp/16/node_modules/pino/pretty.js:131:36)
    at DestroyableTransform.mapLine [as mapper] (/private/tmp/16/node_modules/pino/pretty.js:106:48)
    at DestroyableTransform.transform [as _transform] (/private/tmp/16/node_modules/split2/index.js:33:21)
    at DestroyableTransform.Transform._read (/private/tmp/16/node_modules/readable-stream/lib/_stream_transform.js:159:10)
    at DestroyableTransform.Transform._write (/private/tmp/16/node_modules/readable-stream/lib/_stream_transform.js:147:83)
    at doWrite (/private/tmp/16/node_modules/readable-stream/lib/_stream_writable.js:347:64)
    at writeOrBuffer (/private/tmp/16/node_modules/readable-stream/lib/_stream_writable.js:336:5)
    at DestroyableTransform.Writable.write (/private/tmp/16/node_modules/readable-stream/lib/_stream_writable.js:274:11)
    at EventEmitter.Pino.write (/private/tmp/16/node_modules/pino/pino.js:306:17)
    at EventEmitter.LOG (/private/tmp/16/node_modules/pino/pino.js:423:10)
```

jsumners:
Fixed by #173

PR #173, opened 2017-01-16
Pretty accepts custom logLevel

itavy:
fix for #172

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -7,6 +7,7 @@ var Parse = require('fast-json-parse')
 var chalk = require('chalk')
 
 var levels = {
+  default: 'USERLVL',
   60: 'FATAL',
   50: 'ERROR',
   40: 'WARN',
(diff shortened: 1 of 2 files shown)


davidmarkclements:
should we do this? or should we allow flags to set the level num -> colors?

jsumners:
@davidmarkclements it's not allowing customization of the colors, it's fixing no color assigned to custom levels causing a crash.

itavy:
all other transports accept customLogLevel without any other configuration. pretty i see it as another transport and also it shal support anything that receives from pino;

i usually use pretty for local develop and pino-elasticsearch for production or beta testing; 

IMHO transports should be able to interchange without any other configuration

davidmarkclements:
agreed.

davidmarkclements:
v.3.1.2 published to npm and gh released - thanks @itavy !

PR #174, opened 2017-01-16
changed "silent" level to Infinity

itavy:
enforced silent level to be the highest possible value so no custom levels can be set higher than it.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -48,14 +48,17 @@ function defineLevelsProperty (onObject) {
     },
     enumerable: true
   })
-  Object.defineProperty(onObject.levels.values, 'silent', {value: 100})
-  Object.defineProperty(onObject.levels.labels, '100', {value: 'silent'})
+  Object.defineProperty(onObject.levels.values, 'silent', {value: Number.POSITIVE_INFINITY})
+  Object.defineProperty(onObject.levels.labels, Number.POSITIVE_INFINITY.toString(), {value: 'silent'})
 }
 
 // IIFE so the keys are cached at module load
 var isStandardLevel = (function () {
   var keys = Object.keys(levels)
   return function (level) {
+    if (Number.POSITIVE_INFINITY.toString() === level) {
+      return true
+    }
     return keys.indexOf(level) > -1
   }
 }())
@@ -63,6 +66,9 @@ var isStandardLevel = (function () {
 var isStandardLevelVal = (function () {
   var keys = Object.keys(nums)
   return function (val) {
+    if (!isFinite(val)) {
+      return true
+    }
     return keys.indexOf(val + '') > -1
   }
 }())
@@ -203,7 +209,12 @@ Object.defineProperty(Pino.prototype, 'levelVal', {
 })
 
 Pino.prototype._setLevel = function _setLevel (level) {
-  if (typeof level === 'number') { level = this.levels.labels[level] }
+  if (typeof level === 'number') {
+    if (!isFinite(level)) {
+      throw new Error('unknown level ' + level)
+    }
+    level = this.levels.labels[level]
+  }
 
   if (!this.levels.values[level]) {
     throw new Error('unknown level ' + level)
--- a/pino.js
+++ b/pino.js
@@ -48,15 +48,15 @@ function defineLevelsProperty (onObject) {
     },
     enumerable: true
   })
-  Object.defineProperty(onObject.levels.values, 'silent', {value: Number.POSITIVE_INFINITY})
-  Object.defineProperty(onObject.levels.labels, Number.POSITIVE_INFINITY.toString(), {value: 'silent'})
+  Object.defineProperty(onObject.levels.values, 'silent', {value: Infinity})
+  Object.defineProperty(onObject.levels.labels, Infinity, {value: 'silent'})
 }
 
 // IIFE so the keys are cached at module load
 var isStandardLevel = (function () {
   var keys = Object.keys(levels)
   return function (level) {
-    if (Number.POSITIVE_INFINITY.toString() === level) {
+    if (Infinity === level) {
       return true
     }
     return keys.indexOf(level) > -1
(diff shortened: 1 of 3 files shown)


davidmarkclements:
that's a really good point - does it *need* to be Number.POSITIVE_INFINITY ... I think we can just use `Infinity`

marcbachmann:
Sorry. wrong pr. 😆

davidmarkclements:
Especially since Number.POSITIVE_INFINITY as a string is 'Infinity'

pino.js
```diff
@@ -48,21 +48,27 @@ function defineLevelsProperty (onObject) {
     },
     enumerable: true
   })
-  Object.defineProperty(onObject.levels.values, 'silent', {value: 100})
-  Object.defineProperty(onObject.levels.labels, '100', {value: 'silent'})
+  Object.defineProperty(onObject.levels.values, 'silent', {value: Number.POSITIVE_INFINITY})
+  Object.defineProperty(onObject.levels.labels, Number.POSITIVE_INFINITY.toString(), {value: 'silent'})
```
davidmarkclements: no need to convert to string, adding a numerical key to an object turns it into a string anyway

itavy:
hi,

it is a personal preference from my early days in js when Infinity wasn't
readonly like it is in es5. also i like to have it so to show that i am
using specific number comparison for levelVal.


at the end is a metter of preference since both do the same thing so it can
be changed if you like it so.

pino.js
marcbachmann: After this change it wouldn't be possible anymore to set the `silent` loglevel by assigning `100`.
Would that be a breaking change?

davidmarkclements:
yeah I see your points - for me it's about doing the least - 

I'll defer to @mcollina and @jsumners to weigh in

jsumners:
@itavy we support Node 4.x and later. Browser support is also inline with not needing to use archaic style. So I side with @davidmarkclements.

itavy:
that depends of the use cases of pino.

if people rely on setting by value the silent option, yes. (this option is
not even documented)

if they rely on silent it by enable=false in constructor or by setting it
in level property, than no

jsumners:
"silent" mode is only documented as the `enabled` option -- https://github.com/pinojs/pino/blob/master/docs/API.md#pinooptions-stream

That being said, `var log = pino({level: 'silent'})` works as well. If someone was setting it by `var log = pino({level: 100})` then they were already using internal API. Internal API changes are not breaking changes as long as the published API continues to work the same in my opinion.

davidmarkclements:
let's bring @watson into this - Thomas would you advantage at all from `Number.POSITIVE_INFINITY` or does `Infinity` work for legacy branch too?

davidmarkclements:
@jsumners ... I don't think we've been explicit enough about setting level by number as an internal API - do you know where in the docs we say this? 

I'm not sure if we can even call it an internal API ...

In any event, even if it is internal, this should probably preclude a major release..

pino.js
```diff
@@ -203,7 +209,12 @@ Object.defineProperty(Pino.prototype, 'levelVal', {
 })
 
 Pino.prototype._setLevel = function _setLevel (level) {
-  if (typeof level === 'number') { level = this.levels.labels[level] }
+  if (typeof level === 'number') {
+    if (!isFinite(level)) {
+      throw new Error('unknown level ' + level)
```
davidmarkclements: as a matter of style, we avoid `new` if we possibly can - `Error` can be called without `new` to the same effect

test/addlevel.test.js
```diff
@@ -79,3 +79,12 @@ test('reject already known values', function (t) {
     t.is(e.message.indexOf('level value') > -1, true)
   }
 })
+
+test('reject values of Infinity', function (t) {
+  t.plan(1)
+  try {
+    pino({level: 'foo', levelVal: Infinity})
+  } catch (e) {
+    t.is(e.message.indexOf('level value') > -1, true)
```
davidmarkclements: can we use `t.throws` instead

test/levels.test.js
```diff
@@ -146,18 +146,19 @@ test('silent level', function (t) {
...
-  }, sink(function (chunk, enc, cb) {
-    t.fail('no data should be logged')
-  }))
-
-  Object.keys(pino.levels.values).forEach(function (level) {
-    instance[level]('hello world')
-  })
-  t.end()
-})
+// obsolete by setting silent to Infinity
```
davidmarkclements: please just remove

jsumners:
@davidmarkclements I was referencing the silent level number being internal API.

pino.js
itavy: i used the same style as you used; the same construct is in the next if block; 
do you want to make this change in both places since i am already there ?

davidmarkclements:
ohhh I see - did we doc that 100 is internal though? In a much more limited sense, it still might require a major 

This is possibly a false memory but I'm fairly certain I recall wanting to set it as Infinity in the first place but @mcollina didn't want to... I.. might be making that up though

pino.js
davidmarkclements: yes please - :bowtie:

jsumners:
@davidmarkclements here's all of the commits related to `silent` -- https://github.com/pinojs/pino/search?q=silent&type=Commits&utf8=%E2%9C%93

So it's also documented at -- https://github.com/pinojs/pino/blob/master/docs/API.md#level

In all documentation, the internal value of `silent` is not mentioned. And it's not added to the public `levels` property, either. So anyone using `100` to silence the logger is either an author of Pino, or has dug into the code to find it.

mcollina:
So, let me say what I was thinking for the `silent` level, and yes, @davidmarkclements was right to say I _had_ a strong opinion to be 100. It was before we had custom levels, where we were following only Bunyan spec.

Custom levels changes everything, so I am 👍  on this change.

mcollina:
Just one check before merging.. @watson is this ok to be backported?

Issue #175, opened 2017-01-17
Integration with restify-errors


nodo:
Hi all,

I have an issue in using `pino` and `restify-errors` (https://github.com/restify/errors).

I assumed that when pino see an error object with a stack, it automatically log the stack.

For instance:
```
> var pino = require('pino');
> var log = pino();
> log.info(new Error());
{"pid":70521,"hostname":"mll1p05607.local","level":30,"time":1484658304444,"msg":"","type":"Error","stack":"Error\n    at repl:1:10\n    at sigintHandlersWrap (vm.js:22:35)\n    at sigintHandlersWrap (vm.js:96:12)\n    at ContextifyScript.Script.runInThisContext (vm.js:21:12)\n    at REPLServer.defaultEval (repl.js:313:29)\n    at bound (domain.js:280:14)\n    at REPLServer.runBound [as eval] (domain.js:293:12)\n    at REPLServer.<anonymous> (repl.js:513:10)\n    at emitOne (events.js:101:20)\n    at REPLServer.emit (events.js:188:7)","v":1}
```

However, when using `restify-errors` the stacktrace is ignored.
```
> var e = require('restify-errors');
> log.info(new e.InternalServerError())
{"pid":70521,"hostname":"mll1p05607.local","level":30,"time":1484658353255,"res":{"statusCode":500},"v":1}
```

Note that both errors have a stacktrace:
```
> new Error().stack
'Error\n    at repl:1:1\n    at sigintHandlersWrap (vm.js:22:35)\n    at sigintHandlersWrap (vm.js:96:12)\n    at ContextifyScript.Script.runInThisContext (vm.js:21:12)\n    at REPLServer.defaultEval (repl.js:313:29)\n    at bound (domain.js:280:14)\n    at REPLServer.runBound [as eval] (domain.js:293:12)\n    at REPLServer.<anonymous> (repl.js:513:10)\n    at emitOne (events.js:101:20)\n    at REPLServer.emit (events.js:188:7)'
> new e.InternalServerError().stack
[...]

PR #176, opened 2017-01-17
Fix http response filtering

nodo:
@mcollina In order to distinguish http response I have also tried to use `instanceof`. I decided to go for `setHeader` as you suggested because it seems slightly faster.

I would be happy to improve the PR and integrate your comments.

Fixes https://github.com/pinojs/pino/issues/175

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -388,14 +388,6 @@ function countInterp (s, i) {
   return n
 }
 
-function isHttpRequest (m) {
-  return m.method && m.headers && m.socket
-}
-
-function isHttpResponse (m) {
-  return typeof m.setHeader === 'function'
-}
-
 function genLog (z) {
   return function LOG (a, b, c, d, e, f, g, h, i, j, k) {
     var l = 0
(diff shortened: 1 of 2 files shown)


pino.js
```diff
@@ -400,9 +408,9 @@ function genLog (z) {
       n = [b, c, d, e, f, g, h, i, j, k]
       l = 1
 
-      if (m.method && m.headers && m.socket) {
+      if (isHttpRequest(m)) {
         m = mapHttpRequest(m)
-      } else if (m.statusCode) {
+      } else if (isHttpResponse(m)) {
```
mcollina: can you please inline the checks again? Using functions will increase the number of functions that can be inlined.

nodo: Thanks for the review, sure I can do that. Is that for performance reason?

mcollina: Yes. There is a maximum number of functions that can be inlined automatically by V8.

jsumners:
@mcollina are you going to merge this one?

mcollina:
@jsumners if you have time, go ahead and release this and #174 bumping the minor.

jsumners:
Okay. I'll do it after work.

PR #179, opened 2017-01-25
Reorganize code for legibility

jsumners:
This PR splits up the one massive `index.js` into multiple source files. Over time, the `index.js` has had new features and fixes tacked on in seemingly arbitrary locations (I know we all choose what we think makes sense). The result is a mess of code that is difficult to navigate when working through issues or trying to add a feature.

I have run all of the benchmarks and do not see any real difference between this and the master branch. I am attaching outputs from three runs of `bench-all` on each branch so you can decide for yourself (or show me that I missed something).

[pino-reorg.bench.txt](https://github.com/pinojs/pino/files/729785/pino-reorg.bench.txt)
[pino-master.bench.txt](https://github.com/pinojs/pino/files/729784/pino-master.bench.txt)

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -1,0 +1,49 @@
+'use strict'
+
+var flatstr = require('flatstr')
+
+var levels = {
+  fatal: 60,
+  error: 50,
+  warn: 40,
+  info: 30,
+  debug: 20,
+  trace: 10
+}
+module.exports.levels = levels
+
+var nums = Object.keys(levels).reduce(function (o, k) {
+  o[levels[k]] = k
+  return o
+}, {})
+module.exports.nums = nums
+
+// level string cache
+var lscache = Object.keys(nums).reduce(function (o, k) {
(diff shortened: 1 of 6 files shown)


lib/levels.js
```diff
@@ -0,0 +1,49 @@
...
+
+var levels = {
+  fatal: 60,
+  error: 50,
+  warn: 40,
+  info: 30,
+  debug: 20,
+  trace: 10
+}
+module.exports.levels = levels
```
marcbachmann: What do you think about moving the whole exports to the bottom of the file.
I usually just use `module.exports`
```
module.exports = {
  nums: nums,
  lscache: lscache,
  isStandardLevel: isStandardLevel,
  isStandardLevelVal: isStandardLevelVal
}
```

That way you can make sure that there's nothing else exported.
IMO it improves the readability a bit.

jsumners: Can do.

jsumners: Done.

PR #181, opened 2017-01-26
fix for null proto closes #180

### LINKED ISSUE #180: obj.hasOwnProperty is not a function
Found this deep inside our server logs

```
obj.hasOwnProperty is not a function
    at EventEmitter.asJson (/Users/ronagy/Projects/github/nxtedition/nxt-storage/node_modules/pino/pino.js:175:15)
    at EventEmitter.Pino.write (/Users/ronagy/Projects/github/nxtedition/nxt-storage/node_modules/pino/pino.js:224:16)
    at EventEmitter.LOG (/Users/ronagy/Projects/github/nxtedition/nxt-storage/node_modules/pino/lib/tools.js:116:10)
```

Not sure how to reproduce it.

davidmarkclements:
Master:

Pino average: 271.8075
PinoExreme average: 132.1225

Branch: 

Pino average: 271.576
PinoExreme average: 135.951

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -172,7 +172,7 @@ Pino.prototype.asJson = function asJson (obj, msg, num) {
     }
     for (var key in obj) {
       value = obj[key]
-      if (obj.hasOwnProperty(key) && value !== undefined) {
+      if ((!obj.hasOwnProperty || obj.hasOwnProperty(key)) && value !== undefined) {
         value = this.stringify(this.serializers[key] ? this.serializers[key](value) : value)
         if (value !== undefined) {
           data += ',"' + key + '":' + value
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -377,3 +377,21 @@ test('normalize number to string with an object', function (t) {
 
   instance.info({ answer: 42 }, 1)
 })
+
+test('handles objects with null prototype', function (t) {
+  var instance = pino(sink(function (chunk, enc, cb) {
+    delete chunk.time
+    t.deepEqual(chunk, {
+      pid: pid,
+      hostname: hostname,
+      level: 30,
+      test: 'test',
+      v: 1
+    })
+    t.end()
+    cb()
+  }))
+  var o = Object.create(null)
+  o.test = 'test'
+  instance.info(o)
+})


pino.js
```diff
@@ -172,7 +172,7 @@ Pino.prototype.asJson = function asJson (obj, msg, num) {
     }
     for (var key in obj) {
       value = obj[key]
-      if (obj.hasOwnProperty(key) && value !== undefined) {
+      if (!obj.hasOwnProperty || obj.hasOwnProperty(key) && value !== undefined) {
```
marcbachmann: I'd use parentheses around that condition.
`(!obj.hasOwnProperty || obj.hasOwnProperty(key)) && value !== undefined`

test/basic.test.js
```diff
@@ -377,3 +377,21 @@ test('normalize number to string with an object', function (t) {
 
   instance.info({ answer: 42 }, 1)
 })
+
+test('handles objects with null prototype', function (t) {
+  var instance = pino(sink(function (chunk, enc, cb) {
+    delete chunk.time
```
marcbachmann: What about a test for `undefined` values?
It's basically already covered in other tests, so not necessarily needed.

pino.js
jsumners: I agree.

test/basic.test.js
davidmarkclements: As you say covered - but another pr could make add such a test

pino.js
davidmarkclements: Yeeaaah okay

jsumners:
After this PR, I decided to use the https://www.npmjs.com/package/conf package in a project. In the project I needed to check for a property on the resulting config object. It turns out that `conf` initializes its internal object as `Object.create(null)`. So, on topic, this is a thing that actually occurs. Off-topic, do any of you have any clue why anyone would do this? It very clearly breaks the expectation of an object's prototype.

mcollina:
Speed is one. Objects created in this way are slightly faster. A notable one is _headers object inside http requests.

It is rare but it occurs.

jsumners:
Then a literal should be used:

```
nullPrototype*1000: 0.335ms
literalPrototype*1000: 1.005ms
literal*1000: 0.141ms
nullPrototype*1000: 0.146ms
literalPrototype*1000: 1.089ms
literal*1000: 0.088ms
```

https://gist.github.com/jsumners/26bfee4a1ea04b4d7a5829e5e913ce29

davidmarkclements:
we're using it in hashlru https://github.com/dominictarr/hashlru/pull/6

The memory of that PR tipped me off about the problem

davidmarkclements:
v3.3.2 released

PR #185, opened 2017-02-06
docs: clarify where pretty() is

mcollina:
cc @jsumners Is this clearer?

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,8 +1,8 @@
 # Table of Contents
 
-+ [Constructor](#constructor)
-  * [.pretty](#pretty)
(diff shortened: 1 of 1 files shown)


docs/API.md
```diff
@@ -1,7 +1,7 @@
 # Table of Contents
 
-+ [Constructor](#constructor)
-  * [.pretty](#pretty)
++ [pino](#constructor)
++ [pino.pretty](#pretty)
 + [Logger](#logger)
```
jsumners: Maybe this should be renamed "Logger Instance" as well?

mcollina:
@jsumners done.

Issue #184, opened 2017-02-05
Add CLI formatter function to pino module


jmeas:
Related:

https://github.com/trentm/node-bunyan/pull/102
https://github.com/trentm/node-bunyan/issues/84

This could be pretty handy in different situations. If there's any interest, I could try my hand at whipping up a PR. What do y'all think?

jsumners:
As in the formatter option on `.pretty()`? https://github.com/pinojs/pino/blob/master/docs/API.md#pretty

mcollina:
Yes, I think we should add the possibility to pass that as an argument.

jsumners:
@mcollina I'm actually unsure what the request is asking for. I *think* it's asking for a way to replace the default output of line delimited JSON with something else. In that case, we already support it via `.pretty()`.

I have also found the boilerplate of setting up pretty output to be cumbersome, but have been undecided if the following would be worth adding:

```js
var pino = require('pino')
var log = pino({pretty: true, extreme: false})
```

With the backing code being:

```js
if (pretty) {
  this.stream = pretty()
  this.stream.pipe(process.stdout)
}
```

mcollina:
@jmeas I understood we wanted an equivalent of `bunyan -c 'this.foo == "bar"`, so that you could filter the log lines. Is this correct?

@jsumners I'm 👎  on `{ pretty: true }` option, as we are wasting a lot of CPU time in parsing with that enabled.

jsumners:
@mcollina right. But I find myself adding stuff like the following to every project:

```js
var appConfig = function readTheConfig() {}
var pino = require('pino')
var log
if (appConfig.log.pretty) {
  var pretty = pino.pretty()
  pretty.pipe(process.stdout)
  log = pino(appConfig.log, pretty)
} else {
  log = pino(appConfig.log)
}
```

As I said, I've been torn about submitting a PR that would reduce that to `var log = pino(appConfig.log)`. But if you're against it, then I'll just keep doing what I'm doing.

But the way I read this issue, that's pretty much what @jmeas is wanting. He just wants to also supply the custom formatter function. Maybe I'm way off base.

mcollina:
@jsumners go ahead and send a PR then, that code is ugly, also https://github.com/pinojs/hapi-pino/blob/master/index.js#L28-L32 :D.

jsumners:
@mcollina okay, I'll get it in soon-ish.

jmeas:
[...]

PR #186, opened 2017-02-06
Add prettyPrint to pino constructor options

jsumners:
See the discussion in issue #184 for the motivation to this PR.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -53,6 +53,8 @@
     Default: 'info'.
   * `levelVal` (integer): when defining a custom log level via `level`, set to an
     integer value to define the new level. Default: `undefined`.
+  * `prettyPrint` (boolean|object): enables [pino.pretty](#pretty). This is intended for non-production
+    configurations. This may be set to a configuration object as outlinedin [pino.pretty](#pretty). Default: `false`.
   * `enabled` (boolean): enables logging. Default: `true`
 + `stream` (Writable): a writable stream where the logs will be written.
   Deafult: `process.stdout`
@@ -83,6 +85,8 @@ Returns a new [logger](#logger) instance.
   timestamp to ISO 8601 date format, and reserialize the JSON (equivalent to `pino -t`).
   * `formatter` (function): a custom function to format the line, is passed the
   JSON object as an argument and should return a string value.
+  * `levelFirst` (boolean): if set to `true`, it will print the name of the log
+  level as the first field in the log line. Default: `false`.
 
 ### Example:
 ```js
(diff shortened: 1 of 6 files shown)


docs/API.md
```diff
@@ -53,6 +53,8 @@
     Default: 'info'.
   * `levelVal` (integer): when defining a custom log level via `level`, set to an
     integer value to define the new level. Default: `undefined`.
+  * `prettyPrint` (boolean|object): enables [pino.pretty](#pretty). This is intended for non-production
+    configurations. This may be set to a configuration object as outlinedin [pino.pretty](#pretty). Default: `false`.
```
jmeas: small thing:

```diff
+ outlinedin
- outlined in
```

docs/API.md
```diff
@@ -83,6 +85,8 @@ Returns a new [logger](#logger) instance.
   timestamp to ISO 8601 date format, and reserialize the JSON (equivalent to `pino -t`).
   * `formatter` (function): a custom function to format the line, is passed the
   JSON object as an argument and should return a string value.
+  * `levelFirst` (boolean): if set to `true`, it will print the name of the log
+  level as the first field in the log line. Default: `false`.
```
jmeas: I may be missing something obvious, but I don't see this referenced in the code.

jmeas: Ah, I see – was this just missing from the docs?

jsumners: Yes, it was just missing from the docs. Noticed it while adding the feature.

pino.js
```diff
@@ -30,8 +32,15 @@ function pino (opts, stream) {
...
     iopts = defaultOptions
   }
-  istream = istream || process.stdout
   iopts = Object.assign({}, defaultOptions, iopts)
+  if (iopts.extreme && iopts.prettyPrint) throw Error('cannot enable pretty print in extreme mode')
+  istream = istream || process.stdout
+  if (iopts.prettyPrint) {
+    var pstream = pretty(iopts.prettyPrint)
+    var origStream = istream
+    pstream.pipe(origStream)
```
mcollina: I think we should use `pump`.

jsumners: Done. But I left off an error function because we are not within an actual instance at that point. Thus, we can't emit the error.

mcollina: we can because the logger will be there if an error is ever emitted. It's never emitted synchronously during `pipe` with streams.

jsumners: Okay, I added it. Personally, I don't like it because it relies on hoisting.

davidmarkclements:
note - unrelated failing test in Node 7

jsumners:
I bet that error has to do with issue #131.

mcollina:
@jsumners you bet correctly

jmeas:
🎉

mcollina:
🎉

PR #187, opened 2017-02-07
PR to solve issue #92

jsumners:
- [x] tests waiting for file descriptors
- [x] tests logs are output when event loop isn't really a loop
- [x] removes unexpected call to `process.exit`

In the current state, it solves #131. But it does not fully address #92.

I have created this PR so that we can think on the problem together. Please provide your suggestions, or some commits.


pino.js
```diff
@@ -59,36 +59,39 @@ function pino (opts, stream) {
...
+  function extremeModeExitHandler (code, evt) {
+    var buf = iopts.cache.buf
+    if (buf) {
+      // We need to block the process exit long enough to flush the buffer
+      // to the destination stream. We do that by forcing a synchronous
+      // write directly to the stream's file descriptor.
+      var fd = (istream.fd) ? istream.fd : istream._handle.fd
+      fs.writeSync(fd, buf)
+    }
+    if (!process._events[evt] || process._events[evt].length < 2 || !process._events[evt].filter(function (f) {
```
davidmarkclements: for readability maybe assign the condition to an intermediate var, and give the filter predicate function a name, then pass by ref to filter

pino.js
```diff
@@ -59,36 +59,39 @@ function pino (opts, stream) {
...
+      return f + '' !== events.onExit.passCode + '' && f + '' !== events.onExit.insertCode + ''
+    }).length) {
+      process.exit(code)
+    } else {
+      return 'no exit'
```
davidmarkclements: not keen on returning string here...  where does it go? to a log?

pino.js
```diff
@@ -59,36 +59,39 @@ function pino (opts, stream) {
...
+    var buf = iopts.cache.buf
+    if (buf) {
+      // We need to block the process exit long enough to flush the buffer
+      // to the destination stream. We do that by forcing a synchronous
+      // write directly to the stream's file descriptor.
+      var fd = (istream.fd) ? istream.fd : istream._handle.fd
+      fs.writeSync(fd, buf)
+    }
+    if (!process._events[evt] || process._events[evt].length < 2 || !process._events[evt].filter(function (f) {
+      return f + '' !== events.onExit.passCode + '' && f + '' !== events.onExit.insertCode + ''
```
davidmarkclements: where to passCode and insertCode come from? why are they on an event function?

pino.js
```diff
@@ -59,36 +59,39 @@ function pino (opts, stream) {
...
+    }).length) {
+      process.exit(code)
```
davidmarkclements: I'm not sure it's so bad to call process.exit... what if we supply the option `doNotExitProcessInExtremeModeFdFailure`

pino.js
jsumners: I don't know man, you wrote it :)

https://github.com/pinojs/pino/commit/f7bb83d248634dbb2a83ad272aa6ec6507e5053b

pino.js
jsumners: Recall that calling `process.exit` is the whole reason I opened #92 in the first place. I think that a library shouldn't be killing it's parent application. At least not without letting the parent application do what it needs to do on exit.

pino.js
jsumners: That's you again -- https://github.com/pinojs/pino/commit/16540b81a856feb52eee29fa2d8c6da14fda8388

pino.js
davidmarkclements: ahahahaha

davidmarkclements:
so now - do we merge this or do (as @jsumners originally suggests) use this as a conversation point to finally nail down the whole process.exit thing

@mcollina ?

jsumners:
In line with that question, I had an idea this morning for re-working the event registration that goes a little something like this:

**events.js**
```js
function theWorldIsBurning () {
  if (this.extremeHandlerInvoked) {
    return
  }
  // flush extreme buffer
}

var eventHandler = {
  uncaughtException: theWorldIsBurning,
  sigterm: theWorldIsBurning,
  // other events that handle things differently
}

module.exports = function (pinoInstance, pinoInternalHandler) {
  var handler = Object.create(eventHandler)
  handler.instance = pinoInstance
  return handler
}
```

I _think_ this will make it easier to handle the case where multiple events are emitted during shutdown, and provide a mechanism for working with the parent application in regard to `process.exit`. But I haven't had time to sit down, try it out, and work out the details.

mcollina:
@jsumners 👍 with your example. You can merge or you can keep working here. As you wish.

davidmarkclements:
hehehe can we keep the function name `theWorldIsBurning`

jsumners:
@mcollina @davidmarkclements please review the latest commit. It refactors the events handling. Now, Pino supports an `onTerminated` option to allow the user to handle their cleanup if necessary. If the option is not supplied, it defaults to invoking `process.exit`.

pino.js
```diff
@@ -23,6 +23,9 @@ var defaultOptions = {
   level: 'info',
   levelVal: undefined,
   prettyPrint: false,
+  onTerminated: function () {
+    process.exit(0)
```
mcollina: we should check if we have an error, and in case exit with `process.exit(1)`, that what node would do, and what is expected unix-wise.

jsumners: Done.

jsumners:
I'm torn. On the one hand the functionality should remain the same: if you don't supply an `onTerminated` function then Pino is still exiting as it did before. So that would qualify for minor. But on the other hand this can be a big change to the way shutdown is handled under extreme mode. So a major.

@davidmarkclements do you have a vote?

A note about `SIGHUP`: as I mentioned in Austin, the `SIGHUP` signal is usually a "don't quit, just re-read your configuration" signal in *nix-land [[1](http://perldoc.perl.org/perlipc.html#Handling-the-SIGHUP-Signal-in-Daemons)]. We should probably think about providing a separate handler for that one in a future release. But I'd rather wait to see if anyone says the need it.

pino.js
```diff
@@ -59,36 +63,32 @@ function pino (opts, stream) {
...
-      var buf = iopts.cache.buf
-      if (buf) {
-        // We need to block the process exit long enough to flush the buffer
-        // to the destination stream. We do that by forcing a synchronous
-        // write directly to the stream's file descriptor.
-        var fd = (istream.fd) ? istream.fd : istream._handle.fd
-        fs.writeSync(fd, buf)
-      }
-      if (!process._events[evt] || process._events[evt].length < 2 || !process._events[evt].filter(function (f) {
-        return f + '' !== events.onExit.passCode + '' && f + '' !== events.onExit.insertCode + ''
```
mcollina: this logic should be moved to the new code, because this was preventing `process.exit` in case of an installed `'SIGHUP'` handler. We might also use https://nodejs.org/api/events.html#events_emitter_listenercount_eventname to count those, as we are support 4+ now.

@davidmarkclements what this big check matching strings was doing?

jsumners: Okay, I'll get the intention worked in. This code was too opaque.

mcollina: @jsumners yes it was :/.

jsumners:
I have added a special handler for `SIGHUP`. If we are the only listener for the event then we proceed as we do in all other cases. If we are not the only listener then we simply flush the buffer.

mcollina:
Let's ponder a bit if we want to release it as semver-major or semver-minor.

@watson @marcbachmann ?

jsumners:
I'm thinking major.

mcollina:
me too, but I'd love to ear from the others as well.

davidmarkclements:
semver is about managing expectations - en masse the behavioral change is likely to affect close to 0% of people. If anyone is actually using extreme mode, the change is unlikely to do any harm, or cause a mismatch of expectations. For me this is a minor 

also - great job @jsumners!

lib/events.js
```diff
@@ -1,29 +1,37 @@
...
-  function handle (evt, code) {
-    onExit.passCode = function (code) {
-      if (oneFn.value) { oneFn = once(fn) }
-      oneFn(code, evt)
-    }
-    onExit.insertCode = function () {
-      if (oneFn.value) { oneFn = once(fn) }
-      oneFn(code, evt)
+  function hup (err) {
+    if (process.listenerCount('SIGHUP') === 1) {
```
marcbachmann: @jsumners attention, this doesn't work in node < 4.0
https://nodejs.org/api/events.html#events_emitter_listenercount_eventname

jsumners: @marcbachmann I don't care.

davidmarkclements: we dropped support version less than 4.0 - @watson supports the 0.10-0.12 legacy branch

pino.js
marcbachmann: Oh, here's another mention to that `listenerCount` problem 😄 
@watson would have to back-port it.

davidmarkclements: not necessarily - he could just deprecate extreme mode in legacy - or leave as is with caveats

jsumners:
I don't know why the tests would be timing out on the CI server. They are running fine over here, even on 4.7.3 (which is what Travis says it is using). These tests requiring some timing due to the settling of the stream's file descriptor. If we send signals prior to that, then all bets are off.

lib/events.js
marcbachmann: > @marcbachmann I don't care. Mainline Pino only supports the current LTS releases and up. I'm not concerned with anything other than that.
 
True, there was a discussion about that. Perfectly fine. 👍 
Thereby I'm for the minor change.

davidmarkclements:
other than resolving build issue - am I right in saying the only thing we need to do is decide major/minor? 

I see myself and @marcbachmann are minor - @jsumners you were major - but indicated a possible shift to minor? Can you clarify - and @mcollina what's you preference?

mcollina:
I'm in for major. The _only_ breakage in this PR is that in master, the behavior we have for SIGHUP here is extended to all other events.

I'm not sure we are doing the _right_ thing in master anyway, so we might consider this a bug, release it as minor, and backport it to v2 as well, for @watson sake :).

I'm fine with minor, but if we want to be conservative and not backport it, let's do major.

jsumners:
I think @mcollina is right. The functionality has changed from `master`. This PR only checks for other listeners when the `SIGHUP` event is triggered. In the other cases, assuming no `onTerminated` method has been supplied, we are not checking for other handlers. The thinking behind that is:

1. the other cases are a definite "we need to terminate" scenario
2. we provide very clear documentation about what happens when `onTerminated` is not present

So, I still prefer major.

davidmarkclements:
The argument for disparity between master and legacy is a convincing one - this gives legacy the option to update (and likewise major release) or not. Whereas with a minor, there's more of an impetus for legacy to keep up.

I'm fine with major for that reason

Really hoping 0.10, 0.12 usage die out sooner rather than later.

jsumners:
Now, does anyone have any clues about this Travis nonsense?

marcbachmann:
Looks like that's related to the code coverage built into `tap`.
I'm having the same problem locally when I run `npm run ci`.
`nyc` is probably setting up some exit handlers and therefore your listeners never get notified.

jsumners:
@marcbachmann looks like you are correct -- https://github.com/istanbuljs/nyc/issues/470

jsumners:
I think `nyc` will have to update to `foreground-child` 1.5.6 or later (https://github.com/tapjs/foreground-child/issues/20), and then `node-tap` will need to update to that version of nyc. Outside of that, I don't know how to make these, necessary, tests work under CI.

pinging @isaacs to see if he can shed more light on why https://github.com/pinojs/pino/blob/a78de4fc2bc3835d6ebcba9e8ad9ac2f5f813909/test/events.test.js#L89 and https://github.com/pinojs/pino/blob/a78de4fc2bc3835d6ebcba9e8ad9ac2f5f813909/test/events.test.js#L110 fail with code coverage enabled.

mcollina:
@jsumners that is not the problem. if we bump to tap v10, we have foreground-child v1.5.6 and `npm run ci` is still not working.

isaacs:
I AM SUMMONED.  I have no idea why that's failing, but if there's a problem with tap, nyc, spawn-wrap, and/or foreground-child, I'd be happy to take a look at a reduced test case.

jsumners:
@isaacs @mcollina solved it by removing whatever `SIGHUP` handlers are added by the code coverage libraries prior to us registering our handlers in our test.

Summary of the issue:

1. We have a case where we register for process shutdown events
2. If we are the only people, under normal circumstances, listening for those events then we will handle process exit
3. We have a test case to make sure point 2 is true
4. Test case in 3 was never finishing due to the code coverage hooks into the same events

Basically, the code coverage tool changed the expected environment and I was unaware that happens.

PR #192, opened 2017-02-16
add caller printing debug feature, closes #191

### LINKED ISSUE #191: print caller name for debugging purpose
hi,

it would be nice to have the ability to print the caller function name of the logging functions in order to easily remove useless log statements when you don't know where they are located in big projects.
obvioulsy this feature will kill the awesome performance of pino so it **MUST** be disabled by default.

i will open a PR with the proposed feature in order to discuss about it.

phra:
see #191 for more infos.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -31,6 +31,7 @@ function applyOptions (opts) {
   this.cache = opts.cache
   this.formatOpts = opts.formatOpts
   this.onTerminated = opts.onTerminated
+  this.caller = opts.caller
 
   if (opts.level && opts.levelVal) {
     var levelIsStandard = levels.isStandardLevel(opts.level)
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "flatstr": "^1.0.4",
     "pump": "^1.0.2",
     "quick-format-unescaped": "^1.0.0",
-    "split2": "^2.0.1"
+    "split2": "^2.0.1",
+    "stacktrace-js": "^1.3.1"
   }
 }
--- a/pino.js
+++ b/pino.js
@@ -10,6 +10,7 @@ var events = require('./lib/events')
 var levels = require('./lib/levels')
 var tools = require('./lib/tools')
 var serializers = require('./lib/serializers')
+var stacktrace = require('stacktrace-js')
 
 var LOG_VERSION = 1
 
@@ -27,7 +28,8 @@ var defaultOptions = {
     if (err) return process.exit(1)
     process.exit(0)
   },
-  enabled: true
+  enabled: true,
+  caller: false
 }
 
 function pino (opts, stream) {
@@ -192,6 +194,10 @@ Pino.prototype.asJson = function asJson (obj, msg, num) {
       }
     }
   }
+  if (this.caller) {
+    data += ',"caller":"' + stacktrace.getSync()[3] + '"'
+  }
+
   return data + this.chindings + this.end
 }
 


jsumners:
Can we see some sample output? How about some benchmarks with it disabled versus the master branch?

phra:
i've created a live example of the feature at https://github.com/phra/pino-caller-usage
the performance are the same when the options is disabled thanks to v8 DCE because this [line](https://github.com/phra/pino/blob/features/caller/pino.js#L197) will become `if (false)`.

an output example:
```
╭─phra at debian in /home/phra/git/pino-usage (master ✔)
╰─λ node index.js                                      0 < 21:35:20
{"pid":41795,"hostname":"debian","level":50,"time":1487281257376,"msg":"lol","caller":"Object.<anonymous>()@/home/phra/git/pino-usage/module.js:3:6","v":1}
{"pid":41795,"hostname":"debian","level":30,"time":1487281257380,"msg":"hello world","caller":"Object.<anonymous>()@/home/phra/git/pino-usage/index.js:6:6","v":1}
{"pid":41795,"hostname":"debian","level":50,"time":1487281257380,"msg":"hello world","caller":"Object.<anonymous>()@/home/phra/git/pino-usage/index.js:7:6","v":1}
```
as you can see the logger will display the filename and also the line number where the invocation is located.

jsumners:
I'd still like to see the output of at least `npm run bench-basic` from both branches.

@mcollina @davidmarkclements what are your thoughts on this? If it's accepted, would this necessitate an increase to the log version number?

phra:
see attachments.
[master.perf.txt](https://github.com/pinojs/pino/files/781672/master.perf.txt)
[caller.perf.txt](https://github.com/pinojs/pino/files/781673/caller.perf.txt)

davidmarkclements:
yah it's a nice idea - but I think this should be a plugin - not in core

phra:
@davidmarkclements any ideas how to implement this as a plugin?

jsumners:
I think the simplest solution is just: `log.debug('functionName: this is a log message')`

davidmarkclements:
One potentially elegant way would be to create a child logger with an object that has a caller property, but the property is actually a getter (e.g. defined with Object.defineProperties) - create an error and get the stack in the getter function for caller, and return the relevant call site from the getter. This would result in the same output.

The plugin interface could either be a function that takes a pino instance and returns child logger

davidmarkclements:
using getter for call stack doesn't work - but here you go this will work, feel free to make into a module with tests etc. and we can look at bringing to pino org if you like

```js
function traceCall (pinoInstance) {
  var facade = Object.create(pinoInstance)
  Object.keys(pinoInstance.levels.values)
    .forEach(function (level) {
      facade[level] = makeCallSiteGetter(level)
    })
  function makeCallSiteGetter (level) {
    return function getCallSite () {
      var child = pinoInstance.child({
        caller: Error().stack.split('\n')[3].trim().replace('at ', '')
      })
      return child[level].apply(child, arguments)
    }
  }
  return facade
}
```

example usage
```js
var pino = traceCall(require('./')())
function x() {
  pino.info('hi')
}

x()
```

output: 

```sh
{"pid":25503,"hostname":"NY5-Consult01","level":30,"time":1487342685938,"msg":"hi","caller":"x (/Users/dclements/pino/x.js:20:8)","v":1}
```

phra:
@davidmarkclements thanks, i will do it when i have spare time! 😺

davidmarkclements:
awesome - closing PR and issue - thank @phra :bowtie:

PR #198, opened 2017-03-01
restore previous function when switching level

phra:
in this way a plugin or wrapper can override the default behaviour, also
when changing the log level at runtime.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -104,12 +104,12 @@ function Pino (opts, stream) {
 
 Pino.prototype = new EventEmitter()
 
-Pino.prototype.fatal = tools.genLog(levels.levels.fatal)
-Pino.prototype.error = tools.genLog(levels.levels.error)
-Pino.prototype.warn = tools.genLog(levels.levels.warn)
-Pino.prototype.info = tools.genLog(levels.levels.info)
-Pino.prototype.debug = tools.genLog(levels.levels.debug)
-Pino.prototype.trace = tools.genLog(levels.levels.trace)
+Pino.prototype.fatal = Pino.prototype._fatal = tools.genLog(levels.levels.fatal)
+Pino.prototype.error = Pino.prototype._error = tools.genLog(levels.levels.error)
+Pino.prototype.warn = Pino.prototype._warn = tools.genLog(levels.levels.warn)
+Pino.prototype.info = Pino.prototype._info = tools.genLog(levels.levels.info)
+Pino.prototype.debug = Pino.prototype._debug = tools.genLog(levels.levels.debug)
+Pino.prototype.trace = Pino.prototype._trace = tools.genLog(levels.levels.trace)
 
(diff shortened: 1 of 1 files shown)


jsumners:
@pinojs/core what do you guys think of this? 

I don't think I fully understand how this solves the issue I raised in #191. So I don't want to offer an opinion on it.

mcollina:
@phra I think you should add a unit test, it would clarify the intent of this change.

davidmarkclements:
Here's what I'm seeing:

The underscore prefixed methods add nothing - it's the same as doing `this[key] = Pino.prototype[key]` (because they're two different objects). 

The only actual change is removing the check for supported pino log levels
(`isStandardLevel`) which I assume removes or degrades support for custom log messages

jsumners:
What confuses me is that the `addLevel` tests pass.

davidmarkclements:
I assume by accident though

phra:
good news guys. 🚀 @mcollina @davidmarkclements @jsumners 

i've updated [pino-caller](https://github.com/phra/pino-caller) to work with and without this PR and updated its tests.
imho now it's also more elegant than the previous suggested implementation. 😄 
let me know what you think about this one.

p.s.: to me this PR is also fine, do you have found any bugs?

jsumners:
I like the idea of intercepting `asJson` a lot better. My only concern is that the spread operator requires a harmony flag on LTS v4. So model the way `pino-multi-stream` does the same sort of thing and I think you have a solution -- https://github.com/pinojs/pino-multi-stream/blob/6d5196fe89739b98d816b277840832c814b38518/index.js#L64-L95

Then this PR can be closed. I don't think any of us have found a bug in it, but I believe at least me and @davidmarkclements strongly suspect at least one exists.

phra:
@jsumners sorry, but isn't the `Array.prototype.slice.call(arguments)` trick enough to work on node v4.x.x? 
i fallback to that instead of rest parameters. as you can see now the build is passing now on [travis](https://travis-ci.org/phra/pino-caller/builds/207375668).

so now in my opinion `pino-caller` is ready to be published. waiting for your final approval.

davidmarkclements:
Oh yah nice idea @phra 

It doesn't matter in this case but slice.call on arguments causes deopt on function - that's why we never do that in pino or any non dev libraries 

In @jsumners multi stream module he uses switch with finite arg length - the other way is to do new Array(arguments.length) and then use a for loop to copy args into array - this is the most performant way for non determined arg length, or where there's lots of args - whereas a switch/if statements on arguments.length may be faster in other cases

Closing PR - thanks @phra

phra:
@davidmarkclements i was thinking about the deoptimization problem, i found this very nice [gist](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#3-managing-arguments) that explain it very well.

Issue #206, opened 2017-03-14
[Feature] configure output both by log level and by logger name -dynamically


geonanorch:
I use both `pino` and `pino-debug`. At the moment in a given module I have to:
```
const log = pino.child({ name: '/lib/mymodule' });
const debug = require('debug')('/lib/mymodule');
 . . . .
log.info('some important notice for mymodule');
log.debug('unfiltered debug entry');
 . . . .
debug('filterable debug entry');
```
The 2 `require` above hint at duplication, and even if debug is not enabled for `/lib/mymodule` (`DEBUG=`), the generic `log.info()` and `log.debug()` calls will still output (depending on the logger's level of course).

I understand that `pino-debug`'s goal is to integrate with the `debug` module, therefore I am raising this issue under `pino` itself to suggest:
- implement the same filtering logic for pino loggers and child loggers as in the `debug` module (in other words: orthogonal to the log level, log generation is also filterable by logger name)
- ensure that the current filtering (level+name) can be adjusted at any time by calling a static method with a `map` parameter similar to that of `pino-debug`

Such a feature would:
1. avoid me having to require `debug` for new code, going 100% pino
2. allow to selectively turn on/off logging by modules (or even by function)
3. unify behavior between `pino` and `debug`/`pino-debug`

mcollina:
This is an interesting use case, and the use of env variables keeps coming up.
We have always resisted to this. It's currently implementable with some gymnastic involved, but maybe we could streamline this for pino-debug.

I'm generically against reacting to env variables from a logging lib point of view.

@davidmarkclements @jsumners @watson @marcbachmann any other thoughts?

jsumners:
I abhor environment variables.

@geonanorch for new code, why not just drop `debug`? See https://github.com/jsumners/abstract-logging for an example of a way to support using any logger that isn't `debug`.

jsumners:
As for enabling and disabling logs based on name... I think that would add too much overhead to core. It could easily be done in a transport.

You might also be able to make use of `pino-multistream` for such a scenario, but the transport route is the best way to go.

geonanorch:
@mcollina thanks for the interest. I did not intend to suggest using environment variables, that is something which can easily be implemented by the client code when instantiating pino (sorry, mentioned `DEBUG` because that is what `debug` uses).
[...]

PR #208, opened 2017-03-16
Add logger.level = {Object} support to fine configure child loggers' …

geonanorch:
As discussed in issue #206 an extension of the `logger.level` property to support fine-grained configuration of levels for a family of loggers.

Benchmark results are attached:
  [perf-org.txt](https://github.com/pinojs/pino/files/848062/perf-org.txt)
  [perf-new.txt](https://github.com/pinojs/pino/files/848061/perf-new.txt)


docs/API.md
```diff
@@ -47,9 +47,10 @@
   * `extreme` (boolean): Enables extreme mode, yields an additional 60% performance
     (from 250ms down to 100ms per 10000 ops). There are trade-off's should be
     understood before usage. See [Extreme mode explained](extreme.md). Default: `false`.
-  * `level` (string): one of `'fatal'`, `'error'`, `'warn'`, `'info`', `'debug'`,
-    `'trace'`; also `'silent'` is supported to disable logging. Any other value
+  * `level` (string|object): one of `'fatal'`, `'error'`, `'warn'`, `'info`', `'debug'`,
+    `'trace'`; also `'silent'` is supported to disable logging. Any other string value
     defines a custom level and requires supplying a level value via `levelVal`.
+    Alternatively an object can be provided to specify different levels for (future) child loggers
```
jsumners: Missing a period at the end of the sentence. I'd also like a little more detail. As it is, I really don't know what it even means.

docs/API.md
```diff
@@ -137,8 +138,9 @@ logger.child({ a: 'property' }).info('hello child!')
 Creates a child logger, setting all key-value pairs in `bindings` as properties
 in the log lines. All serializers will be applied to the given pair.
 
-Child loggers use the same output stream as the parent and inherit
-the current log level of the parent at the time they are spawned.
+Child loggers use the same output stream as the parent and inherit the current
+log level of the parent at the time they are spawned (in case an object was set
+for the parent's level, then the log level of the matching rule).
```
jsumners: How about: `... are spawned. If an object was used to define the parent's level, then the the child will use the level specified by a matching rule (see discussion below).`

docs/API.md
```diff
@@ -223,6 +225,11 @@ benchPinoExtremeChildChild*10000: 150.143ms
 ### Example:
 ```
 logger.level = 'info'
+logger.level = {
+    'myobj:*': 'debug',  // child loggers whose 'name' binding matches
+    'myobj:comm:*': 'trace',  // even more info required for a subset
+    '*': 'info'  // this logger's level + default for child loggers
+}
 ```
```
jsumners: Make this a second example, e.g.:

```
logger.level = 'info'
```

or

```
logger.level = { ... }
```

docs/API.md
```diff
@@ -242,6 +249,24 @@ The logging level is a *minimum* level. For instance if `logger.level` is
 
 You can pass `'silent'` to disable logging.
 
+It is possible to pass an object to set different levels to the logger and its
```
jsumners: `... different levels for the logger...`

docs/API.md
```diff
@@ -242,6 +249,24 @@ The logging level is a *minimum* level. For instance if `logger.level` is
...
+children. The syntax for designating the target loggers is borrowed from the
+[debug](https://github.com/visionmedia/debug#wildcards) module's `DEBUG` variable
+(also supported in [pino-debug](https://github.com/pinojs/pino-debug)):
++ use `'*'`(star) as wildcard (rules are ordered so that `'*'` will not override
+`'mylib*'`).
```
jsumners: I'm confused here. Are you suggesting that they read the `debug` module's documentation to learn how to specify configuration for Pino? I'd rather see at least a summary explanation here.

docs/API.md
```diff
@@ -242,6 +249,24 @@ The logging level is a *minimum* level. For instance if `logger.level` is
...
+
+Note that currently the following properties are _not_ supported:
++ multiple patterns separated by `','`(comma) or `' '`(space) &ndash;declare
+  several rules instead
++ exclusion patterns indicated by a leading `'-'`(minus) sign
+
+The match is done against the child logger binding `'name'`, therefore in order
+to use this feature one must instantiate children for example with:
```
jsumners: `... must instantiate children with a name like so:`

pino.js
```diff
@@ -58,6 +58,8 @@ function pino (opts, stream) {
     buf: ''
   }
   iopts.chindings = ''
+  iopts.levelMap = Object(null)
```
jsumners: @mcollina or @davidmarkclements might disagree, but I'd rather see `iopts.levelMap = {}`. I do not like the broken object returned by `Object(null)` (which, in my opinion, should be `Object.create(null)` anyway).

pino.js
```diff
@@ -117,7 +119,7 @@ Object.defineProperty(Pino.prototype, 'levelVal', {
     return this._levelVal
   },
   set: function setLevelVal (num) {
-    if (typeof num === 'string') { return this._setLevel(num) }
+    if (typeof num !== 'number') { return this._setLevel(num) }
```
jsumners: This was here to catch assignments like `logger.level = 'info'` or `logger.level = '30'`. Why have you changed this? Is it integral to the new feature?

jsumners:
I submitted my review incorrectly and can't seem to fix it. Consider them as "requested changes" not just comments.

docs/API.md
davidmarkclements: perhaps show a small example string, and a small example with object

docs/API.md
davidmarkclements: oh haha you - yo already did

docs/API.md
geonanorch: No, I do not want people to think that they need `debug` to use that feature (or to use `pino` for that matter), but felt obliged to credit that project. But you are right it is confusing, I will rephrase to just describe the syntax and add a parenthesis `(inspired by the debug and pino-debug modules)`

pino.js
geonanorch: No problem, I am not religious about this.

pino.js
geonanorch: For one thing I need to handle the case where `num` is an object, so yes for the new feature.
For another I thought that the test was wrong anyway: `setLevelVal()` is only meant to handle numerical values so anything else should be dispatched or thrown.

docs/API.md
jsumners: I'm fine with referencing the `debug` docs. I just want to get a full understanding from  Pino's own docs.

geonanorch:
Thanks for reviewing. I will try to add at least one test and address all comments &ndash;and requested-changes-as-comments

geonanorch:
Apologies, pushed too early, unit test is not ready yet

geonanorch:
@mcollina , I just committed 2 unit tests + a fix for dynamic changes.

Side-note: there is an issue (not related to my change) with the test `'level-change event'` in `levels.test.js`, it causes all subsequent tests to fail with:
```
    not ok 3 - test after end() was called # time=7.746ms
      ---
      stack: |
        EventEmitter.handle (test/levels.test.js:12:7)
        EventEmitter.setLevelVal (pino.js:125:12)
        EventEmitter._setLevel (pino.js:163:17)
        EventEmitter.applyOptions (lib/tools.js:43:8)
        EventEmitter.Pino (pino.js:104:22)
        pino (pino.js:91:16)
        Test.<anonymous> (test/levels.test.js:52:18)
        loop (node_modules/function-loop/index.js:35:15)
      at:
```
I did not try to fix it, do not understand enough about either `tap` or `pino`...

jsumners:
@geonanorch that means you have introduced a bug. Look at the test that is failing. It is the one for setting the level by a number -- https://github.com/geonanorch/pino/blob/0224ab155cb9d55bf01b8b5c020b44baa2333f70/test/levels.test.js#L38

geonanorch:
@jsumners the issue has been there beforehand &ndash;and CI is still happy. You can reproduce the problem _on your current master_ as follow:
  - edit `test/levels.test.js` to move the test `'level-change even'` anywhere else than at the bottom of the file
  - run full test suite through `npm` or even just `node_modules/.bin/tap test/levels.test.js`

Looks like that test is able to damage pino's or tap's state permanently, meaning there is a bug somewhere..Maybe I should have reported this through a separate channel...

mcollina:
@geonanorch can you please open a separate issue on that broken test? If reordering makes your test pass, reordering is good.

jsumners:
@geonanorch sorry about this, but I refactored the core code. You'll need to update your branch to incorporate the changes.

geonanorch:
Ah... That's the problem with actively maintained projetcts... Unless you are just testing my motivation! ;-)
Anyway, changes reapplied and ready for your consideration.

pino.js
```diff
@@ -160,23 +172,33 @@ Object.defineProperty(pinoPrototype, 'child', {
     }
     data = this.chindings + data.substr(0, data.length - 1)
 
+    let levelChild = bindings.level || this.level
+    const level = this.levelMap
+    Object.keys(level).sort(tools.byPrecision).map((pattern) => {
+      const regexp = new RegExp('^' + pattern.replace(/\*/g, '.*?') + '$')
+      if (bindings.name && regexp.test(bindings.name)) levelChild = level[pattern]
+    })
```
mcollina: thiw should be a for loop, otherwise we are losing performance on child creation.

lib/tools.js
```diff
@@ -118,11 +120,24 @@ function genLog (z) {
...
+  var aix = a.indexOf('*')
+  var bix = b.indexOf('*')
+  if (aix === -1 && bix === -1) return 0
+  if (~aix && ~bix) {
+    if (aix > bix) return -1
+    if (aix < bix) return 1
+    return a.length < b.length ? 1 : (a.length === b.length ? 0 : -1)
+  }
+  return ~bix ? -1 : 1
+}
```
geonanorch: done, also restructured it in the hope that it is easier to read.

test/helper.js
```diff
@@ -24,5 +24,16 @@ function check (t, chunk, level, msg) {
   })
 }
 
+function check2 (t, chunk, wanted) {
```
geonanorch: I did not dare replace the original `'check()'` with this new version, but that was the intent... Renamed it to `'checkGen'`

test/helper.js
```diff
@@ -24,5 +24,16 @@ function check (t, chunk, level, msg) {
...
+  t.deepEqual(chunk, Object.assign({
+    pid: pid,
+    hostname: hostname,
+    v: 1
+  }, wanted))
+}
+
 module.exports.sink = sink
 module.exports.check = check
+module.exports.check2 = check2
```
geonanorch: That would cause code duplication, repeating the same `require`s as in `helpers.js`. This new helper allows more concise (and more flexible) coding as the original `check()` and could be used instead for new tests:
```checkGen(t, chunk, wanted)```
instead of:
```check(t, chunk, wanted.level, wanted.msg)```

Comment (mcollina) [inline test/helper.js]: It's easy to support both ways, just check if the 3rd argument is an object or not.

File: pino.js (hunk 9 above)
Comment (geonanorch) [inline pino.js]: You are right, and actually most of the work can be done at the time of setting the `level` property, long before children are created. So now pre-computing the regexps and storing them in an array instead of a map. The array is iterated with a for loop.

File: test/helper.js (hunk 12 above)
Comment (geonanorch) [inline test/helper.js]: Okay, done.

**Comment by davidmarkclements** [comment]
do we have benchmarks for this?

**Comment by jsumners** [comment]
I'm about to publish a module that does this as a transport. I just have to work out a couple more tests.

**Comment by jsumners** [comment]
@geonanorch please check out https://github.com/pinojs/pino-filter . I haven't published it to npmjs.com yet. I want to make sure it works as you expect it to first.

**Comment by geonanorch** [comment]
@davidmarkclements benchmark results were included at the beginning of this issue. Since the same approach is used as for integer levels (logger functions set to noop in a _prior_ configuration call) a performance impact can only occur while creating children (2 internal object properties initialized with empty object/array, and one array iteration &ndash;empty array in your benchmark). And I nevertheless optimized child creation at @mcollina's request.

For this insignificant hit (feel free to verify by running the benchmarks) we get:
* improved coverage of the `debug` module to the point that it is no longer needed (I believe `pino` is only missing the timing info) 
* dynamic logging level configuration possible (unlike `debug`)

Together with the [Bunyan](https://github.com/trentm/node-bunyan) compatibility, I believe that it makes `pino` an excellent cover-all-options solution.

The branch is again in conflict. Let me know if you would like me to reapply the changes again &ndash;this time please only if you are ready to merge it right away, I already spent a lot of time on this as it is.

@jsumners  I had a quick look at the code of [pino-filter](https://github.com/pinojs/pino-filter). While functionally it almost certainly does what was discussed in issue #206  and in this thread, it involves testing the configuration for every logging call, which I reckon is far less efficient. What do the benchmark results look like?
Using signals for config changes is not compatible with window, the approach of [pino-arborsculpture](https://github.com/pinojs/pino-arborsculpture) (also by you!) is imho preferable.

**Comment by davidmarkclements** [comment]
Putting the transform in another process is more about keeping the event loop of the serving process free (and also keeps core as simple as possible, and also means less maintenance of browser pino to keep up)

On the other hand.. I do quite like the functionality as a core API ... 

I don't want to waste your time @geonanorch let's come to a definitive whether we want this in core 

I vote yes 

@jsumners  @mcollina  - yay or nay

File: pino.js [hunk 13]
```diff
@@ -75,6 +75,28 @@ Object.defineProperty(pinoPrototype, 'levelVal', {
 
 Object.defineProperty(pinoPrototype, '_setLevel', {
   value: function _setLevel (level) {
+    if (typeof level === 'object') {
+      // process and save rules for future offspring
+      // rules is an ordered (more specific first) list of [regexp, level]
+      this.levelRules = Object.keys(level).sort(tools.orderByPrecision).map((pattern) => {
```
jsumners: I like arrow functions too, but the code style of the project uses the full `function` declaration style.

pino.js
```diff
@@ -75,6 +75,28 @@ Object.defineProperty(pinoPrototype, 'levelVal', {
...
+        return [new RegExp('^' + pattern.replace(/\*/g, '.*?') + '$'), level[pattern]]
+      })
+      this.level = defaultOptions.level
+      // apply rules to existing offspring
+      for (let j = 0, jlen = this.children.length; j < jlen; j++) {
```
jsumners: I think V8 does the `jlen` optimization itself nowadays. And we don't use `let` in Pino; this should be a `var` declaration.

pino.js
```diff
@@ -160,23 +182,38 @@ Object.defineProperty(pinoPrototype, 'child', {
     }
     data = this.chindings + data.substr(0, data.length - 1)
 
+    const nameChild = bindings.name || this.name
+    let levelChild = bindings.level || this.level
+    // child log level will be overridden if parent rule matches
+    for (let i = 0, ilen = this.levelRules.length; i < ilen; i++) {
```
jsumners: `let` and `ilen` again.

jsumners:
@geonanorch it doesn't really test the configuration for every logging call. It does determine if filters match, but it does so in a short-circuit fashion. As for reloading the config, I'm not a Windows guy so I do not consider it when doing things. But I have heard they have a Linux emulator in Win10 that Node stuff could be run under. Regardless, it's an initial implementation to prove the viability. If it ends up as the solution then it could certainly be improved (with your help?).

@davidmarkclements I leave that up to you and @mcollina. I have been quite clear that I think Pino has all of the features necessary to support this as a transport (and I have proved it so). That said, the benchmarks provided look very nice. Everything except for child creation is consistently below the master branch results. If nothing else, I'd like to see whatever changes caused those reductions to be included.

The questions that have me on the fence, so to speak, are:

1. Does this add too much complexity for maintenance?
2. Does this make things confusing for the user?

I truly do not know the answer to either of those. Thus my position.

satazor:
I've been wanting a similar feature for quite some time.

What I want is to be able to propagate the loglevel of a parent to its children. This feature lets me do that, so +1 by me.

mcollina:
If child creation gets on par with master, I'm +1 for something similar.

> Does this add too much complexity for maintenance?

A bit, but it seems a needed feature. If it does not carry a performance penalty, I think we can support it.

> Does this make things confusing for the user?

I do not like the fact that we accept a `pino.level = {}` property access. I think that should just be the actual level. Maybe we can have `setChildrenLevels()` or something similar.

geonanorch:
Okay, I will reapply the changes to the current master (including latest requests for changes), can you confirm that there are no impending commits?
Re `setChildrenLevels()` no problems, note the  borderline case `setChildrenLevels('*': 'info'), which also sets the current logger's level (`setArborescenceLevels()` would be semantically a better match but too complicated I guess).

mcollina:
@geonanorch note that we need to solve the child creation slowdown before landing, you should probably focus on that first.

geonanorch:
@mcollina I am not aware of any slowdown? I highlighted the non-impact 4 days ago &ndash;and proposed that you guys run the benchmarks yourselves for verification. Is there something specific you would like me to do?

mcollina:
@geonanorch it seems @jsumners report a slowdown in child creation (conception benchmarks) 

> Everything except for child creation is consistently below the master branch results.

this PR:
```
benchPinoCreation*10000: 190.466ms
benchPinoExtremeCreation*10000: 114.567ms
```

master:
```
benchPinoCreation*10000: 183.904ms
benchPinoExtremeCreation*10000: 94.618ms
```

davidmarkclements:
Just for context child creation is particular important for common scenarios like http - where pino-http creates a new child per request

jsumners:
@geonanorch we can't guarantee the repository won't change. PRs are merged according to their readiness, not a "first come first serve" basis.

As for the benchmarks, I was just making an observation based on the provided output. The child creation benchmark was the only one that displayed a discrepancy in the wrong direction. Typically, when we see something like this we will run the benchmark a few consecutive times on each branch to get a better idea of the performance. For example, I have the following script:

```sh
#!/bin/bash

if [ "$1" == "" ]; then
  echo "must supply a benchmark name"
  exit 1
fi

echo '' > /tmp/${1}.bench.txt
exec >> /tmp/${1}.bench.txt

npm run ${1}
npm run ${1}
npm run ${1}
```

geonanorch:
Sorry, I had not understood how important the child creation use case is.
I have merged again, included the requested changed and used the following script to run the benchmark:
```
#!/bin/bash
for i in 1 2 3 4
do
  echo loop $i
  (cd pino.org; npm run bench-conception > bench-conception${i}.log) &
  (cd pino; npm run bench-conception > bench-conception${i}.log) &
  sleep 5
done
```
Results (`pino.org` being your master, `pino` my fork):
```
$ ./bench.sh
loop 1
loop 2
loop 3
loop 4
$ grep 'PinoExtremeCreation average' pino*/bench*.log
pino/bench-conception1.log:PinoExtremeCreation average: 112.2345
pino/bench-conception2.log:PinoExtremeCreation average: 112.7385
pino/bench-conception3.log:PinoExtremeCreation average: 112.209
pino/bench-conception4.log:PinoExtremeCreation average: 113.253
pino.org/bench-conception1.log:PinoExtremeCreation average: 90.58449999999999
pino.org/bench-conception2.log:PinoExtremeCreation average: 90.77850000000001
pino.org/bench-conception3.log:PinoExtremeCreation average: 95.86
pino.org/bench-conception4.log:PinoExtremeCreation average: 96.0865
ronan@axisparc:~ $ grep 'PinoCreation average' pino*/bench*.log
pino/bench-conception1.log:PinoCreation average: 200.8015
pino/bench-conception2.log:PinoCreation average: 198.751
pino/bench-conception3.log:PinoCreation average: 201.96050000000002
pino/bench-conception4.log:PinoCreation average: 199.909
pino.org/bench-conception1.log:PinoCreation average: 185.078
pino.org/bench-conception2.log:PinoCreation average: 188.5795
pino.org/bench-conception3.log:PinoCreation average: 188.68
pino.org/bench-conception4.log:PinoCreation average: 181.06
```
So for child creation it looks like a fixed cost of 10-15, which translates to 5% in normal mode and 20% in extreme mode.
As  you can see in the `child()` method, child creation involves initialization of a few new internal properties and an array push &ndash;the latter being responsible for half the hit. I cannot see anything to improve, the call is yours.

pino.js
```diff
@@ -306,9 +352,3 @@ module.exports.stdSerializers = {
   res: serializers.asResValue,
   err: serializers.asErrValue
 }
-module.exports.pretty = pretty
-Object.defineProperty(
-  module.exports,
-  'LOG_VERSION',
-  {value: LOG_VERSION, enumerable: true}
-)
```
jsumners: Why have you removed these? I know that the log version is defined on the `pinoPrototype` so that removal could be a 👍, but I don't know if there is a use case for it being on the module (@davidmarkclements @mcollina ?).

The export of `pretty`, though, needs to remain. Did I miss that export in your changes? Or is it actually removed?

geonanorch: Apologies, this was not intentional, must be a bad case of mouse select + delete, and as it did not break the tests I did not spot it. I will put that block back in.

mcollina:
I am 👎  on anything that slow things down child creation. I think this is due to the `children.push()` call. Now I also note that, this behaviour would likely leak memory over time, as the children are never removed from the array in the `pino-http` usecase.

geonanorch:
Mmmm.... in that case (children never removed) the feature would have to be conditional with `if (this.childLevelsEnabled) { . . . }` and activated in constructor or separately. That would have the advantage that core performance is even less affected, but maybe a bit awkward for a core functionality.

jsumners:
What are our current thoughts on this? From what I'm reading it seems the consensus may be that solving the memory leak issue will lead to too much complexity in core. If that is the case, should this be closed in favor of improving `pino-filter`?

ping: @geonanorch @davidmarkclements @mcollina

mcollina:
`pino-filter` 👍 .

geonanorch:
If I read the [pino-http](https://github.com/pinojs/pino-http/blob/master/logger.js) code correctly, it is creating a child for _each_ HTTP request; that child includes a copy of the request, that is the _only_ difference to the parent. The child is used a _single_ time (upon request completion).
I do not understand why a child for such a use case, it seems that storing the request and logging it through the parent logger would be enough. The change suggested in this thread seems a more natural use case for children.
I reckon that `pino-filter` will not have nearly as good a performance as the core change proposed here.
At any rate I cannot spend more time on this, do with this issue as you like.
Thank you for regularly providing feedback over the last 3 weeks.

mcollina:
> If I read the pino-http code correctly, it is creating a child for each HTTP request; that child includes a copy of the request, that is the only difference to the parent. The child is used a single time (upon request completion).

This is actually not true. The full reason of `pino-http` is to provide the application for a logger for each request, with a _unique id_. So you can log lines specific to that request, and reconstruct what happened in case of a bug.

davidmarkclements:
@geonanorch thank you for all your work here - this is the first time we've had a stalling PR on this project, I think because there's a lot of moving parts and opinions at play - we really appreciate your patience on this - whether or not we merge or borrow from your work the conversation you've generated is genuinely a massive contribution in itself

jsumners:
Given the abandonment of the OP, and the consensus among @pinojs/core that `pino-filter` is the best route, I'm closing this PR. If someone wants to tackle the issues raised and re-propose it, they are welcome to do so. Otherwise, `pino-filter` needs some TLC so that it can be released.

PR #211, opened 2017-03-18
Refactor construction to avoid `new`

jsumners:
I'm not sure if you guys want this. I just wanted to see how Pino would perform if it didn't rely on the `new` construct. I am attaching single runs of the full benchmark suite from the current master and this refactor branch. It would be nice to know how the runs compare on some other systems. I'll let you be the judge of the results.

There are a lot of changes here. You may want to simply view the full file instead of the diff. I started the work from the issue-209 branch in PR #210.

One nice change here, in my opinion, is that we don't have to expose the `_Pino` property for things like `pino-multi-stream`. Those sort of modules can simply get a new instance out of the factory function like normal and do true JavaScript "inheritance".

[refactor.bench.txt](https://github.com/pinojs/pino/files/852001/refactor.bench.txt)
[master.bench.txt](https://github.com/pinojs/pino/files/852002/master.bench.txt)

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -30,6 +30,199 @@ var defaultOptions = {
   enabled: true
 }
 
+var pinoPrototype = Object.create(EventEmitter.prototype, {
+  silent: {
+    value: tools.noop,
+    enumerable: true
+  },
+  stream: {
+    value: process.stdout,
+    writable: true
+  }
+})
+
+var levelMethods = ['fatal', 'error', 'warn', 'info', 'debug', 'trace']
+levelMethods.forEach(function (m) {
+  Object.defineProperty(pinoPrototype, m, {
+    value: tools.genLog(levels.levels[m]),
+    enumerable: true,
+    writable: true
+  })
+})
+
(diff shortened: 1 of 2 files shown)


mcollina:
benchmarks are good and just a bit better, just double check but it seems the 'object' one is slightly slower than master. Maybe it was just a bad run, but better check.

If that's ok, 👍  for me.

jsumners:
I think it's just variations in CPU scheduling. I am attaching the results of three consecutive runs of `bench-object` on both master and refactor branches. The second run of each shows this. And in each of the three runs the `PinoUnsafeExtremeObj` results are nearly identical.

[bench-object.refactor.txt](https://github.com/pinojs/pino/files/852451/bench-object.refactor.txt)
[bench-object.master.txt](https://github.com/pinojs/pino/files/852452/bench-object.master.txt)

davidmarkclements:
oh very nice - I refactored to use constructor because of perf benefits over closure

but if we can use Object.create (which is fine to use now since we're not trying to support v. old JS) instead of `new` then all the better 👍

PR #213, opened 2017-03-20
Add profiling scripts

jsumners:
The goal of this PR is to add a few profiling scripts that exercise the same things as the benchmark scripts. I'm submitting my initial implementation of the basic profile to get your feedback. I don't think this is something that can be automated, but I do think it would be useful when things are majorly changed.

As an example:

I'm not sure when those crept in, if they were always present, or what. But I do know we should look into why they are showing up.

### DIFF
--- a/docs/profiling.md
+++ b/docs/profiling.md
@@ -1,0 +1,6 @@
+# Profiling Pino
+
+1. Install `0x`: `npm install -g 0x`
+2. Run a profile script: `0x profile/basic.profile.js`
+3. View the resulting flamegraph
+4. Enter "pino" in the search field to view the relevant results
--- a/profile/basic-alt.profile.js
+++ b/profile/basic-alt.profile.js
@@ -1,0 +1,32 @@
+'use strict'
+
+var fs = require('fs')
+var pino = require('../')
+var dest = fs.createWriteStream('/dev/null')
+
+var log = pino(dest)
+
+setTimeout(() => {
+  var i = 0
+  var max = 10000
+
+  // warmup
+  for (i; i < max; i += 1) {
+    log.info('hello world')
+  }
+
+  // iteration 1
+  for (i; i < max; i += 1) {
+    log.info('hello world')
+  }
+
+  // iteration 2
+  for (i; i < max; i += 1) {
+    log.info('hello world')
+  }
+
+  // iteration 3
+  for (i; i < max; i += 1) {
+    log.info('hello world')
+  }
+}, 500)
--- a/profile/basic.profile.js
+++ b/profile/basic.profile.js
@@ -1,0 +1,30 @@
+'use strict'
+
+var fs = require('fs')
+var http = require('http')
+var pino = require('../')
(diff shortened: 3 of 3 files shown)


davidmarkclements:
Good shout @jsumners 

Maybe it can be automated 

@mcollina what if we break out the 0x part of autocannon-ci 
into 0x-ci and used it here?

davidmarkclements:
Oh wait you're doing it in a server context - I think that should 
be done for pino-http but not pino - why add the extra stacks?

jsumners:
I started with several `for` loops but it was too fast (😉) so 0x couldn't trace anything. Putting the for loops in a `setTimeout(500)` worked, but it was a de-opt of its own. So I tried this approach. But that's why I started the PR with so little finished -- to get your help figuring out a good approach.

davidmarkclements:
Oh interesting - same story for setImmediate?

mcollina:
@jsumners I think we can use https://github.com/mcollina/autocannon-ci, if our goal is to generate flamegraphs for HTTP requests.

davidmarkclements:
yah I think that could (and should) be a goal for pino-http

the goal here (as I understand it) is purely to profile pino core - it's just the best way to do that so far is to wrap it in an http server

jsumners:
@davidmarkclements `setImmediate` had the same problem as the basic for loops version -- it completes too quickly for 0x to do anything. FYI, this version of the profile script is also not optimized. But I don't really care about the profile script itself being optimized, so I can easily rewrite it back to using a `setTimeout`.

The goal is to give us a way to, at least manually, verify that core remains as optimized as we can make it.

jsumners:
`node --prof --trace-opt --trace-deopt profile/basic-alt.profile.js 1> trace.out`

[trace.out.txt](https://github.com/pinojs/pino/files/861744/trace.out.txt)
[isolate-0x103000000-v8.processed.log.txt](https://github.com/pinojs/pino/files/861743/isolate-0x103000000-v8.processed.log.txt)

jsumners:
So there are a couple of questions:

1. is this worth the time?
2. which style should be used? The `alt` script I added this morning needed as many iterations as I gave it (5 total) to even register a flamegraph at all.

mcollina:
I don't know if it's worth the time.
I prefer the `alt` version.

davidmarkclements:
Wait. It may be because 0x has a default delay of 300ms to skip module loading stacks and allow for perf and dtrace init 

Trying setting 0x duration to 0

jsumners:
@davidmarkclements `-d 0` allows `basic-alt.profile.js` to act like it is generating data for 0x when the iterations are unwrapped from the `setTimeout`. However, it does not actually generate a flamegraph.

I think I'm going to agree with @mcollina, and say this endeavor is not worth the effort. I think that as long as our benchmarks stay consistent (or even decrease) we are in good shape no matter what optimizations are being done, or not, by V8. So I'm going to close this PR. I'll leave the branch in case we decide otherwise later.

PR #214, opened 2017-03-20
Move functions out of Object.defineProperty

jsumners:
This PR moves functions out of `Object.defineProperty(instance, 'name', {value: func(){}})` so that V8 can eagerly compile them. Work in PR #213 shows that when the functions are inlined, V8 will lazily compile them.

bench-object output:
```
==========
object averages
BunyanObj average: 913.2774999999999
WinstonObj average: 2009.229
BoleObj average: 293.574
LogLevelObject average: 2424.0535
PinoObj average: 270.6385
PinoUnsafeObj average: 266.355
PinoExtremeObj average: 126.708
PinoUnsafeExtremeObj average: 128.728
==========
```

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -50,43 +50,46 @@ levelMethods.forEach(function (m) {
   })
 })
 
-Object.defineProperty(pinoPrototype, 'levelVal', {
-  get: function getLevelVal () {
-    return this._levelVal
-  },
-  set: function setLevelVal (num) {
-    if (typeof num === 'string') { return this._setLevel(num) }
+function getLevelVal () {
+  return this._levelVal
+}
+function setLevelVal (num) {
+  if (typeof num === 'string') { return this._setLevel(num) }
 
-    if (this.emit) {
-      this.emit('level-change', this.levels.labels[num], num, this.levels.labels[this._levelVal], this._levelVal)
-    }
-
-    this._levelVal = num
-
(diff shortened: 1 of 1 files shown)


jsumners:
For some reason unbeknownst to me, this does not resolve `asJson` being lazily compiled. It is fixing the `write` method, though.

davidmarkclements:
Beautiful - can we see comparison benchmarks to master 

I see this pattern a lot in core - instead of assigning to function expression, assigns to fn statement reference so this totally makes sense 

Also this is a really nice win because I prefer this pattern 

When you say "not lazy compiled" you mean v8 is able to optimise? Can use --trace-opt to see reason for asJson - but may not be possible to opt, something's aren't

jsumners:
It's possible I'm just trying to clear things in the flamegraph and not really accomplishing anything.

[bench-object.master.txt](https://github.com/pinojs/pino/files/856676/bench-object.master.txt)
[bench-object.fix-deopts.txt](https://github.com/pinojs/pino/files/856677/bench-object.fix-deopts.txt)

mcollina:
I prefer this approach, even though I do not think it would yield any performance improvement after a while. Maybe it helps V8 figure out what to optimize, but it gets there anyway.

davidmarkclements:
I was going to say the same thing

PR #218, opened 2017-03-21
browser write hook

davidmarkclements:
* introduces object based logging that corresponds to pino logs
* object based logging can be turned on independently with `asObject` option
* supply an individual write function to capture all logs
* supply an object to capture logs on level by level basis
* if an object, any methods that aren't supplied fallback to logging to console
* `asObject` is mandatorily enabled if `write` is supplied

cc @cyberthom @temsa

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1,16 +1,27 @@
 'use strict'
 
+var format = require('quick-format-unescaped')
+
 module.exports = pino
 
 var _console = global.console || {}
-function Pino () {}
-Pino.prototype = _console
 
 function pino (opts) {
   opts = opts || {}
+  opts.browser = opts.browser || {}
+  var proto = opts.browser.write || _console
+  if (opts.browser.write) opts.browser.asObject = true
+
+  var levels = ['error', 'fatal', 'warn', 'info', 'debug', 'trace']
+
+  if (typeof proto === 'function') {
+    proto.error = proto.fatal = proto.warn =
+    proto.info = proto.debug = proto.trace = proto
+  }
+
   var level = opts.level || 'info'
   var val = pino.levels.values[level]
-  var logger = new Pino()
+  var logger = Object.create(proto)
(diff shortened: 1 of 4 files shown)


browser.js
```diff
@@ -44,10 +56,31 @@ function pino (opts) {
...
+        o.msg = msg
+        write.call(logger, o)
+      }
+    })(logger[k], k)
+  }
   return logger
 }
 
-pino._Pino = Pino
+// pino._Pino = Pino
```
jsumners: Why retain this line at all?

browser.js
```diff
@@ -44,10 +56,31 @@ function pino (opts) {
     return new Child(this)
   }
   logger.levels = pino.levels
+  return !opts.browser.asObject ? logger : asObject(logger, levels)
+}
+
+function asObject (logger, levels) {
+  var k
+  for (var i in levels) {
```
marcbachmann: `levels` is an array in here.
It looks a bit strange that you're using a normal for loop everywhere else except here.

browser.js
```diff
@@ -1,16 +1,27 @@
...
-function Pino () {}
-Pino.prototype = _console
 
 function pino (opts) {
   opts = opts || {}
+  opts.browser = opts.browser || {}
+  var proto = opts.browser.write || _console
+  if (opts.browser.write) opts.browser.asObject = true
+
+  var levels = ['error', 'fatal', 'warn', 'info', 'debug', 'trace']
```
marcbachmann: I guess you can move that array into the `asObject` function. You don't need it anywhere else.

mcollina: I do not want to allocate this every time we call `asObject`.

browser.js
mcollina: good spot!

browser.js
davidmarkclements: accident

PR #226, opened 2017-03-24
add forceColor option to pino pretty pipe

atis--:
Sometimes you want color even if the output does not go directly to a tty.

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -65,7 +65,7 @@ function pretty (opts) {
 
   stream.pipe = function (dest, opts) {
     ctx = new chalk.constructor({
-      enabled: !!(chalk.supportsColor && dest.isTTY)
+      enabled: !!(chalk.supportsColor && dest.isTTY || opts && opts.forceColor)
     })
 
     levelColors = {


davidmarkclements:
Hey @atis I think this is definitely worth doing, couple of things 

1. We'll also need to hook this up with the CLI in bin.js, I think -c should be fine for forcing color
2. Can you please document the flag in docs/cli.md ?
3. Can you also please add it to usage.txt
4. The pino.pretty api doc will also need to be updated, https://github.com/pinojs/pino/blob/master/docs/API.md#pretty
5. Can you add test(s) for the option in tests/pretty.test.js
6. The build is failing because of lint rules, we use standard, it should have ran as a precommit hook (maybe it did and you fixed but forgot to git add)

atis:
@davidmarkclements i think you got the wrong mention. OP is @atis--

jsumners:
@davidmarkclements he was able to commit it because of https://github.com/observing/pre-commit/issues/99

davidmarkclements:
Sorry @atis! 

@jsumners ah ok seems like piping is the thing to do here

jsumners:
@atis-- are you willing to address the issues raised by @davidmarkclements ?

atis--:
Sorry, but I'm a bit swamped with other things right now. Also this may be easier/faster for someone who's familiar with the code base. Btw the pre-commit hook didn't run because I had not "npm installed" the dependencies. Oh well.

sandyplace:
If anyone hasn't tackled this yet. I essentially have implemented the same option. I can submit a pull request with the 6 items outlined by @davidmarkclements if you are interested.

jsumners:
@sandyplace please do.

davidmarkclements:
yay

PR #233, opened 2017-04-15
Configurable `messageKey` for pino and pretty

### LINKED ISSUE #231: configurable `msg` property name?
This looks like a great tool, and I'm eager to use it on an app that I'm using with Stackdriver on Google Cloud.

Stackdriver plays nicely with JSON logs, but will automatically use a `message` property in the JSON object to label the log output, while Pino favors a hard-coded `msg` prop for similar functionality - would there be interest in making `msg` configurable or should I just manually assign `message` in the object I pass to the logger?

The downside of the latter option is that I wouldn't get the highlighted `msg` string in the pretty-printed logs in dev.

mmcgahan:
Fixes #231 

I haven't added docs or tests for this specific feature, but wanted to share the code for my first pass. If it looks reasonable, I'll go ahead and flesh out the rest.

---

There doesn't appear to be any noticeable effect on performance - `runbench` times vary a bit but seem to be about the same as your current `master` code

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -32,7 +32,7 @@ function messagePropArg () {
     return
   }
   var messagePropIndex = process.argv.indexOf('-m') + 1
-  var messageProp = process.argv.length <= messagePropIndex &&
+  var messageProp = process.argv.length > messagePropIndex &&
     process.argv[messagePropIndex]
 
   if (!messageProp) {
--- a/bin.js
+++ b/bin.js
@@ -13,7 +13,8 @@ if (arg('-h') || arg('--help')) {
 } else {
   process.stdin.pipe(pretty({
     timeTransOnly: arg('-t'),
-    levelFirst: arg('-l')
+    levelFirst: arg('-l'),
+    messageProp: messagePropArg()
   })).pipe(process.stdout)
 }
 
@@ -24,4 +25,18 @@ function usage () {
 
 function arg (s) {
   return !!~process.argv.indexOf(s)
+}
+
+function messagePropArg () {
+  if (!arg('-m')) {
+    return
+  }
+  var messagePropIndex = process.argv.indexOf('-m') + 1
+  var messageProp = process.argv.length <= messagePropIndex &&
+    process.argv[messagePropIndex]
+
+  if (!messageProp) {
+    throw new Error('-m flag provided without a string argument')
+  }
+  return messageProp
 }
--- a/bin.js
+++ b/bin.js
@@ -14,7 +14,7 @@ if (arg('-h') || arg('--help')) {
   process.stdin.pipe(pretty({
     timeTransOnly: arg('-t'),
     levelFirst: arg('-l'),
-    messageProp: messagePropArg()
+    messageKey: messageKeyArg()
   })).pipe(process.stdout)
 }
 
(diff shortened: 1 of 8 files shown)


pino.js
```diff
@@ -121,7 +122,7 @@ function asJson (obj, msg, num) {
   // to catch both null and undefined
   /* eslint-disable eqeqeq */
   if (msg != undefined) {
-    data += ',"msg":' + JSON.stringify('' + msg)
+    data += ',"' + this.messageProp + '":' + JSON.stringify('' + msg)
```
mcollina: can we just generate the encoded version, like we had ',"msg":' before?

Also, the messageProp would need to be encoded for " and the like.

pretty.js
```diff
@@ -56,6 +56,8 @@ function pretty (opts) {
   var timeTransOnly = opts && opts.timeTransOnly
   var formatter = opts && opts.formatter
   var levelFirst = opts && opts.levelFirst
+  var messageProp = opts && opts.messageProp
```
mcollina: this should be documented in the CLI tool

mmcgahan: Added a `-m <key>` option for the CLI tool. It's a bit heavier than the `-t` and `-l` flags but hopefully simple enough to use

It doesn't support CLI arg formats like `-m<key>` or `-m=<key>` but without adding some heavy arg-parsing dependencies I think that's the best solution.

pino.js
```diff
@@ -121,7 +122,7 @@ function asJson (obj, msg, num) {
   // to catch both null and undefined
   /* eslint-disable eqeqeq */
   if (msg != undefined) {
-    data += ',"msg":' + JSON.stringify('' + msg)
+    data += ',' + JSON.stringify('' + this.messageProp) + ':' + JSON.stringify('' + msg)
```
mcollina: the first concatenation can (and should) be cached in a variable.

bin.js
```diff
@@ -13,7 +13,8 @@ if (arg('-h') || arg('--help')) {
 } else {
   process.stdin.pipe(pretty({
     timeTransOnly: arg('-t'),
-    levelFirst: arg('-l')
+    levelFirst: arg('-l'),
+    messageProp: messagePropArg()
```
davidmarkclements: can we consider `messageLabel`

davidmarkclements: or maybe `msgKey`

jsumners: I like `msgKey`.

pino.js
```diff
@@ -258,6 +259,7 @@ function pino (opts, stream) {
   // internal options
   iopts.stringify = iopts.safe ? stringifySafe : JSON.stringify
   iopts.formatOpts = {lowres: true}
+  iopts.messageKeyString = ',' + JSON.stringify('' + iopts.messageKey) + ':'
```
jsumners: Are you anticipating someone using a key that would break simply doing: `iopts.messageKeyString = `,"${iopts.messageKey}":`?

mmcgahan: No, just following the idiom of the [current line formatting](https://github.com/pinojs/pino/pull/233/files#diff-04e6fb630166e65e0231f8c63a731aa0L124) as [suggested](https://github.com/pinojs/pino/pull/233#discussion_r111674233). String templates seem like they could tighten up a few things in this code but I figured I shouldn't introduce syntax that wasn't already used elsewhere - particularly if there are Node version compatibility concerns.

Happy to change if there's interest, though!

mmcgahan: Actually it looks like there's a string template in lib/tools.js, so I guess that's a green light - I prefer the string template solution so I'll make that change and then we should be good to go

mcollina:
We are node 4+, so template strings are now ok.
Check if there are no performance regressions and go for them.

mmcgahan:
Performance appears unchanged - `basic`, `child`, `grandchild`, and `conception` all report similar numbers +/- 10 ms on repeated runs.

PR #235, opened 2017-04-17
Add property to expose Pino version

jsumners:
This PR exposes the current version as a property on the logger. Since we don't expose the actual prototype, and it really can't be exposed, we can't use `pinoPrototype.isPrototypeOf(instance)` to check if a logger instance is an instance of Pino. This PR gives a rudimentary way to validate instances.


jsumners:
Merged. Will wait for #233 before release.

Issue #202, opened 2017-03-07
time plugin system


davidmarkclements:
Just released http://npm.im/fast-date

It means we can get full UTC timestamps with with performance close to Date.now()

```
NativeDate*100000: 215.076ms
FastDate*100000: 124.236ms
DateNow*100000: 123.043ms
```

It grabs this functionality from internal HTTP code, which means that it's using the same date cache as the HTTP module does (for generating Date headers) see https://github.com/davidmarkclements/fast-date#how and https://github.com/davidmarkclements/fast-date#more-benchmarks

The caveat is - to use the same cache, you have to load fast-date (and therefore pino) **BEFORE** you the http module is loaded, otherwise it'll use a second cache which is still performant but less so. 

The question is how do we want to integrate.. 

1. Add fast utc time as an option
2. Create v2 pino log output, make fast utc the default, of milliseconds and hrtime tuple output as other options - change the slowtime option etc - release new major pino version, backport it to legacy branch

I prefer option 2 even though it's much more work it means we have higher parity with bunyan/winston/bole log formats for very slight cost

jsumners:
I don't like the thought of loading the `http` module just to get timestamping in a logger. Not everything Pino gets used for is exposed to the web, or even directly uses `http`.

davidmarkclements:
yeah fair point - it doesn't load the whole http module, just a piece
and if you really wanted to avoid that - you could do require('fast-date/fallback') (we could rename) to get the fallback implementation

jsumners:
And after reading that it is using an *internal* `http` module API, I'm even less warm to the idea. I don't think the roughly 3% improvement in timestamping is worth relying on an internal API.

davidmarkclements:
Not relying on it - if it fails, we have a fallback that is still quite fast

davidmarkclements:
also it's a ~57% improvement isn't it? (124/215)

mcollina:
so, I'm 👎 on having pino depends on private APIs from core.
I'm 👍  on refactoring pino, so that we can plug that in as an addon.

Something like:

```js
pino({
  now: require('fast-date')
})
```

jsumners:
Fastdate avg reqs/s: 19103.28
Native avg reqs/s: 18582.8

18582.8 / 19103.28 = 0.972757 => 1- 0.972757 = 0.027243 => ~3% difference

Did I math wrong?

davidmarkclements:
[...]

PR #239, opened 2017-04-29
Implement pluggable time functions

jsumners:
This PR implements the time function plugin system discussed in #202 . It also removes the `slowtime` option.  Thus, this should necessitate a major version bump.


docs/API.md
```diff
@@ -38,13 +42,12 @@
...
-  * `slowtime` (boolean): Outputs ISO time stamps (`'2016-03-09T15:18:53.889Z'`)
-    instead of Epoch time stamps (`1457536759176`). **WARNING**: This option
-    carries a 25% performance drop. We recommend using default Epoch timestamps and transforming logs after if required. The `pino -t` command will do this for
-    you (see [CLI](cli.md)). Default: `false`.
+  * `timestamp` (boolean|function): Enables or disables the inclusion of a timestamp in the
+    log message. If a function is supplied, it must synchronously return a JSON string
+    representation of the time, e.g. `"time":1493426328206 (which is the default).
+    If set to `false`, no timestamp will be included in the output.
+    See [stdTimeFunctions](#stdTimeFunctions) for more information.
+    Caution: any sort of formatted time will significantly slow down Pino's performance.
```
mcollina: it's not exactly clear that a user should pass the functions defined in the pino modules.

jsumners:
@mcollina note that @davidmarkclements wanted to deprecate `slowtime` -- https://github.com/pinojs/pino/issues/202#issuecomment-285053264 . I don't think it's worth keeping around for some indeterminate amount of time, which will likely lead to it never getting removed. So I went ahead and fully deprecated it.

docs/API.md
jsumners: I updated this text in 7455cd9

davidmarkclements:
we could output a deprecation notice, and remove in a the next major

mcollina:
Given the amount of churn needed for a semver-major release, I think the best option is to deprecate the option, but keep supporting it until we have to bump for some other reason.
We have a lot of dependents now, so we should be more careful than in the past.

What do the others thinks? I'm ok if you all want to bump the major: I just want to be lazy and not update my deps

davidmarkclements:
Do agree

jsumners:
I have added the `slowtime` option back in. It emits a deprecation warning when used. I also created a project for tracking such things for v5.0.0 -- https://github.com/pinojs/pino/projects/1

PR #241, opened 2017-05-02
Clarify serializers further

AlexTes:
This commit removes wrong wording, clarifies the matching happens based on the key and explains we match only for top-level properties of logged objects.

Let me know how you feel about the wording. I'd be happy to tweak it a little further.

(Sick response time you got there @mcollina 🕙😉! and I'm benchmarking if those single letter vars in `genLog` really matter, if they don't I'm coming back one more time 😛!)

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -40,8 +40,8 @@
   * `name` (string): the name of the logger. Default: `undefined`.
   * `serializers` (object): an object containing functions for custom serialization
     of objects. These functions should return an JSONifiable object and they
-    should never throw. Each property name of this object will match to a
-    property name is logged objects.
+    should never throw. When logging an object, each top-level property matching the exact key of a serializer 
+    will be serialized using the defined serializer.
   * `timestamp` (boolean|function): Enables or disables the inclusion of a timestamp in the
     log message. If a function is supplied, it must synchronously return a JSON string
     representation of the time, e.g. `,"time":1493426328206 (which is the default).


jsumners:
I'm not a fan of the single letter variable names either, but for the most part it's because there isn't a name to assign. Most of those variables are to cover the variable number of parameters that can be passed as values for format string substitutions. This a simple optimization instead of always looping over the arguments object.

mcollina:
Regarding the `genLog` single-letters-variables, they are there for a reason: they actually _mean_ "first, second, third, ...", **but** because the length of function counts in checking if it can be optimized or not, they are the quickest shorter alias.

AlexTes:
Maybe I'm reading things wrong, but the 'a' through 'k' variable make total sense to me knowing pino has a maximum number of log values it takes on. But that's 
> Most of the variables [..]

like jsumners mentions. mcollina seems to talk about the same. I'm interested in the rest. 

Why 'z' instead of 'level'. I had to look up invocations to figure this out. Alternatively, one could hunt for the third argument to `pinoWrite`, and find through `asJson` that the `lscache` has to do with logging levels.

In any case 'l', 'm', 'n', 'o', 'p' are not just values to log, they have special meaning. I already have names in mind for a few of them. Can't promise anything but, how about I open a PR where I use naming and comments to express the semantics of the code better? And if you guys go: 'but the performance 😲!' I'd love to try and see about getting some benchmarks that hopefully reward my strong faith in V8 😛.

davidmarkclements:
Hahaha IT WAS I! Who did this dastardly thing with the params. 

We used arguments object, at first but we found this "11 inputs" approach to be faster but we also
found that unless you squeeze that function under a certain amount of chars (512 I think) 
*including* function signature, whitespace AND comments then V8 refuses to inline it

Having said that, I'm all for finding a clearer way without exceeding the limit

Also, as we move tpwards turbofan in node 8 this limit may no longer be a constraint,
however, if we plan tp support 4 and 6 for a while it will still be years before it's not a constraint *to us*

jsumners:
> if we plan [to] support 4 and 6 for a while it will still be years before it's not a constraint to us

My vote is to only support things that have the "LTS" badge, or newer.

AlexTes:
@davidmarkclements you sly person you! Interesting, I didn't know about the small function inlining. Again, I'll see if I can find the time to run a benchmark and look at the effects, if things look interesting at all you'll hear from me.

Sidenote: I think this PR can be merged.

PR #245, opened 2017-05-26
Use _headers instead of _header for the http res serializer.

mcollina:
Fixes https://github.com/pinojs/pino-http/issues/32.

Unfortunately, I think this is semver major. What do we want to do?

### DIFF
--- a/lib/serializers.js
+++ b/lib/serializers.js
@@ -13,7 +13,7 @@ function asReqValue (req) {
 function asResValue (res) {
   return {
     statusCode: res.statusCode,
-    header: res._header
+    headers: res._headers || {}
   }
 }
 
--- a/test/http.test.js
+++ b/test/http.test.js
@@ -105,7 +105,9 @@ test('http response support', function (t) {
       v: 1,
       res: {
         statusCode: originalRes.statusCode,
-        header: originalRes._header
+        headers: {
+          'x-pino': 42
+        }
       }
     })
     cb()
@@ -113,6 +115,7 @@ test('http response support', function (t) {
 
   var server = http.createServer(function (req, res) {
     originalRes = res
+    res.setHeader('x-pino', '42')
     res.end('hello')
     instance.info(res, 'my response')
   }).listen(function (err) {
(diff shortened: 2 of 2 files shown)


jsumners:
You do realize `response.getHeaders()` exists?

mcollina:
> You do realize `response.getHeaders()` exists?

Yes! I LGTM it: https://github.com/mscdex/io.js/commit/af370928f23a5209298d481e6959d5e058c27a4b. Unfortunately it is in Node v7.7.0+.

jsumners:
Damn! I was just looking at the docs in Dash (it pulls in the latest).

jsumners:
Given that it is definitely an API change, it should be a major bump.

mcollina:
How do you feel about bumping semver-major for this? 

We can do it semver-minor by _not removing `header`_, and just `headers`. However, it seems too expensive from a call perspective.

jsumners:
Oh, I wasn't paying attention to which repository I was looking at. I thought I was looking at `pino-http`. So we have the possibilities:

1. remove `header` replacing it with `_headers` [API change for major]
2. keep `header` and add `_headers` along with more overhead per request [minor compatible]
3. keep `header`, add `_headers`, and add an option to pick which one to use [minor compatible]

Options 2 and 3 add a new item to track in project 1. Option 3 adds more complexity to the initial configuration.

I suppose my preferred order among those options would be: 1, 3, 2. So if @davidmarkclements agrees, then this PR can be merged and then another to close out project 1 prior to releasing v5.0.0.

davidmarkclements:
But why though? 

Didn't we deliberate go with the _header string for perf? What's the benchmark diff?

mcollina:
We did `_header`  to do to what Bunyan was doing. I'm not convinced either we should change this, if we should print `_header`  or `_headers` at all, or something else.

davidmarkclements:
Are we agreed that performance should come first? 

If so the decision is easy: which is fastest? (this may actually change in a few months of course...) 

On the other hand, there's a bigger picture - there's demand for a headers object in`pino-http`, 
and grabbing the headers object in core is better than parsing the headers string in `pino-http`

Then again, not everyone needs the headers object.

So: what if we make it configurable : headers string or headers object, users choice. Default to the fastest (string)

mcollina:
@davidmarkclements there is also a security angle, not providing _any_ headers would remove any potential cookie leaks, and it would be faster.
So, we should leave this as is, and change `pino-http`. Do we all agree?

davidmarkclements:
if security angle, and not providing *any* headers is best, then we should change PR to remove headers from core no?

mcollina:
@davidmarkclements probably yes, but we'd break bunyan compat: https://github.com/trentm/node-bunyan/blob/master/lib/bunyan.js#L1113.

davidmarkclements:
ah - so would changing header to headers

what if we turn it off by default, and allow it to be turned on

are we mature enough to neglect a little bit of parity with bunyan?

mcollina:
@davidmarkclements that's my idea. Where all those serializers should live. Here or in pino-http?

jsumners:
I think if it has to do with an HTTP request then it should be in an HTTP specific module. Pino should be a logger and that's it. 

Or, there could be a module of standard serializers, which might include more than we currently offer, that anyone (pino, pino-http, etc) could use.

davidmarkclements:
except that, for *best* security we should turn headers off here as well

if we turn them off, that's a major, and if anyone's relying on it, we need to supply a flag to turn headers back on 

so.. that leads me towards having a flag in core

mcollina:
I think we should just do the base case here (security and performance), and then anyone can provide their own serializer anyway. We can document how to do it in the docs.

jsumners:
bump: @mcollina @davidmarkclements

mcollina:
Are we all of the opinion to remove `headers` from the standard serializers? It might be __better__ than this.

jsumners:
Yes. Issue #265 settled on that and is just waiting for someone to do the work.

mcollina:
Then closing.

PR #248, opened 2017-05-29
Created a package-lock.json file

mcollina:
As introduced by NPM@5.

Some more details on this are available: https://github.com/npm/npm/pull/16441

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6044 @@
+{
+  "name": "pino",
+  "version": "4.5.2",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abbrev": {
+      "version": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
+      "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU=",
+      "dev": true
+    },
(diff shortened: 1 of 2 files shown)


jsumners:
I don't understand the point of this file. That's what the package.json is for, right?

mcollina:
It will be created by npm@5 everywhere, it's the same as yarn.lock, but compatible with the NPM ecosystem.
On the good side, it adds a level of checks on the security of our modules with committed shas.

I'm fine _not_ adding it, but then we will `.gitignore` it.

davidmarkclements:
From the issue, I understand that it's similar to npm-shrinkwrap with a main difference being npm-shrinkwrap get's published to npm, whereas package-lock.json does not

Since we never saw fit to use npm-shrinkwrap, not sure of the benefits of package-lock.json

On a significantly sized application with lots of deps, it could be beneficial - but is it going to just add unnecessary maintenance overhead here?

jsumners:
I don't agree with NPM 5's automatic addition of this file nor its automatic saving of dependencies. Thus, I don't agree with adding this file to the project.

mcollina:
Fantastic. I'll add it to `.gitignore` then.

PR #251, opened 2017-06-01
using this.stringify instead of JSON.stringify

doron2402:
typo

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -1,6 +1,7 @@
 'use strict'
 
 var EventEmitter = require('events').EventEmitter
+var stringifySafe = require('fast-safe-stringify')
 var fs = require('fs')
 var pump = require('pump')
 var flatstr = require('flatstr')
@@ -266,7 +267,7 @@ function pino (opts, stream) {
   }
 
   // internal options
-  iopts.stringify = iopts.safe ? require('fast-safe-stringify') : JSON.stringify
+  iopts.stringify = iopts.safe ? stringifySafe : JSON.stringify
   iopts.formatOpts = {lowres: true}
   iopts.messageKeyString = `,"${iopts.messageKey}":`
   iopts.end = ',"v":' + LOG_VERSION + '}\n'
--- a/pino.js
+++ b/pino.js
@@ -1,6 +1,5 @@
 'use strict'
 
-var stringifySafe = require('fast-safe-stringify')
 var EventEmitter = require('events').EventEmitter
 var fs = require('fs')
 var pump = require('pump')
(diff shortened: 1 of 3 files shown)


mcollina:
Can you add a unit test for this?
What's the problem this is causing you?

jsumners:
I must admit, I'm confused by this as well. You are claiming this is a typo, but Pino's prototype chain is `EventEmitter => Function => Object`. Thus, `.stringify` isn't even a member of the Pino object.

doron2402:
@mcollina 
- I can add a unit test.
- no problem, just don't want to use `JSON.stringify`

@jsumners 
 - just trying to keep it more consistency,  it uses ` this.stringify` in line 138 and 143.

jsumners:
Okay, I forgot about https://github.com/pinojs/pino/blob/e6fbe72f14b9683abcbff95039c4ceb4db89021d/pino.js#L270

I agree this PR should proceed.

doron2402:
@jsumners 
```
Object.defineProperty(pinoPrototype, 'asJson', {
  enumerable: true,
  value: asJson
})
```

jsumners:
@doron2402 what is your point?

doron2402:
to your comment: Thus, .stringify isn't even a member of the Pino object.

jsumners:
@doron2402 the code you quoted doesn't actually impart the `.stringify` method. It's the line I linked in my previous response that does. I had merely forgot we have that option.

doron2402:
should the test live under `test/escaping.test.js` ?

mcollina:
@doron2402 would you mind adding that unit test? We need it to merge this.

jsumners:
@doron2402 since @mcollina didn't answer your question, put the test wherever you think it makes sense.

mcollina:
I think `test/basic.test.js` is ok.

doron2402:
cool will do, I'll add it 
thanks @jsumners @mcollina

pino.js
```diff
@@ -267,7 +266,7 @@ function pino (opts, stream) {
   }
 
   // internal options
-  iopts.stringify = iopts.safe ? stringifySafe : JSON.stringify
+  iopts.stringify = iopts.safe ? require('fast-safe-stringify') : JSON.stringify
```
mcollina: can you move this back to the top?

test/basic.test.js
```diff
@@ -442,3 +443,25 @@ test('children with same names render in correct order', function (t) {
 
   root.child({a: 1}).child({a: 2}).info({a: 3})
 })
+
+//  https://github.com/pinojs/pino/pull/251 - use this.stringify
```
mcollina: there is a double space here

doron2402: 💯

pino.js
doron2402: 👍

doron2402:
done!

Thanks @mcollina

PR #254, opened 2017-06-07
Fix crash in request serializer when behind proxy

joakimbeng:
When used behind a proxy the `req.connection` is not always set which throws an error in the serializer. So this change makes sure we're not trying to log properties on an undefined object.


mcollina:
Can you please add a unit test for this?

joakimbeng:
There you go! :)

PR #255, opened 2017-06-07
Added basic multistream support

mcollina:
This is a work-in-progress. If we need to go forward, it would require more work to iron out all the details. This is just to check with you if it is something we might want.

MultiStream support is shared by all pino instances, and it allows to drop log lines after the fact.
You will require to enable all the log levels down the lowest one you want to log to. Instead of wrapping multiple pino instances like pino-multi-stream does, it adds a "metadata" protocol between pino and the target stream, so that it knows at which level the log line was logged into.

These are the numbers from pino-multi-stream:

```
benchBunyanTen*10000: 5050.589ms
benchPinoMSTen*10000: 5530.074ms
benchPinoNewTen*10000: 4588.214ms
benchBunyanFour*10000: 2166.170ms
benchPinoMSFour*10000: 1525.052ms
benchPinoNewFour*10000: 1352.454ms
benchBunyanOne*10000: 673.961ms
benchPinoMSOne*10000: 357.467ms
benchPinoNewOne*10000: 271.021ms
```

PinoNew is the code in this PR. To note that with one single destination, I get the same numbers with a normal stream.

### DIFF
--- a/test/multistream.test.js
+++ b/test/multistream.test.js
@@ -1,0 +1,39 @@
+'use strict'
+
+const test = require('tap').test
+const pino = require('..')
+const sink = require('./helper').sink
+
+test('baisc multistream', function (t) {
+  t.plan(2)
+
+  const multistream = pino.multistream()
+  const instance = pino({ level: 'debug' }, multistream)
+
+  multistream.add(30, sink(function (chunk, enc, cb) {
+    delete chunk.time
+    delete chunk.pid
+    delete chunk.hostname
+    t.deepEqual(chunk, {
+      level: 30,
+      msg: 'info log line',
+      v: 1
+    })
+    cb()
+  }))
+
+  multistream.add(20, sink(function (chunk, enc, cb) {
+    delete chunk.time
+    delete chunk.pid
+    delete chunk.hostname
+    t.deepEqual(chunk, {
+      level: 20,
+      msg: 'debug log line',
+      v: 1
+    })
+    cb()
+  }))
+
+  instance.info('info log line')
+  instance.debug('debug log line')
+})
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -1,0 +1,33 @@
+'use strict'
+
+function MultiStream () {
+  if (!(this instanceof MultiStream)) {
+    return new MultiStream()
+  }
+
+  this.streams = new Map()
+
+  // enable metadata on pino
+  this[Symbol.for('needsMetadata')] = true
+}
+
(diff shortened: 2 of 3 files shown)


test/multistream.test.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const test = require('tap').test
+const pino = require('..')
+const sink = require('./helper').sink
+
+test('baisc multistream', function (t) {
+  t.plan(2)
+
+  const multistream = pino.multistream()
```
jsumners: I think people are going to get linter errors if this API ships as-is. The `lib/multistream.js` should export a simple factory function that returns an `Object.create()` of an object literal. Or am I missing some technicality that would prevent such?

davidmarkclements:
Amazing work @mcollina - awesome stuff on the benchmarks

I think that we should keep this outside of core though, and still recommend the `pino-tee` approach for most people who don't have constraints that require it multi streaming to be done in process

If we expose the `pinoPrototype` object,  and change 

```js
Object.defineProperty(pinoPrototype, 'write', {
   value: pinoWrite
 })
```

to 

```js
Object.defineProperty(pinoPrototype, 'write', {
   value: pinoWrite,
    configurable: true
 })
```

We can then create new major of pino-multi-stream that replaces the `pinoWrite` function

This was the documentation for the `multistream` function that the `needMetadata` symbol will be clearer for those who need in-process multistreaming

Also - I think we should direct users to to `pino-tee` in `pino-multi-stream` in the interest of championing a best practice for normal constraints

test/multistream.test.js
```diff
@@ -0,0 +1,39 @@
...
+      level: 20,
+      msg: 'debug log line',
+      v: 1
+    })
+    cb()
+  }))
+
+  instance.info('info log line')
+  instance.debug('debug log line')
+})
```
davidmarkclements: would love to see a test for child loggers with multistream - just to see what that looks like

test/multistream.test.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const test = require('tap').test
+const pino = require('..')
+const sink = require('./helper').sink
+
+test('baisc multistream', function (t) {
```
davidmarkclements: basic*

jsumners:
Exposing the prototype object would not be enough @davidmarkclements . The factory function we export does a lot more work on top of just `Object.create(pinoPrototype)`.

mcollina:
This is absolete, and part of pino-multi-stream.

PR #258, opened 2017-06-08
Added metadata protocol

### LINKED ISSUE #249: Support multiple output streams
Hi there,

Does Pino support multiple output streams? I took a look at the API and source, and it looks like currently Pino only supports one output stream. We have use cases where we'd like to output the logs both to stdout and HTTP, and multiple output streams, similar to what [bunyan has](https://github.com/trentm/node-bunyan#streams) would be required for this use case.

Additionally, if we can change the API to support adding an additional stream after construction, it would also be helpful -- as some streams have to be created asynchronously.

mcollina:
This can be used to implement #145, but done separately so we can keep multi-stream support out of the core `pino` module.

This was partially done in https://github.com/pinojs/pino/pull/255.

cc @yunong

Fixes #249

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -27,6 +27,7 @@
     + [.epochTime](#epochTimeFunction)
     + [.slowTime](#slowTimeFunction)
     + [.nullTime](#nullTimeFunction)
++ [Metadata Support](#metadata)
 
 # module.exports
 
@@ -76,7 +77,8 @@
   * `enabled` (boolean): enables logging. Default: `true`
   * `browser` (Object): browser only, may have `asObject` and `write` keys, see [Pino in the Browser](../readme.md#browser)
 + `stream` (Writable): a writable stream where the logs will be written.
-  Default: `process.stdout`
+  It can also receive some log-line [metadata](#metadata), if the
+  relative protocol is enabled. Default: `process.stdout`
 
 ### Example:
 ```js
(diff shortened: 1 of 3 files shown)


docs/API.md
```diff
@@ -576,3 +578,31 @@ performance penalty.
 
 Returns an empty string. This function is used when the `timestamp` option
 is set to `false`.
+
+<a id="metadata"></a>
+# Metadata
+
+A destination stream can have a property `Symbol.for('needsMetadata')`
+to indicate that for every log line that is written, it wants set to
```
jsumners:
```
... for every log line written, the following properties of the stream should be set:

* `stream.lastLevel`: the last logging level used
...
```

Also, what does "last" refer to here? The current level being used, or the level previous to the current level being used? That is, if we have:

```
log.debug('foo')
log.info('bar')
```

Which of those levels would be assigned to `stream.lastLevel` when `log.info('bar')` is logged? The same question applies to the other `last` properties.

mcollina: It's the current level/msg/logger etc.
when write is called with `log.debug('foo')`, `lastLevel`  will be `20`, `lastMsg` will be `'foo'` and so on.

mcollina: updated @jsumners can you check again?

docs/API.md
```diff
@@ -583,8 +583,8 @@ is set to `false`.
 # Metadata
 
 A destination stream can have a property `Symbol.for('needsMetadata')`
-to indicate that for every log line that is written, it wants set to
-itself:
+to indicate that for every log line written, the following properties of the stream
+should be set:
 
 * the last logging level as `this.lastLevel`
```
jsumners: I still think `this` should be `stream` to make it clear they are properties of the stream and not Pino.

Issue #263, opened 2017-07-03
Clarify out-of-process log processing reason


jsumners:
Issues #111 and #262, amongst other comments, show that we are not clear enough in our documentation about the reasoning behind not parsing logs in-process. We should take the information at the end of #111 and put it prominently in the readme.

I'll submit a PR soonish.

jsumners:
Fixed by #267

PR #267, opened 2017-07-06
Add a detailed overview of transports to initial readme

jsumners:
See issue #263 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ It also includes a shell utility to pretty-print its log files.
 * [API ⇗](docs/API.md)
 * [Extreme mode explained ⇗](docs/extreme.md)
 * [Pino Howtos ⇗](docs/howtos.md)
-* [Transports with Pino ⇗](docs/transports.md)
+* [Transports with Pino](#transports)
 * [Pino in the browser](#browser)
 * [Caveats](#caveats)
 * [Team](#team)
@@ -139,6 +139,40 @@ In many cases, pino is over 6x faster than alternatives.
 For a fair comparison, [LogLevel](http://npm.im/loglevel) was extended
 to include a timestamp and [bole](http://npm.im/bole) had
 `fastTime` mode switched on.
+
+<a name="transports"></a>
+## Transports
+
+A transport in most logging libraries is something that runs in-process to
+perform some operation with the finalized log line. For example, a transport
+might send the log line to a standard syslog server after processing the log
+line and reformatting it. For details on implementing, and some already written,
+transports, see our [Transports⇗](docs/transports.md) document.
+
+> **Pino *does not* natively support in-process transports.**
+
+Pino does not support in-process transports because Node processes are
+single threaded processes (ignoring some technical details). In order for Pino
+to achieve the speed it achieves, Pino purposefully offloads the handling of
+logs, and their ultimate destination, to external processes so that the
+threading capabilities of the OS can be used (or other CPUs).
+
+One consequence of this methodology is that "error" logs do not get written to
+`stderr`, as might be exepected. However, since Pino logs are in a parseable
+format, it is possible to use tools like [pino-tee][pino-tee] or [jq][jq]
+to work with the logs. For example, to view only logs marked as "error" logs:
+
+```
+$ node an-app.js | jq 'select(.level == 50)'
+```
+
+In short, the way Pino generates logs:
+
+1. Reduces the impact of logging on your application to an extremely minimal amount.
+2. Gives greater flexibility in how logs are processed and stored.
+
+[pino-tee]: https://npm.im/pino-tee
+[jq]: https://stedolan.github.io/jq/
 
 <a name="browser"></a>
 ## Pino in the browser
--- a/README.md
+++ b/README.md
@@ -158,9 +158,9 @@ logs, and their ultimate destination, to external processes so that the
 threading capabilities of the OS can be used (or other CPUs).
 
 One consequence of this methodology is that "error" logs do not get written to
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -140,6 +140,45 @@ For a fair comparison, [LogLevel](http://npm.im/loglevel) was extended
...
+
+A transport in most logging libraries is something that runs in-process to
+perform some operation with the finalized log line. For example, a transport
+might send the log line to a standard syslog server after processing the log
+line and reformatting it. For details on implementing, and some already written,
+transports, see our [Transports⇗](docs/transports.md) document.
+
+> **Pino *does not* natively support in-process transports.**
+
+Pino does not support in-process transports because Node processes are
```
davidmarkclements: this isn't true - we do now (multi stream)

README.md
```diff
@@ -140,6 +140,45 @@ For a fair comparison, [LogLevel](http://npm.im/loglevel) was extended
...
+single threaded processes (ignoring some technical details). In order for Pino
+to achieve the speed it achieves, Pino purposefully offloads the handling of
```
davidmarkclements: This  indicates that's how Pino achieves high speed - it isn't. 

It's rather that we recommend out-of-process as the optimum

README.md
```diff
@@ -140,6 +140,45 @@ For a fair comparison, [LogLevel](http://npm.im/loglevel) was extended
...
+logs, and their ultimate destination, to external processes so that the
+threading capabilities of the OS can be used (or other CPUs).
+
+One consequence of this methodology is that "error" logs do not get written to
```
davidmarkclements: this is normative for log4j, we're a log4j logger, so while we should mention this I don't think it should be skewed by other peoples misunderstandings of log4j - can we elaborate that it's because of this *and* log4j that all logs go to the same destination stream (as set by the user, defaulting to stdout)

README.md
```diff
@@ -140,6 +140,45 @@ For a fair comparison, [LogLevel](http://npm.im/loglevel) was extended
...
+`stderr`, as might be exepected. However, since Pino logs are in a parseable
```
davidmarkclements: typo: expected

davidmarkclements: probably would avoid saying it's "as might be expected" 
 
whilst it might be, it's not an expectation based on reasonable assumption, rather its a by product of familiarity bias

README.md
```diff
@@ -140,6 +140,45 @@ For a fair comparison, [LogLevel](http://npm.im/loglevel) was extended
...
+```
+$ node an-app.js | jq 'select(.level == 50)'
+```
+
+In short, the way Pino generates logs:
+
+1. Reduces the impact of logging on your application to an extremely minimal amount.
+2. Gives greater flexibility in how logs are processed and stored.
+
+Given all of the above, Pino clearly promotes out-of-process log processing.
```
davidmarkclements: should we create two headings - one for transports and the other for log processing?

README.md
jsumners: That requires a separate package. Out of the box, `npm install pino`, we don't deal with in-process transports.

README.md
jsumners: Hmm, I meant to imply that it is one of the factors in Pino's speed. How about "...details). Given this restriction, one of the methods Pino employs to achieve its speed is to purposfully offload..."?

README.md
jsumners: I specifically left log4j out of this because what you are saying about log4j isn't actually true. We only follow the API it defines (`.error`, `.debug`, etc). But that's only half(?) of log4j. The other part of log4j is that it handles log management. Look at the log4j configuration for Elastic Search -- https://github.com/elastic/elasticsearch/blob/847378a43b066b33f731e4948589f6040f69d2f6/distribution/src/main/resources/config/log4j2.properties . All of that mess is configuring the log4j library to write the combined stream out to various destinations. So by saying we conform to the log4j standard, we'd be suggesting that we should support writing to multiple streams by default.

README.md
jsumners: I'll strike the clause.

README.md
jsumners: I'm trying to directly respond to questions around "transports" that have been raised multiple times. So I'm trying to give strong reasons for the way things are. Does that make sense?

README.md
mcollina: @jsumners is correct. But we should mention multi-stream.

README.md
mcollina: This is the optimimum for a lot of reasons, including performance. Logs are better sent somewhere else for handling, maybe even docker, or somewhere else. Definitely transports do not fit the single-process module that Node.js has.

README.md
mcollina: I'm very -1 on supporting multiple streams by default, even though that is possible.

I'm -1 on mentioning log4j, however we should include a little compatibility table somewhere, discussing how far we are from the other loggers API.

README.md
mcollina: I'm ok on the way @jsumners wrote it.

README.md
jsumners: @mcollina I think such a table should be a separate PR.

mcollina: 👍

README.md
davidmarkclements: ok 👍

README.md
jsumners: So is my proposed "fix" acceptable?

davidmarkclements: :+1:

jsumners:
@davidmarkclements I believe I have addressed your issues that we have agreed need addressing.

PR #268, opened 2017-07-11
Fast child

mcollina:
From:

```
benchBunyanCreation*10000: 840.878ms
benchBoleCreation*10000: 340.021ms
benchPinoCreation*10000: 357.164ms
benchPinoExtremeCreation*10000: 180.822ms
```

To:

```
benchBunyanCreation*10000: 795.231ms
benchBoleCreation*10000: 334.262ms
benchPinoCreation*10000: 323.676ms
benchPinoExtremeCreation*10000: 161.601ms
```

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -19,21 +19,21 @@ function copy (a, b) {
   return a
 }
 
-function applyOptions (that, opts) {
-  that.serializers = opts.serializers
-  that.chindings = opts.chindings
+function applyOptions (self, opts) {
+  self.serializers = opts.serializers
+  self.chindings = opts.chindings
 
   if (opts.level && opts.levelVal) {
     var levelIsStandard = isStandardLevel(opts.level)
     var valIsStandard = isStandardLevelVal(opts.levelVal)
     if (valIsStandard) throw Error('level value is already used: ' + opts.levelVal)
-    if (levelIsStandard === false && valIsStandard === false) that.addLevel(opts.level, opts.levelVal)
+    if (levelIsStandard === false && valIsStandard === false) self.addLevel(opts.level, opts.levelVal)
   }
-  that._setLevel(opts.level)
+  self._setLevel(opts.level)
   var str = baseLog +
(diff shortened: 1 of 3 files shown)


mcollina:
This improves fastify throughput by 10%+, from 14k to 16k on my machine, as now `child()` can inline in pino-http.

lib/tools.js
```diff
@@ -18,42 +19,21 @@ function copy (a, b) {
...
-  this.name = opts.name
-  this.timestamp = opts.timestamp
-  this.slowtime = opts.slowtime
-  this.chindings = opts.chindings
-  this.cache = opts.cache
-  this.formatOpts = opts.formatOpts
-  this.onTerminated = opts.onTerminated
-  this.messageKey = opts.messageKey
-  this.messageKeyString = opts.messageKeyString
+function applyOptions (that, opts) {
```
davidmarkclements: can we call `that`  `instance`

pino.js
```diff
@@ -122,23 +125,19 @@ Object.defineProperty(
...
-  if (!msg) {
-    if (objError) {
-      msg = obj.message
-    } else {
-      msg = undefined
-    }
-  }
+  var hasObj = obj !== undefined && obj !== null
+  var objError = hasObj && obj instanceof Error
+  msg = !msg && objError ? obj.message : msg || undefined
```
davidmarkclements:
```js
msg = msg ? (objError ? obj.message : msg): undefined
```

pino.js
```diff
@@ -297,7 +288,31 @@ function pino (opts, stream) {
   var instance = Object.create(pinoPrototype)
   instance.stream = istream
   tools.defineLevelsProperty(instance)
-  tools.applyOptions.call(instance, iopts)
+
+  instance.stringify = iopts.stringify
```
jsumners: Moving these back here from `applyOptions` must not have the performance impact that originally necessitated them in `applyOptions`. That's probably because we aren't using `new` any longer :)

mcollina: Yes, I think that's the reason. In this way `applyOptions` inline.

PR #270, opened 2017-07-23
Basic docsify support

### LINKED ISSUE #269: use docsify for the website
https://docsify.js.org seems pretty, and it has an embedded search. Do we like it?

mcollina:
This commit adds [docsify](https://docsify.js.org) support to our docs. Test with

- `npm install docsify-cli -g`
- `docsify serve .`

Try it and let me know what you think.

schermata 2017-07-23 alle 14 51 54

Fixes #269.


index.html
```diff
@@ -0,0 +1,23 @@
...
+  <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1.0, maximum-scale=1.0, minimum-scale=1.0">
+  <link rel="stylesheet" href="//unpkg.com/docsify/lib/themes/vue.css">
+</head>
+<body>
+  <div id="app"></div>
+</body>
+<script>
+  window.$docsify = {
+    name: 'pino',
+    loadSidebar: true,
```
jsumners: The documentation says this can be a string path to the sidebar file. How about `/docsify/sidebar.md` instead?

index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
```
jsumners: Shouldn't this file be in `/docs`?

mcollina: If we place this file in /docs, we would have to duplicate our README. I thought it was preferrable this way.

jsumners:
Also, it looks like you branched from tag 4.7.0 instead of 4.7.1.

index.html
mcollina: I'll see what I can do.

index.html
jsumners: I'm really unclear about where it is supposed to be placed. The documentation is very lacking in this area (and in general, really).

jsumners: Does the `basePath` configuration variable alleviate needing to update the readme?

mcollina: The needs come from howe we configure gh-pages. You can configure to use the things in docs/ or in /. If you use the first, you can't access anything in /.

jsumners: 👍

mcollina:
@jsumners I moved the sidebar into the `docsify`.

Are you folks ok if we publish this? It seems better than nothing :).

PR #280, opened 2017-08-05
change pretty default output localTime

victor0801x:
change pretty.formatTime parse to localTime
(original is convert date.toISOString())

note: this is my first time pr, if do any wrong, I'm very sorry

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -214,4 +214,3 @@ function pretty (opts) {
 }
 
 module.exports = pretty
-
--- a/pretty.js
+++ b/pretty.js
@@ -28,7 +28,6 @@ function toTimezoneOffset (aMinTimeoffset) {
   var tz = aMinTimeoffset || new Date().getTimezoneOffset()
   var tmp = Math.abs(tz)
 
-  // var offset = String(Math.floor(tmp / 60)).padStart(2, '0') + ':' + String(tmp % 60).padStart(2, '0')
   var offset = _lpadzero(String(Math.floor(tmp / 60)), 2) + ':' + _lpadzero(String(tmp % 60), 2)
   return tz > 0 ? '-' + offset : '+' + offset
 }
--- a/pretty.js
+++ b/pretty.js
@@ -122,7 +122,7 @@ function pretty (opts) {
     if (timeTransOnly) {
       value.time = (localTime)
         ? asLocalISODate(value.time, dateFormat)
-        : asISODate(value.time)
+        : asISODate(value.time, dateFormat)
       return JSON.stringify(value) + eol
     }
 
@@ -179,8 +179,12 @@ function pretty (opts) {
     return line
   }
 
-  function asISODate (time) {
-    return new Date(time).toISOString()
+  function asISODate (time, dateFormat) {
+    if (dateFormat) {
+      return asLocalISODate(time, dateFormat, 0)
+    } else {
+      return new Date(time).toISOString()
+    }
   }
 
   function asLocalISODate (aTime, aFormat, aMinuteTZ) {
@@ -188,11 +192,13 @@ function pretty (opts) {
     var format = aFormat || 'YYYY-MM-DDThh:mm:ss.SSSTZ'
     var date = new Date(time)
     // make independent of the system timezone
-    var tzOffset = aMinuteTZ || date.getTimezoneOffset()
+    var tzOffset = (aMinuteTZ === undefined)
+      ? date.getTimezoneOffset()
+      : aMinuteTZ
     date.setUTCMinutes(date.getUTCMinutes() - tzOffset)
     var year = format.indexOf('YYYY') > -1
       ? date.getUTCFullYear()
-      : date.getUTCFullYear().toString().substring(2, 4)
+      : date.getUTCFullYear().toString().slice(2, 4)
     var month = _lpadzero(date.getUTCMonth() + 1, 2)
     var day = _lpadzero(date.getUTCDate(), 2)
     var hour = _lpadzero(date.getUTCHours(), 2)
@@ -220,7 +226,7 @@ function pretty (opts) {
       } else {
         return '[' + ((localTime)
           ? asLocalISODate(value.time, dateFormat)
-          : asISODate(value.time)) + ']' + after
+          : asISODate(value.time, dateFormat)) + ']' + after
       }
     } catch (_) {
       return ''
--- a/pretty.js
+++ b/pretty.js
@@ -73,6 +73,7 @@ function pretty (opts) {
   var timeTransOnly = opts && opts.timeTransOnly
   var formatter = opts && opts.formatter
   var dateFormat = opts && opts.dateFormat
+  var localTime = opts && opts.localTime
   var levelFirst = opts && opts.levelFirst
   var messageKey = opts && opts.messageKey
   var forceColor = opts && opts.forceColor
@@ -119,7 +120,9 @@ function pretty (opts) {
     }
 
     if (timeTransOnly) {
-      value.time = asISODate(value.time, dateFormat)
+      value.time = (localTime)
+        ? asLocalISODate(value.time, dateFormat)
+        : asISODate(value.time)
       return JSON.stringify(value) + eol
     }
 
@@ -176,13 +179,17 @@ function pretty (opts) {
     return line
   }
 
-  function asISODate (aTime, aFmt, aTZO) {
+  function asISODate (time) {
+    return new Date(time).toISOString()
+  }
+
+  function asLocalISODate (aTime, aFormat, aMinuteTZ) {
     var time = aTime
-    var format = aFmt || 'YYYY-MM-DDThh:mm:ss.SSSTZ'
+    var format = aFormat || 'YYYY-MM-DDThh:mm:ss.SSSTZ'
 
     var date = new Date(time)
     // make independent of the system timezone
-    var tzOffset = aTZO || date.getTimezoneOffset()
+    var tzOffset = aMinuteTZ || date.getTimezoneOffset()
     date.setUTCMinutes(date.getUTCMinutes() - tzOffset)
     var year = format.indexOf('YYYY') > -1
       ? date.getUTCFullYear()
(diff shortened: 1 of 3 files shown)


mcollina:
Thanks for your contribution! The build on travis is failing, you might have changed some of the existing functionality. Can you check? you can run the unit tests via npm test. 

Can you also add some to cover the new functionality?

victor0801x:
this should not break anything on tests. 
```
test/addlevel.test.js ............................... 21/21
test/basic.test.js .................................. 98/98
test/browser.test.js .............................. 168/168
test/error.test.js .................................. 10/10
test/escaping.test.js ............................... 39/39
test/events.test.js ................................. 13/13 6s
test/extreme.test.js .................................. 7/7 2s
test/http.test.js ................................... 19/19
test/levels.test.js ................................. 37/37
test/metadata.test.js ............................... 24/24
test/pretty.test.js ................................. 43/43
test/serializers.test.js .............................. 9/9
test/timestamp.test.js .............................. 16/16
total ............................................. 504/504

  504 passing (13s)

  ok

```
only changes pretty.js internal `formatTime()` to call `asLocalDate()`, return local-time, `'YYYY-MM-DDThh:mm:ss.SSSTZ', // '2017-08-05T12:30:45.789+08:00'`, instead of `asISODate()`, return utc iso-format-time.

currently, does not support custom input format time string. I need to figure out how to parse bin.js command input custom string, then pass to pretty.js option

mcollina:
I think adding unit tests to this PR should be a first priority. You can use http://npm.im/minimist to parse input, if we aren't already doing so.

victor0801x:
sorry, I'm not familiar with github. i close first

mcollina:
Why did you close? It's ok ;). I'm happy to review things as you upload them.

jsumners:
@victor0801x sorry about this, but I merged in a couple other PRs that cause this to have conflicts. Will you resolve them so this can be merged?

jsumners:
#317 also provides an improved API for the custom formatter function. You may want to investigate using it to solve this issue.

victor0801x:
sorry for reply so late, recently do not have spare time til March, hope someone can improve or fix this

victor0801x:
@jsumners I fixed the conflict, but ci is failed ?

jsumners:
This test is failing:

https://github.com/pinojs/pino/blob/c77d8ec5ce37f56cbdf1689d4ec9ac585eb6b74b/test/browser.test.js#L217-L230

@davidmarkclements any ideas why?

@victor0801x you have left some dead code and a TODO in place:

https://github.com/victor0801x/pino/blob/c37a2acc5d621bd96f412ac007cffc420b694e4b/pretty.js#L189-L195

Why?

victor0801x:
1. original `blabla |pino -t ` will convert Epoch timestamp into iso timestamp by call asISODate (UTC)
This pr will convert into local-iso-timestamp by calling asLocalDate
Not sure, everybody accept. also, not sure asISODate  function should be remove or not ? 
so I temporarily comment in order to pass the test.

2. I think maybe can convert into user specified timezone by ` |pino -t +8` or ` |pino --tz +8`.
also not sure this is necessary when default is convert into localISODate

jsumners:
My suggestion in https://github.com/pinojs/pino/pull/280#issuecomment-342483584 was to consider providing an alternative formatter function that outputs the date in this manner. That would make it easier to swap between the two types of logging.

victor0801x:
how can I pass the formater function from cli ?

jsumners:
As with your current approach, which is fine I think, you'd have to add some sort of switch to toggle them.

davidmarkclements:
I think it may be due to a patch update to quick-format-unescaped - there was an escaping bug for a particular case - need to look into it more

jsumners:
@pinojs/core what do you think of this PR?

test/pretty.test.js
```diff
@@ -367,6 +367,24 @@ test('works without time', function (t) {
...
+    t.ok(line.match(/(?!^)INFO.*/), 'includes level')
+    t.ok(line.indexOf('' + process.pid) > 0, 'includes pid')
+    t.ok(line.indexOf('' + hostname) > 0, 'includes hostname')
+    t.ok(Date.parse(line.substr(1, 29)) === parseInt(now), 'local date timezone')
+    return line
+  }))
+
+  var instance = pino(prettier)
+
+  instance.info('>' + Date.now() + '<' + ' hello world')
```
mcollina: This Date.now() may not be doing what you think. Can you remove it?

pretty.js
```diff
@@ -23,6 +23,27 @@ var standardKeys = [
   'v'
 ]
 
+function toTimezoneOffset (aMinTimeoffset) {
+  // +/- minute timeoffset
+  var tz = aMinTimeoffset || new Date().getTimezoneOffset()
+  var tmp = Math.abs(tz)
+
+  // var offset = String(Math.floor(tmp / 60)).padStart(2, '0') + ':' + String(tmp % 60).padStart(2, '0')
```
mcollina: can you remove this commented out line?

pretty.js
```diff
@@ -155,17 +177,43 @@ function pretty (opts) {
...
+    var _format = format
+      .replace(/Y{1,4}/g, year)
+      .replace(/MM/g, month)
+      .replace(/DD/g, day)
+      .replace(/hh/g, hour)
+      .replace(/mm/g, minute)
+      .replace(/ss/g, second)
+      .replace(/SSS/g, milli)
+      .replace(/TZ/g, toTimezoneOffset(tzOffset))
+    return _format
```
mcollina: are we sure there is no common utility we can use instead of this block of replace?

test/pretty.test.js
victor0801x:
````js
instance.info('>' + Date.now() + '<' + ' hello world')
var now = line.slice(line.indexOf('>') + 1, line.indexOf('<'))
t.ok(Date.parse(line.substr(1, 29)) === parseInt(now), 'local date timezone')
````
According to MDN, `Date.now()` will return the number of milliseconds elapsed since January 1, 1970 00:00:00 UTC. right?
this is what I try to do
```
[2018-02-10T17:32:02.719+08:00] INFO (8484 on jimmy): >1518255122719< hello world
Date.parse('2018-02-10T17:32:02.719+08:00') === 1518255122719
           ^^^ local datetime with tz ^^^
```

File: pretty.js (hunk 3 above)
Comment (victor0801x) [reply pretty.js]: I think this is minimal implemantation without any deps.
this is a log libary not a datetime libary !? XD
there must have another fast convert implemantation

File: test/pretty.test.js (hunk 1 above)
Comment (mcollina) [reply test/pretty.test.js]: I don't think this is a good way to test this. You are assuming that the time picked from `instance.info()` is the same of the log message. However, this is not guaranteed. 
A better approach might be to mock `Date.now()`.

File: pretty.js (hunk 3 above)
Comment (mcollina) [reply pretty.js]: Can you make the _previous code_ as the default (`new Date(time).toISOString()`)?

Comment (victor0801x) [reply pretty.js]: sure, add new function asLocalISODate and keep asISODate as the previouse code
and add CLI options (`--localTime`) transform to localISOTime with tz(`YYYY-MM-DDThh:mm:ss.SSSTZ`)
but, not sure how to add unit test for `asLocalISODate`, only bin.js for pretty print

File: test/pretty.test.js [hunk 4]
```diff
@@ -354,7 +354,7 @@ test('works without time', function (t) {
   var prettier = pretty()
   prettier.pipe(split(function (line) {
     t.ok(line.match(/.*hello world$/), 'end of line matches')
-    t.ok(line.match(/^INFO.*/), 'includes level')
+    t.ok(line.match(/(?!^)INFO.*/), 'includes level')
```
mcollina: why this change?

victor0801x: my bad, I finally figure out why previouse code doing this
```js
    line = (levelFirst)
        ? asColoredLevel(value) + ' ' + formatTime(value)
        : formatTime(value, ' ') + asColoredLevel(value)
```
I change levelFirst code  into
```js
    line = (levelFirst)
        ? asColoredLevel(value) + ' ' + formatTime(value)
        : formatTime(value) + ' ' + asColoredLevel(value)
```
cause the start of output is `a space + level`, then missing time test will fail
i will make this to previouse code

victor0801x: I remove the previouse test I add, because the default output is `asISODate`, I do not have to add this test again. 
I need some help, how to add test for `asLocalISODate`, it only expose to CLI use. Separate code into js? expose method from pretty.js?

jsumners: Simplest (being a relative term) would be to use `child_process.fork` and parse the `stdout` stream of the child.

victor0801x:
not sure, why node v6.x ci failed !!
I add --localTime transform to local-iso-date & --dateFormat for parse custom format test
@mcollina @jsumners any advises

jsumners:
No idea. I restarted the job and it worked.

test/pretty.test.js
```diff
@@ -380,3 +381,66 @@ test('does not throw error when enabled with stream specified', function (t) {
...
+
+  child.on('close', function () {
+    line = line.slice(0, -1 * (os.EOL).length)
+    var formatDate = line.slice(line.indexOf('[') + 1, line.indexOf(']'))
+    var toISODate = formatDate.replace(/\//g, '-').replace(/,/g, ':').replace(/_/g, '.')
+    t.ok(line.match(/.*hello world$/), 'end of line matches')
+    t.ok(line.match(/(?!^)INFO.*/), 'includes level')
+    t.ok(line.indexOf('' + process.pid) > 0, 'includes pid')
+    t.ok(line.indexOf('' + hostname) > 0, 'includes hostname')
+    t.ok(Date.parse(toISODate) === parseInt(now), 'custDateFormat <-> UTC match')
```
mcollina: This needs to be checked using an interval. These numbers are not assured to be identical on a CI system, as they are dependent on what else is running.

victor0801x: how can i check custom dateformat or local-time without passing epo

victor0801x: how can I check transformed dateformat or local-time is correct? 
this is my thought, I pass time from msg (log.time = log.msg), then i can check the datestring

mcollina: How about 

```
t.ok(Date.parse(toISODate) > parseInt(now) - 2000)
t.ok(Date.parse(toISODate) < parseInt(now) + 2000)
```

victor0801x: sure, I misunderstand your words. 
by the way, is there any suggestions/tricks for comparing floating-point ?

PR #282, opened 2017-08-08
Make _baseLog configurable

Rauno56:
I don't personally see why a general logger should force `pid` and `hostname` to be there in the first place. In some cases(containers, faas) those might be so short-lived or random that it would convey next to no information. In other cases some would just like to structure `pid` and `hostname` differently - under another property for example.

This PR doesn't change the default behaviour but at least makes it possible to overwrite `_baseLog` on runtime removing `name`, `pid` and `hostname` from log events.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -1,5 +1,6 @@
 'use strict'
 
+var os = require('os')
 var EventEmitter = require('events').EventEmitter
 var stringifySafe = require('fast-safe-stringify')
 var fs = require('fs')
@@ -34,6 +35,10 @@ var defaultOptions = {
   },
   enabled: true,
   messageKey: 'msg'
+}
+var defaultBaseLog = {
+  pid: process.pid,
+  hostname: os.hostname()
 }
 
 var pinoPrototype = Object.create(EventEmitter.prototype, {
@@ -336,6 +341,15 @@ function pino (opts, stream) {
       var fd = (istream.fd) ? istream.fd : istream._handle.fd
       fs.writeSync(fd, buf)
     }
+  }
+
+  var baseLog = (typeof iopts.baseLog === 'object') ? iopts.baseLog : defaultBaseLog
+  instance = instance.child(baseLog)
+
+  if (iopts.name !== undefined) {
+    instance = instance.child({
+      name: iopts.name
+    })
   }
 
   return instance
--- a/pino.js
+++ b/pino.js
@@ -133,7 +133,7 @@ function asJson (obj, msg, num) {
   var hasObj = obj !== undefined && obj !== null
   var objError = hasObj && obj instanceof Error
   msg = !msg && objError ? obj.message : msg || undefined
-  var data = this._baseLog + this._lscache[num] + this.time()
+  var data = '{' + this._lscache[num] + this.time()
   if (msg !== undefined) {
     // JSON.stringify is safe here
     data += this.messageKeyString + JSON.stringify('' + msg)
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -1,12 +1,7 @@
 'use strict'
 
-var flatstr = require('flatstr')
 var format = require('quick-format-unescaped')
 var util = require('util')
-var pid = process.pid
-var os = require('os')
-var hostname = os.hostname()
-var baseLog = flatstr('{"pid":' + pid + ',"hostname":"' + hostname + '",')
 var levels = require('./levels')
 var serializers = require('./serializers')
 var isStandardLevelVal = levels.isStandardLevelVal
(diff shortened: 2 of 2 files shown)


mcollina:
@Rauno56 how about we allow to avoid setting them in the first place? I prefer **not** to give outside control of an internal optimization that we might want to change.

Rauno56:
Setting pid and hostname you mean? I'm all for it, but that would be semver
major change.

mcollina:
not really, if the default stays as it is.

davidmarkclements:
Yah just allow hostname and pid to be set via options object - if they're set to false then don't include them - then assemble baselog based on the options

davidmarkclements:
based on #283 maybe we should take it further and allow other keys to be defined as well

Rauno56:
To reiterate: What _baseLog gives us is basically the same thing as child loggers - no special reason to include them out of the box.

I'd make so that "_baseLog"(now the name does really not make sense) will we just "{" during the initialization and for backwards compatibility we'd add an property to options object - something like "baseLog" which would be over-writable, would take an object and basically act just like creating an child logger right after initialization. That would somewhat simplify internal API.

Eventually I'd even go as far as to remove mentioned default of adding pid and hostname during initialization altogether with some major version update, but that, I guess, is another topic.

Rauno56:
I updated this PR according to my proposal.

pino.js
```diff
@@ -338,6 +343,15 @@ function pino (opts, stream) {
     }
   }
 
+  const baseLog = typeof iopts.baseLog === 'object' ? iopts.baseLog : defaultBaseLog
```
jsumners: The usage of `const` is inconsistent with the rest of the codebase. Please change this to `var`. Also, I'd like to see some grouping here to make the ternary easier to read.

Rauno56:
Sure thing. This is a proposal, I'll not implement tests if even the concept does not land well with the maintainers.

pino.js
Rauno56: Good catch, thanks. Perhaps this rule should be integrated to linter? Or start using const and let instead elsewhere to change the base.
Could you please elaborate more on the mentioned grouping you'd like to see? Can't seem to find a clear example of that to follow.

jsumners: `(typeof iopts.baseLog === 'object') ? ...`

davidmarkclements: let's not call it baseLog

how about just calling it either `baseKeys` or `base`

just to define it further

* defaults to having pid and hostname
* may only be an object
* these can be removed if set to `false` e.g. `{base: { pid: false }}`
* can also have other preset values which will be added into `_baseLog` e.g. `{base: { container: 'my-container-name' }}`
* child loggers inherit the base (I think this happens anyway)

davidmarkclements: * also need to think about processing keys which are objects or arrays: e.g. `{base: { trace: ['a', 'b', 'c']}`

mcollina:
I am 👍  with the change as long as:

- benchmarks are not affected
- there are no broken tests for the current functionality
- there are tests for the new functionality

Rauno56:
Awesome, thanks. I'll continue with that then.

pino.js
```diff
@@ -338,6 +343,15 @@ function pino (opts, stream) {
...
   }
 
+  var baseLog = (typeof iopts.baseLog === 'object') ? iopts.baseLog : defaultBaseLog
+  instance = instance.child(baseLog)
+
+  if (iopts.name !== undefined) {
+    instance = instance.child({
+      name: iopts.name
+    })
+  }
```
mcollina: You can use `Object.assign` to only call `child` once

pino.js
```diff
@@ -338,6 +343,15 @@ function pino (opts, stream) {
     }
   }
 
+  var baseLog = (typeof iopts.baseLog === 'object') ? iopts.baseLog : defaultBaseLog
+  instance = instance.child(baseLog)
+
+  if (iopts.name !== undefined) {
+    instance = instance.child({
```
davidmarkclements: rather than creating two children, we can just merge this into `baseLog` - this will avoid duplicate name keys

pino.js
```diff
@@ -35,6 +36,10 @@ var defaultOptions = {
   enabled: true,
   messageKey: 'msg'
 }
+var defaultBaseLog = {
```
davidmarkclements: at this point the term "baseLog" has changed conceptually, for historical reasons and to save confusion we should name it differently, even internally 

can we please call this `defaultBase`

pino.js
```diff
@@ -338,6 +343,15 @@ function pino (opts, stream) {
     }
   }
 
+  var baseLog = (typeof iopts.baseLog === 'object') ? iopts.baseLog : defaultBaseLog
```
davidmarkclements: `baseLog` -> `base` 
`iopts.baseLog` -> `iopts.base`
`defaultBaseLog` -> `defaultBase`

pino.js
davidmarkclements: LOL I said that about the same time as you but it's batched in review - to reiterate - good reason for doing this is to avoid duplicate keys in JSON output

robmcguinness:
I'm interested in this feature and don't mind contributing if the PR has stalled. Please let me know if I can help.

jsumners:
@robmcguinness go for it.

robmcguinness:
Submitted #324. Apologies if there was a better way to reuse this PR.

Rauno56:
Nope... My thoughts were moving in the same direction. Good job!

PR #284, opened 2017-08-10
Really force color if forceColor: true

Rauno56:
chalk does not output any color if level=0 even if enabled=true.
This checks for that case and really-really forces at least some color.

Monkey-patched pipe method also lacked return breaking possible expectations for the behaviour of core pipe method.


Rauno56:
Alternative to forcing level would be to run tests with environment variable `FORCE_COLOR=1` which causes exactly the same effect.

mcollina:
Thanks!!!!

Issue #281, opened 2017-08-07
Not inspecting surplus arguments


asafh:
While the API doc for the logging methods describes
> ... If more args follows msg, these will be used to format msg using util.format.  

The actual behaviour is different when sending more arguments than interpolation placeholders. While console.log, util.format and the npm module 'debug' all behave similarly (concatenating with space these extras after running them through util.inspect), pino simply truncates these extra arguments - tools.js:94:
`        n.length = l + countInterp(a, '%d') + countInterp(a, '%s') + 1
`  
Doesn't necessarily feel like a speed optimization since it just blocks this use case (to avoid these objects being inspected, don't send them in the first place - same result).  
If this makes sense to you too, would the correct logic to apply in genLog be: if no %j and countInterp(%j,%d,%s) <= n.length we can call quick escape, otherwise pass all of n into util.format.

What do you think?

mcollina:
Can you make an example with how pino behaves now, and how you would like it to behave?

asafh:
For this code:
```
const util = require("util");
const pino = require('pino')({});
const additionalObject = new Error("ErrName");
const value = "fromInteropt";
pino.info("pino %s", value, additionalObject);
console.log("console interopt=%s", value, additionalObject);

const debug = require("debug")("debug-tag");
debug("debug interopt=%s", value, additionalObject);
```
You can see both console and debug print the error message and stack, while pino only prints an empty object (though earlier I thought even that doesn't happen, I should investigate further).
```
$ DEBUG=* node test.js
{"pid":14342,"hostname":"asafh-mac.local","level":30,"time":1502110326544,"msg":"pino fromInteropt {}","v":1}
console interopt=fromInteropt Error: ErrName
    at Object.<anonymous> (/Users/asafh/test/test.js:4:26)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
    at tryModuleLoad (module.js:466:12)
    at Function.Module._load (module.js:458:3)
    at Function.Module.runMain (module.js:605:10)
    at startup (bootstrap_node.js:158:16)
    at bootstrap_node.js:575:3
  debug-tag debug interopt=fromInteropt Error: ErrName
    at Object.<anonymous> (/Users/asafh/test/test.js:4:26)
    at Module._compile (module.js:569:30)
    at Object.Module._extensions..js (module.js:580:10)
    at Module.load (module.js:503:32)
[...]

PR #285, opened 2017-08-17
Issue #281, not truncating surplus format arguments when using util.format

asafh:
This changes has surplus message format values (surplus - more than the number of placeholder tokens in the message string, specifically those we count - %d, %j, %s) would still be present in the message result, regardless whether util.format or quick-format-unescaped is used (which is dependent on whether or not a %j placeholder is present).  
After this change the behaviour will match the documentation in the case we have a %j placeholder, and won't change otherwise - though it's fairly similar already (the difference being, quick-format-unescaped uses JSON.stringify for this surpluses, which is considerable faster but doesn't give the same result as util.format which the documentation suggests is used).  
Performance wise, the behaviour is the same as long as you don't pass surpluses, with the benefit of not having to count the placeholders.

PS. Notice I also removed the wrap of the util.format call with the template literal, as it did nothing - it had no additional value other than the return value, so it acted as a cast to string, but util.format always returns a string.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -91,8 +91,7 @@ function genLog (z) {
     if (p > 1) {
       l = typeof a === 'string' ? countInterp(a, '%j') : 0
       if (l) {
-        n.length = l + countInterp(a, '%d') + countInterp(a, '%s') + 1
-        o = `${util.format.apply(null, n)}`
+        o = util.format.apply(null, n)
       } else {
         o = format(n, this.formatOpts)
       }


mcollina:
Can you add a unit test for this?

lib/tools.js
```diff
@@ -91,8 +91,7 @@ function genLog (z) {
     if (p > 1) {
       l = typeof a === 'string' ? countInterp(a, '%j') : 0
       if (l) {
-        n.length = l + countInterp(a, '%d') + countInterp(a, '%s') + 1
-        o = `${util.format.apply(null, n)}`
+        o = util.format.apply(null, n)
```
davidmarkclements: the ES6 string is deliberate - (pretty much every oddity in this code is deliberate) - it triggers an optimization (can't remember what, but this is (or was) a technique is being used in core)

to determine whether we can remove the template string optimization we'll need benchmarks for 4, 6, and 8 *just* for the removal of the back ticks - can we put the back ticks back in for this PR and if you're up for it we can look at the backtick removal  in a separate a PR with it's own benchmarks

lib/tools.js
```diff
@@ -91,8 +91,7 @@ function genLog (z) {
     if (p > 1) {
       l = typeof a === 'string' ? countInterp(a, '%j') : 0
       if (l) {
-        n.length = l + countInterp(a, '%d') + countInterp(a, '%s') + 1
```
davidmarkclements: the original reason for truncating arguments to interpolation char count was performance, 
at the time there was a *massive* performance hit from having an argument length greater that interpolation char count in util.format 

this was known behaviour at the time of writing the code, but somewhere along the line the documentation was written without knowledge of this behaviour 

it may not be true any more, we need benchmarks from Node 4, 6, and 8

davidmarkclements:
hey @asafh thanks for the PR :bowtie: 

you're right, we need to resolve this in some way, whether that means correcting the documentation or finding a way to make the API more intuitive without significant performance degradation

it may be that the original performance problem with passing extra args has gone since we first wrote this, could you please run and add benchmark results for node 4, 6 and 8 (also put the backticks back, that's also an optimization)

jsumners:
Node 4 is in maintenance mode. My opinion is it can be ignored.

davidmarkclements:
@jsumners from one perspective, completely agree

from another - the more enterprise users we have, the longer the tail is for supporting older Node

we're fortunate that @watson picked up the baton for 0.10 and 0.12 - but that right there is a signal in itself regarding ideal vs real adoption rates

lib/tools.js
asafh: Sure, I'll just remove it from this PR and figure whether or to remove it separately.

lib/tools.js
asafh: Wouldn't that classify as usage issue? Truncating these arguments is basically just blocking that use case (even inconsistently - depending on placeholder presence, it might cause a silent loss of information).  
I'd argue that regardless of whether it still is costly (which is likely), this would be an explicit usage choice. Just like passing huge JSON objects to be serialized, passing objects without placeholders to be inspected is the user's choice when writing the code. Would you agree?

davidmarkclements: I don't think I agree that it's a usage *choice* - not in the sense of a *conscious choice* where effects are known. That's the difference between this and passing the large objects. With the large objects, you know you're paying that price. With this, things can slow drastically and it's completely unclear why. 

It's been a while, my understanding is that the additional args are mixed into the object (in the case of objects) or concatted to the end at some other point

If that's not the case, I don't think losing info is acceptable either ... but nor is throwing 

But let's see what the benchmarks tell us for 4, 6, 8 – since the simplest solution (yours) may be perfectly fine. If it isn't we need to figure another way around it

jsumners:
@asafh are you still working on this? Or should I close it?

davidmarkclements:
closing for now

PR #291, opened 2017-09-13
epoch time is in seconds not ms

doron2402:
Epoch, also known as Unix timestamps, is the number of seconds (not milliseconds!) that have elapsed since January 1, 1970 at 00:00:00 GMT (1970-01-01 00:00:00 GMT). 
taken from https://www.freeformatter.com/epoch-timestamp-to-date-converter.html

### DIFF
--- a/lib/time.js
+++ b/lib/time.js
@@ -5,7 +5,7 @@ function nullTime () {
 }
 
 function epochTime () {
-  return ',"time":' + Date.now()
+  return ',"time":' + Math.round(Date.now()/1000.0)
 }
 
 function slowTime () {


jsumners:
Technically, the epoch is the literal time, not the integer representation of it. So while the function name can be seen as a misnomer for general parlance, it is still correct. We use the milliseconds representation because that is what `new Date()` expects, and is the normal method of working with the epoch in JavaScript.

Maybe you'd like to alter this PR to add a new `unixTime` function?

doron2402:
Sounds good

doron2402:
fix in a seperate [pr](https://github.com/pinojs/pino/pull/292)

PR #292, opened 2017-09-13
unixTime

doron2402:
Epoch, also known as Unix timestamps, is the number of seconds (not milliseconds!) that have elapsed since January 1, 1970 at 00:00:00 GMT (1970-01-01 00:00:00 GMT).
taken from https://www.freeformatter.com/epoch-timestamp-to-date-converter.html

See [PR](https://github.com/pinojs/pino/pull/291)

@jsumners

### DIFF
--- a/lib/time.js
+++ b/lib/time.js
@@ -8,6 +8,10 @@ function epochTime () {
   return ',"time":' + Date.now()
 }
 
+function unixTime () {
+  return ',"time":' + Math.round(Date.now() / 1000.0)
+}
+
 function slowTime () {
   return ',"time":"' + (new Date()).toISOString() + '"'
 }
@@ -15,5 +19,6 @@ function slowTime () {
 module.exports = {
   nullTime: nullTime,
   epochTime: epochTime,
+  unixTime: unixTime,
   slowTime: slowTime
 }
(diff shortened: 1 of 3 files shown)


doron2402:
√ update

lib/time.js
```diff
@@ -8,12 +8,17 @@ function epochTime () {
   return ',"time":' + Date.now()
 }
 
+function unixTime () {
+  return ',"time":' + Math.round(Date.now() / 1000.0)
```
davidmarkclements: is there a specific reason to divide by `1000.0` and not `1000`?

jsumners: Is this a blocker for you @davidmarkclements ?

doron2402: habit from strictly typed languages, don't have a strong feeling about this one. your guys call.

doron2402: I think it's a good practice to divide float with another float I get that it's not an issue in JS land.

jsumners: Unless @davidmarkclements chimes back in before I get to a computer, with time to do it, I'll merge.

doron2402: Thanks, @jsumners again your call don't have a strong feeling about this one.

PR #307, opened 2017-10-12
add pino-redis

buianhthang:
Add pino-redis to the transports section.


mcollina:
I am a bit lost about the usecase of Redis as a transport, considering it is using SETEX as a command, meaning it is rewriting the same key over and over again.

buianhthang:
@mcollina with `ttl` option, we can temporary store in Redis in a short time for next steps (data from the latest will be used) and will go away when finished.

mcollina:
Thanks!!

PR #310, opened 2017-10-18
browser serializers, closes #219

### LINKED ISSUE #219: serializers support in the browser
cc @davidmarkclements @temsa

davidmarkclements:
* see readme for full description - default mode differs from server pino in that it applies serializers to all objects prior to passing to console.log, `asObject` mode essentially ends up the same as server pino
* needed for relaying log messages (want to be able to serialize on the client before sending to server)
* 100% cov

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,7 +193,7 @@ By default, in the browser,
 ### Browser Options
 
 Pino can be passed a `browser` object in the options object,
-which can have a `write` property or an `asObject` property.
+which can have the following properties: 
 
 #### `asObject` (Boolean)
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -241,6 +241,67 @@ var pino = require('pino')({browser: {write: {
...
+})
+// following will apply myCustomSerializer to the custom property,
+// but will not apply anotherSerizlier to another key
+pino.info({custom: 'a', another: 'b'})  
+```
+
+When `serialize` is `true` the standard error serializer is also enabled (see https://github.com/pinojs/pino/blob/master/docs/API.md#stdSerializers).
+This is a global serializer which will apply to any `Error` objects passed to the logger methods.
+
+If `serialize` is an array the standard error serializer is also automatically enabled, it can
```
jsumners: Why? The above documentation says that the array value selectively enables the specified serializers. So why would it also enable the error serializer when that wasn't specified?

browser.js
```diff
@@ -5,12 +5,29 @@ var format = require('quick-format-unescaped')
...
+  var serialize = Array.isArray(opts.browser.serialize)
+    ? opts.browser.serialize.filter(function (k) {
+      return k !== '!stdSerializers.err'
+    })
+    : opts.browser.serialize === true ? Object.keys(serializers) : false
+  var stdErrSerialize = opts.browser.serialize
+
+  if (
+    Array.isArray(opts.browser.serialize) &&
+    opts.browser.serialize.indexOf('!stdSerializers.err') > -1
```
jsumners: Does `[].includes()` not exist in browsers?

README.md
davidmarkclements: Because the error serializer is global (based on whether it's an instance of `Error`, rather than based on an object key) 

Initially I made is so if you supply an array you could do

`{ serialize: ['stdSerializers.err'] }` 

But then figured it should actually be opt out, because if you enable some serializers, you probably want to have Error serialization as well (because you're opting into using serialization)

I know it's a tricky one, for me this seemed the best set of tradeoffs, you can explicitly disable Error serializing if you want:

`{ serialize: ['!stdSerializers.err', 'someOtherSerializer'] }` 

I think this makes more sense because it provides the most likely desired behaviour with least config

browser.js
davidmarkclements: because it's browser-side code I'm going for lowest common denominator, includes is ES7 so it's definitely not going to be in IE11 down which I'll wager are still in use - however we could do a polyfill since `includes` is actually more performant (it exits once an element is discovered). I'm not sure if the extra code is worth it at this stage, would rather upgrade in the future when we know there will be no breakage.

davidmarkclements: to elaborate further: the priority optimization with browser pino is payload optimization rather than execution speed, this is why I've stayed away from polyfills. 

You could argue that `Object.assign` is out baseline though, if all browsers that support `Object.assign` support `Array.prototype.includes` then it's safe to use and for legacy browsers it's on the user to polyfill if the need arises 

Still, from an opportunity cost perspective, for now I'm not sure this is worth the research to decide on the change (having said that, happy upon insistence to look into)

jsumners: I was seriously asking. I do so little front end stuff that I have no idea what is available there.

davidmarkclements: on cursory look there's no IE support for includes, but there's also no support for Object.assign so...

PR #312, opened 2017-10-20
browser transmit API

davidmarkclements:
This adds a new API to pino browser, which facilitates log messages to be sent across the wire 

the user supplies a `send` function via a `transmit` object on the `browser` options object, 
the `send` function is passed `level` and `logEvent` params. The `logEvent` represents the log message, including child bindings and their hierarchy, each argument passed to the log method (after serializers have been applied, which is mandatory even if not enabled), and the level label and value. 

The `send` function is totally decoupled from the `write` function, and can use a different level which can be set on the `transmit` object. 

See the readme update for more details

The main point of this API is to open up a new type of client side plugin, e.g. 

```js
var pino = require('pino')
var logger = pino({
  browser: {
    transmit: require('pino-splunk-transmitter')
  }
})
```

Next steps after this is to build `pino-transmit` (sort of a browser-server analogy to `pino-socket`) . `pino-transmit` would be specifically for encoding and relaying messages across the wire using the [beacon API](https://developer.mozilla.org/en-US/docs/Web/API/Beacon_API) on the client side (falling back to XHR), and then receiving, decoding those messages on the server.

Hypothetical modules like `pino-splunk-transmitter` could then be built on top of `pino-transmit`, but `pino-transmit` would also usable as a standalone plugin with the `browser.transmit` api.

Coverage is 100%

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,73 @@ first object (as in server pino).
 
 For more info on serializers see https://github.com/pinojs/pino/blob/master/docs/API.md#parameters.
 
+#### `transmit` (Object)
+
+An object with `send` and `level` properties.
+
+The `transmit.level` property specifies the minimum level (inclusive) of when the `send` function
+should be called, if not supplied the `send` function be called based on the main logging `level`
+(set via `options.level`, defaulting to `info`). 
+
+The `transmit` object must have a `send` function which will be called after 
+writing the log message. The `send` function is passed the level of the log 
+message and a `logEvent` object. 
+
+The `logEvent` object is a data structure representing a log message, it represents
+the arguments passed to a logger statement, the level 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -302,6 +302,73 @@ first object (as in server pino).
...
+The `level` holds the label (for instance `info`), and the corresponding numerical value 
+(for instance `30`). This could be important in cases where client side level values and 
+labels differ from server side.
+
+The point of the `send` function is to remotely record log messages: 
+
+```js
+var pino = require('pino')({
+  browser: {
+    relay: {
```
mcollina: is it `transmit` or `relay`?

jsumners:
👍

Nice addition.

davidmarkclements:
yeah it's been in the pipeline for a while

PR #315, opened 2017-10-24
browser-transmit-timestamp

davidmarkclements:
pretty big omission from the new transmit API, 
logEvents should have a timestamp so when it reaches the server we know
the time of the logEvent (the logEvent doesn't contain the final object, so no `time` prop)

Code has been refactored so we only wrap
the logger function once, and then perform operations (serialize, asObject mode, transmit) from there instead of having heirarchy of function wrappers. 

This also means we can take a timestamp right when the log message is called, and then use that in asObject for `time` prop and as the `ts` prop in the `logEvent` - so we have a consistent timestamp

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,11 +322,15 @@ The `logEvent` format is structured like so:
 
 ```js
 { 
+  ts = Number,
   messages = Array, 
   bindings = Array, 
   level: { label = String, value = Number}
 }
 ```
+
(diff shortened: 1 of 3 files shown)


davidmarkclements:
would this be a patch or minor? 

technically I think it's a minor, because it (backwards compat) extends the API 

but it could be argued that this should have been there in the first place, in which case, it's a bug, so patch

jsumners:
I vote patch.

PR #320, opened 2017-10-29
Added isLevelEnabled function

Epitrochoid:
Adds a function to the Pino object that returns true if its argument is the current level or higher, false otherwise. The use case is to be able to put costly log computations behind a an `if` statement and only calculate them if a certain log level is enabled. e.g.

```javascript
if (logger.isLevelEnabled('debug')) {
  logger.debug(fetchLargeDocument())
}
```

Currently to do this I have to wrap the logger and include a function similar to the one this PR adds.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -250,6 +250,15 @@ Object.defineProperty(pinoPrototype, 'addLevel', {
   value: addLevel
 })
 
+function isLevelEnabled (logLevel) {
+  var logLevelVal = this.levels.values[logLevel]
+  return logLevelVal && (logLevelVal >= this._levelVal)
+}
+Object.defineProperty(pinoPrototype, 'isLevelEnabled', {
+  enumerable: true,
+  value: isLevelEnabled
+})
+
 function pino (opts, stream) {
   var iopts = opts
   var istream = stream
--- a/test/islevelenabled.test.js
+++ b/test/islevelenabled.test.js
@@ -1,0 +1,53 @@
+'use strict'
+
+var test = require('tap').test
+var pino = require('../')
+
+test('can check if current level enabled', function (t) {
+  t.plan(1)
+
+  var log = pino({level: 'debug'})
+  t.is(true, log.isLevelEnabled('debug'))
+})
+
+test('can check if level enabled after level set', function (t) {
+  t.plan(2)
+
+  var log = pino()
+  t.is(false, log.isLevelEnabled('debug'))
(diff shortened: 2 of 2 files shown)


jsumners:
What's wrong with `if (logger.level > 30) logger.debug('something')`?

pino.js
```diff
@@ -250,6 +250,19 @@ Object.defineProperty(pinoPrototype, 'addLevel', {
   value: addLevel
 })
 
+function isLevelEnabled (logLevel) {
+  var logLevelVal = this.levels.values[logLevel]
+  if (logLevelVal && logLevelVal >= this._levelVal) {
```
jsumners: How about `return logLevelVal && (logLevelVal >= this._levelVal)` instead of the `if` statement?

test/islevelenabled.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+test('can check if custom level is enabled', function (t) {
+  t.plan(3)
+
+  var log = pino({level: 'debug'})
+  log.addLevel('foo', 35)
+  t.is(true, log.isLevelEnabled('foo'))
+  t.is(true, log.isLevelEnabled('error'))
+  t.is(false, log.isLevelEnabled('trace'))
+})
```
jsumners: For completeness sake, maybe add a test:

```js
var log = pino()
t.is(false, log.isLevelEnabled('debug'))
log.level = 'debug'
t.is(true, log.isLevelEnabled('debug'))
```

Epitrochoid:
@jsumners in my case we use this pattern pretty often and we would end up with a lot of magic numbers throughout the code base. It gets even hairier because we have a shared library adding a bunch of company wide standard log levels as custom log levels to pino. Someone else using the library would have to pull the repo and look at the `pino.addLevel()` calls to figure out what integer they need to reference for a log name they already know. This seems like standard functionality in other logging libraries with Bunyan returning true or false on an empty call (e.g. `bunyan.debug() == true`) and raptor-logging providing `logging.isDebugEnabled()` for each of its standard levels.

pino.js
Epitrochoid: Good suggestion, fixed.

test/islevelenabled.test.js
Epitrochoid: Added as 'can check if level enabled after level set', hopefully that describes what you meant the test to do.

PR #323, opened 2017-10-31
IE 10 invalid calling object binding fix

sara-gudeman:


browser.js
```diff
@@ -146,13 +146,14 @@ function wrap (opts, logger, val, level) {
     return function LOG () {
       var ts = Date.now()
       var args = new Array(arguments.length)
+      var proto = (Object.getPrototypeOf && Object.getPrototypeOf(this) === _console) ? _console : this
```
sara-gudeman: As per your suggestion @davidmarkclements. Let me know if this is what you had in mind :)

PR #324, opened 2017-11-01
Optional baseLog

robmcguinness:
This is essentially #282 with requested changes. Please review and provide feedback

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -345,13 +345,14 @@ function pino (opts, stream) {
   }
 
   var base = (typeof iopts.base === 'object') ? iopts.base : defaultBase
-  instance = instance.child(base)
 
   if (iopts.name !== undefined) {
-    instance = instance.child({
+    base = Object.assign({}, base, {
       name: iopts.name
     })
   }
+
+  instance = instance.child(base)
 
   return instance
 }
--- a/pino.js
+++ b/pino.js
@@ -36,7 +36,8 @@ var defaultOptions = {
   enabled: true,
   messageKey: 'msg'
 }
-var defaultBaseLog = {
+
+var defaultBase = {
   pid: process.pid,
   hostname: os.hostname()
 }
@@ -343,8 +344,8 @@ function pino (opts, stream) {
     }
   }
 
-  var baseLog = (typeof iopts.baseLog === 'object') ? iopts.baseLog : defaultBaseLog
-  instance = instance.child(baseLog)
+  var base = (typeof iopts.base === 'object') ? iopts.base : defaultBase
+  instance = instance.child(base)
 
   if (iopts.name !== undefined) {
     instance = instance.child({
--- a/pino.js
+++ b/pino.js
@@ -1,5 +1,6 @@
 'use strict'
 
+var os = require('os')
 var EventEmitter = require('events').EventEmitter
 var stringifySafe = require('fast-safe-stringify')
 var fs = require('fs')
@@ -34,6 +35,10 @@ var defaultOptions = {
   },
   enabled: true,
   messageKey: 'msg'
+}
+var defaultBaseLog = {
+  pid: process.pid,
+  hostname: os.hostname()
 }
 
 var pinoPrototype = Object.create(EventEmitter.prototype, {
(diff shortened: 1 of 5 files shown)


pino.js
```diff
@@ -338,6 +344,18 @@ function pino (opts, stream) {
...
 
+  var base = (typeof iopts.base === 'object') ? iopts.base : defaultBase
+
+  if (iopts.name !== undefined) {
+    base = Object.assign({}, base, {
+      name: iopts.name
+    })
+  }
+
+  if (base != null) {
```
jsumners: Is there a reason for the `!=` as apposed to `!==` or `(!base)`?

pino.js
```diff
@@ -36,6 +37,11 @@ var defaultOptions = {
   messageKey: 'msg'
 }
 
+var defaultBase = {
```
jsumners: Why isn't this part of the `defaultOptions` object?

pino.js
```diff
@@ -338,6 +344,18 @@ function pino (opts, stream) {
...
+  var base = (typeof iopts.base === 'object') ? iopts.base : defaultBase
+
+  if (iopts.name !== undefined) {
+    base = Object.assign({}, base, {
+      name: iopts.name
+    })
+  }
+
+  if (base != null) {
+    instance = instance.child(base)
```
jsumners: The default instance should not be a child.

cc: @davidmarkclements @mcollina

pino.js
robmcguinness: no reason my mistake

pino.js
robmcguinness: I simply renamed `defaultBase` according to the comments in #282 (from a previously submitted PR). I don't see a reason why the `base` object can't be a part of `defaultOptions`. Would you rather have something like:

pino.js
robmcguinness: I was following the comments from https://github.com/pinojs/pino/pull/282#pullrequestreview-57188442. Unfortunately I don't know enough about Pino to understand the implications of using a child logger for the default instance. When I get a moment I'll dig deeper into Pino.

pino.js
jsumners: Yes, please.

pino.js
jsumners: It seems odd to me. But that's why I pinged the other two, so they can provide enlightenment.

mcollina: why not? I think it's the best way to do it. Considering that pino children do not keep a reference to their parent, I see no obvious problem with this.

pino.js
```diff
@@ -128,7 +133,7 @@ function asJson (obj, msg, num) {
   var hasObj = obj !== undefined && obj !== null
   var objError = hasObj && obj instanceof Error
   msg = !msg && objError ? obj.message : msg || undefined
-  var data = this._baseLog + this._lscache[num] + this.time()
+  var data = '{' + this._lscache[num] + this.time()
```
mcollina: can you embed `{` into `this._lscache`?

pino.js
mcollina: The only side effect of this is that it makes the base-log properties overridable by a log line. I don't think it's a bad thing: https://github.com/pinojs/pino/blob/master/pino.js#L136-L138.

jsumners: Okay.

pino.js
robmcguinness: @mcollina not sure if this is what you wanted:

mcollina: ??

robmcguinness: You asked: 

> can you embed { into this._lscache?

So I provided some examples above and what I thought that meant. Apologies if I misinterpreted the ask.

robmcguinness: added one more commit. let me know if that is what you were requesting.

mcollina:
Folks, weight in if this is semver-major or not.

jsumners:
It isn't. The default operation of the API hasn't changed from a user perspective. It would merely be possible to do more things.

robmcguinness:
conflicts should be resolved now

mcollina:
@jsumners @davidmarkclements may I get a couple of LGTMs?

jsumners:
Sorry. Thought I already had.

Issue #326, opened 2017-11-02
Use CRLF on Windows


tniessen:
pino seems to always emit `\n`, but it should at least be configurable to emit `\r\n` on Windows, or even make it a default.

This seems to be caused by https://github.com/pinojs/pino/blob/ce10ba1075791d57977bc01cfc5db6aac0b25bf4/pino.js#L277
and by multiple other occurrences in https://github.com/pinojs/pino/blob/master/pretty.js.

jsumners:
Hmm, I was prepared to shoot this down, but it looks like the spec allows for `\r\n`:

http://specs.okfnlabs.org/ndjson/
> Each JSON text MUST conform to the [RFC7159] standard and MUST be written to the stream followed by the newline character \n (0x0A). The newline charater MAY be preceeded by a carriage return \r (0x0D). The JSON texts MUST NOT contain newlines or carriage returns.

Would you like to submit a PR to enable this option?

tniessen:
And by multiple other occurrences in https://github.com/pinojs/pino/blob/master/pretty.js

davidmarkclements:
any tools built around pino at this point may be relying on just `\n` - in there interest of not breaking ecosystem libs let's make this an opt-in setting

tniessen:
@davidmarkclements Yes, #327 makes it completely optional. I would actually prefer to always use `\n` but some Windows tools were designed for `\r\n`.

PR #327, opened 2017-11-02
Add crlf support

tniessen:
This is an attempt to add support for CRLF line endings.

Fixes: https://github.com/pinojs/pino/issues/326

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -77,6 +77,7 @@
     See [Extreme mode explained](extreme.md) for more detail.
   * `enabled` (boolean): enables logging. Default: `true`
   * `browser` (Object): browser only, may have `asObject` and `write` keys, see [Pino in the Browser](../readme.md#browser)
+  * `crlf` (boolean): emit `\r\n` instead of `\n`. Default: `false`.
 + `stream` (Writable): a writable stream where the logs will be written.
   It can also receive some log-line [metadata](#metadata), if the
   relative protocol is enabled. Default: `process.stdout`
@@ -113,6 +114,7 @@ Returns a new [logger](#logger) instance.
   message. Default: `msg`.
   * `forceColor` (boolean): if set to `true`, will add color information to the formatted output
   message. Default: `false`.
+  * `crlf` (boolean): emit `\r\n` instead of `\n`. Default: `false`.
 
 ### Example:
 ```js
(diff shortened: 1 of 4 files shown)


docs/API.md
```diff
@@ -77,6 +77,7 @@
     See [Extreme mode explained](extreme.md) for more detail.
   * `enabled` (boolean): enables logging. Default: `true`
   * `browser` (Object): browser only, may have `asObject` and `write` keys, see [Pino in the Browser](../readme.md#browser)
+  * `crlf` (boolean): emit `\r\n` instead of `\n`. Default: `false`.
```
davidmarkclements: instead of "emit" can we say:

logs newline delimited JSON with `\r\n` instead of `\n`

mcollina:
can you add a unit test for this?

docs/API.md
tniessen: Changed, PTAL

test/crlf.test.js
```diff
@@ -0,0 +1,42 @@
+'use strict'
+
+var test = require('tap').test
+var pino = require('../')
+var Writable = require('stream').Writable
+
+class LogWriteStream extends Writable {
```
davidmarkclements: I realise it's just test, but we don't use classes - take a look at the other tests 
for how we do this (helpers.sink)

tniessen:
@davidmarkclements I removed the class and added a test for `pretty`.

docs/API.md
```diff
@@ -113,6 +114,7 @@ Returns a new [logger](#logger) instance.
   message. Default: `msg`.
   * `forceColor` (boolean): if set to `true`, will add color information to the formatted output
   message. Default: `false`.
+  * `crlf` (boolean): emit `\r\n` instead of `\n`. Default: `false`.
```
jsumners: This language should match the language in the options for the regular constructor.

tniessen: I changed the wording above to

> logs newline delimited JSON with `\r\n` instead of `\n`. Default: `false`.

as suggested by @davidmarkclements, but this is not necessarily true for `pretty` (which can produce arbitrary strings, not just JSON), right?

jsumners: Good call. Ignore me.

PR #332, opened 2017-11-14
Do not throw if the passed stream is stdout

mcollina:
improvement on https://github.com/pinojs/pino/pull/321


mcollina:
releasing

PR #333, opened 2017-11-17
Fix bug added to pretty after #324 merge

taylon:
It seems that after #324 was merged a bug was introduced to pretty.
It would consider lines missing hostname and pid invalid, which is not true after #324 

I'm not familiar with the code at all so not sure it detecting a pino line only with "v" is ok.
Let me know if there is anything that needs to change and I'll happily adapt =D

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -46,10 +46,7 @@ function filter (value, messageKey, eol) {
 }
 
 function isPinoLine (line) {
-  return line &&
-    line.hasOwnProperty('hostname') &&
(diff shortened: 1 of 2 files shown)


jsumners:
Ouch. I forgot about that. I'm not sure checking only for `v` is sufficient, but it may be the only choice since it's basically the only static property left. I wonder if we could add a `pino: true` property when the base log is altered from the default? What do you think @davidmarkclements @mcollina?

PR #335, opened 2017-11-28
Fix typo in `withSpaces` function

zenflow:
I wondered why using the "pretty" feature didn't make my logs look that pretty.

Then I looked at the code and found out why: It looks like someone passed a string to `value.split()` when they clearly meant to pass a regex.

Tested this fix with my project, and it looks much better.

Before fix:
```
[2017-11-28T21:45:08.685Z] INFO (16172 on MIS7501): incoming request
    reqId: 1
    req: {
  "id": 1,
  "method": "GET",
  "url": "/api/auth/getCurrentUser",
  "remoteAddress": "::1",
  "remotePort": 58871
}
```

After fix:
```
[2017-11-28T21:46:25.319Z] INFO (15812 on MIS7501): incoming request
    reqId: 1
    req: {
      "id": 1,
      "method": "GET",
      "url": "/api/auth/getCurrentUser",
      "remoteAddress": "::1",
      "remotePort": 58881
    }
```

### DIFF
--- a/pretty.js
+++ b/pretty.js
@@ -24,7 +24,7 @@ var standardKeys = [
 ]
 
 function withSpaces (value, eol) {
-  var lines = value.split('\r?\n')
+  var lines = value.split(/\r?\n/)
   for (var i = 1; i < lines.length; i++) {
     lines[i] = '    ' + lines[i]
   }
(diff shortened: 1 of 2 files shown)


davidmarkclements:
very nice, thank you! once we have a test for this, we'll merge straight away

zenflow:
Added a test that would have caught this regression :)

Issue #339, opened 2017-12-18
Broken test on Node 9


mcollina:
```
   # Subtest: terminates on SIGHUP when no other handlers registered
        1..2
        not ok 1 - should be equal
          ---
          found: 1
          wanted: 0
          compare: ===
          at:
            line: 100
            column: 7
            file: test/events.test.js
            type: ChildProcess
          stack: |
            ChildProcess.<anonymous> (test/events.test.js:100:7)
          source: |
            t.is(code, 0)
          ...
        
        ok 2 - should not be equal
        # failed 1 of 2 tests
    not ok 5 - terminates on SIGHUP when no other handlers registered # time=1031.376ms
```

from https://travis-ci.org/pinojs/pino/jobs/318109604.

PR #340, opened 2017-12-18
Fixed tests on Node 9

mcollina:
Fixes https://github.com/pinojs/pino/issues/339

### DIFF
--- a/test/events.test.js
+++ b/test/events.test.js
@@ -5,6 +5,7 @@ var path = require('path')
 var writeStream = require('flush-write-stream')
 var fork = require('child_process').fork
 var spawn = require('child_process').spawn
+var semver = require('semver')
 var fixturesPath = path.join(__dirname, 'fixtures', 'events')
 
 test('no event loop logs successfully', function (t) {
@@ -97,7 +98,11 @@ test('terminates on SIGHUP when no other handlers registered', function (t) {
   }))
 
   child.on('exit', function (code) {
-    t.is(code, 0)
+    if (semver.gt(process.version, '9.0.0')) {
+      t.is(code, 1)
+    } else {
+      t.is(code, 0)
+    }
   })
 
   child.on('close', function () {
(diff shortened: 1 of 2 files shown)


test/events.test.js
```diff
@@ -97,7 +98,11 @@ test('terminates on SIGHUP when no other handlers registered', function (t) {
   }))
 
   child.on('exit', function (code) {
-    t.is(code, 0)
+    if (semver.gt(process.version, '9.0.0')) {
+      t.is(code, 1)
```
jsumners: Why would the return code suddenly be `1` on Node >=9.0.0?

mcollina: I have no idea. However I have reproduced it even without pino, if you send SIGHUP on Node 9, you get an exit code of 1.

jsumners: There's something off here. When I run the following script with 9.3.0 and send it `kill -HUP`, I get an exit code of `0`:

```js
'use strict'

process.on('SIGHUP', () => {
  console.log(arguments)
  process.exit()
})

setInterval(() => {}, 100)
```

output:
```js
{ '0': {},
  '1':
   { [Function: require]
     resolve: { [Function: resolve] paths: [Function: paths] },
     main:
      Module {
        id: '.',
        exports: {},
        parent: null,
        filename: '/private/tmp/19/foo.js',
        loaded: true,
        children: [],
        paths: [Array] },
     extensions: { '.js': [Function], '.json': [Function], '.node': [Function] },
     cache: { '/private/tmp/19/foo.js': [Module] } },
  '2':
   Module {
     id: '.',
     exports: {},
     parent: null,
     filename: '/private/tmp/19/foo.js',
     loaded: true,
     children: [],
     paths:
      [ '/private/tmp/19/node_modules',
        '/private/tmp/node_modules',
        '/private/node_modules',
        '/node_modules' ] },
  '3': '/private/tmp/19/foo.js',
  '4': '/private/tmp/19' }
```

mcollina:
Closed in favour of https://github.com/pinojs/pino/pull/342.

PR #351, opened 2018-02-02
Updated copyright year to 2018

mcollina:
As titled


mcollina:
CI is failing, I think for some updated dependency that we use on the browser side. 
@davidmarkclements can you please check?

davidmarkclements:
struggling to reproduce this locally

PR #354, opened 2018-02-06
Ignore domain related error properties

natanavra:
Although it is deprecated, these properties are still being added to the error objects: https://nodejs.org/api/domain.html#domain_additions_to_error_objects

This leads to confusing long, unreadable error logs on syntax errors and such (e.g.TypeError "Cannot read property 'x' of undefined")

This ignores these properties on the standard error serializer.

### DIFF
--- a/lib/serializers.js
+++ b/lib/serializers.js
@@ -36,7 +36,7 @@ function asErrValue (err) {
     stack: err.stack
   }
   for (var key in err) {
-    if (obj[key] === undefined) {
+    if (obj[key] === undefined && !key.startsWith('domain')) {
       obj[key] = err[key]
     }
   }
--- a/test/error.test.js
+++ b/test/error.test.js
@@ -79,7 +79,7 @@ test('type, message and stack should be first level properties', function (t) {
 
 test('err serializer', function (t) {
   t.plan(2)
-  var err = Object.assign(new Error('myerror'), {foo: 'bar'})
+  var err = Object.assign(new Error('myerror'), {foo: 'bar', domain: {}, domainEmitter: {}, domainThrown: true, domainBound: function () {}})
   var instance = pino({
     serializers: {
       err: pino.stdSerializers.err


mcollina:

lib/serializers.js
```diff
@@ -36,7 +36,7 @@ function asErrValue (err) {
     stack: err.stack
   }
   for (var key in err) {
-    if (obj[key] === undefined) {
+    if (obj[key] === undefined && !key.startsWith('domain')) {
```
davidmarkclements: is this the fastest way to do this?

davidmarkclements:
@natanavra could you please add a benchmark comparison - specifically for an error object that has an enumerable property that *isn't* "domain" and check the performance cost before and after,
so a benchmark logging: `var err = Object.assign(Error(), {domailing: 'aaaaah'})`

There's also two problems I can see here:

1. this approach will also mean, that users can't log error objects with any keys starting with domain. "Domain" is quite a loaded word, and errors are often embellished with additional properties (e.g. ([http-errors](http://npm.im/http-errors)

2. Even though domains are deprecated and will be removed, there's still a lot of legacy users of domains in enterprise, who will probably use an ecosystem module replacement built on async hooks. They may want to retain that info.

natanavra:
@davidmarkclements Regarding a benchmark, I'm not sure it's relevant since it will only happen on error logs and IMO optimizing this is over-engineering.

Regarding the two problems:
1) If a user wants to include a domain, he can, just not part of the error object.
2) The domain as I understand is environment additional data, with a lot of cyclical dependencies, so you'd want to avoid printing it anyway (very long output filled with nulls)

davidmarkclements:
hey @natanavra - that depends on how many error objects an application may log, which depends on how an application uses error objects. It's probably not a major use case but we have a philosophy of finding the fastest approach for each logic path – which we need benchmarks for

To address your points

1. Yes, but no one is going to know that. If anyone attaches a key that starts with `domain` to an error object, it's going to disappear. For instance consider a case where a server responds to requests across multiple domains (or subdomains). And when there's an error they attach `domain` to the Error object for reference. It's going to inexplicably disappear. At the very least we need to document this, or find another way – e.g. check the actual domain keys and objects 

2. I do agree – it's better to filter this out by default we just need to make sure the approach is optimum for ergonomics as well as performance

jsumners:
A better solution may be to provide an alternate standard error serializer that does this work. User's will be able to opt-in to any downsides by simply `pino({ serializers: {err: pino.domainErr} })`.

natanavra:
@jsumners Agreed, this could be the silver lining for this case, we'll add another serializer and update the documentation.
@davidmarkclements Benchmark can be found here: https://jsperf.com/filter-object-keys-in-a-for-loop we have a 20% speed difference when testing only for the keys found in the nodejs documentation.

davidmarkclements:
cool! glad we've gotten a faster way

what we need though in terms of benchmarks, is a comparison of https://github.com/pinojs/pino/tree/master/benchmarks before and after the change – what impact on perf does this for our benchmark cases - if you run both sets on your machine and post here for comparison that helps us to assess

are we saying the pino.domainErr serializer *removes* the domains or *preserves* them?

jsumners:
@natanavra would you like to add this an optional serializer in https://github.com/pinojs/pino-std-serializers ?

jsumners:
I'm going to close this due to inactivity. If you would like to continue this work, please re-open or send as a PR for `pino-std-serializers`.

PR #358, opened 2018-02-16
Fix bin not waiting for stdin to close before exiting

### LINKED ISSUE #357: piping pino break servers that catch sigint when using ctrl-c
If you pipe `pino` to get a nicely formatted output for your logs using ctrl-c can cause the server
to throw an exception.
This is because the shell will send SIGINT to both pino and your server code.
pino will immediately exit but if your server is trying to do some graceful termination and logging in he process it will crash since pino already closed stdout.

Example server code -
```js
'use strict'

const pino = require('pino')()

pino.info('starting server')
const interval = setInterval(function () {
	pino.info('server running')
}, 1000);


process.on('SIGINT', () => {
	pino.info('server stopping server')
	setTimeout(() => {
		clearInterval(interval)
		pino.info('server stopped')
	}, 1000)
})
```

if you run `node server.js | pino`
and terminate the process using ctrl-c the following error is thrown -
```shell
events.js:183
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at _errnoException (util.js:1024:11)
    at WriteWrap.afterWrite [as oncomplete] (net.js:867:14)
```

yonjah:
Fix #357 

I don't think this should break anything since if stdout ends, process will immediately exit.
The process will wait for 2 sec before terminating. I think this should be enough time for most processes to exit but a longer timeout might be required.

This feature (and the timeout) can be implemented behind flags but I wanted to keep the usage simple

### DIFF
--- a/bin.js
+++ b/bin.js
@@ -3,6 +3,7 @@
 'use strict'
 
 var pretty = require('./pretty')
+var fs = require('fs')
 
 module.exports = pretty
 
@@ -17,20 +18,13 @@ if (arg('-h') || arg('--help')) {
     forceColor: arg('-c'),
     messageKey: messageKeyArg()
   })).pipe(process.stdout)
-  process.stdin.on('end', closeProcess)
-  process.on('SIGINT', waitClose)
-}
-
-function waitClose () {
-  setTimeout(closeProcess, 2000)
-}
-
-function closeProcess () {
-  process.exit()
+  if (!process.stdin.isTTY && !fs.fstatSync(process.stdin.fd).isFile()) {
+    process.once('SIGINT', function noOp () {})
+  }
 }
 
 function usage () {
-  return require('fs')
+  return fs
       .createReadStream(require('path').join(__dirname, 'usage.txt'))
 }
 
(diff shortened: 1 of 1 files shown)


mcollina:
I think you can avoid the timeout with the following rules:

1. if `fs.fstatSync(process.stdin.fd).isFile()` returns `true`, you can exit immediately
2. if `process.stdin.isTTY` is `true`, you can exit immediately
3. if both the above are `false` it means that we are piping, and you can do nothing _on the first SIGINT_ (waiting for the upstream to finish), and then exit immediately on the second SIGINT.

bin.js
```diff
@@ -17,6 +17,16 @@ if (arg('-h') || arg('--help')) {
     forceColor: arg('-c'),
     messageKey: messageKeyArg()
   })).pipe(process.stdout)
+  process.stdin.on('end', closeProcess)
```
jsumners: It seems to me you could just pass `process.exit` in both instances instead of creating a redundant `closeProcess` function.

mcollina:
@yonjah by any chance, do you have a Windows machine available to check this? I'm traveling until the end of the week, and I would like to make sure this works as expected there too.

I know some colleagues had issues with SIGINT and Windows, I'll do some research as well.

yonjah:
@mcollina Sorry not really. 
I'm not sure I'll even know how to pipe commands on windows.

Saw a few projects using https://www.appveyor.com/ for windows ci so might be useful

mcollina:
My colleagues confirms it’s ok on Windows.

PR #364, opened 2018-02-22
Refactor to use pino-std-serializers

jsumners:
This PR removes the bundled serializers and replaces them with the equivalents from the `pino-std-serializers` module.


davidmarkclements:
really great work @jsumners

Issue #362, opened 2018-02-20
How to serialize the whole object, not properties?


PavelPolyakov:
Hello,

Please, advice what is expected way to achieve the desired behaviour.

I use `pino` through fastify, therefore I use it usually through the request directly:
```
request.log.info('message');
request.log.error(error);
```

Imagine the next code:
```js
const axios = require('axios');

(async () => {
    try {
        const r = await axios.get('https://google.com/404');
    }catch(error) {
        console.log(error);
    }
})()
```

`axios` will throw the `error` with the GIANT object, like:
https://pastebin.com/ZXccMRvw

`pino` will try to log it, while I want to work on the serialisation of the logged object.

BUT, serializers object supports only per property serialization. In my case these are three properties `request`, `response`, `config` only the fact that three of them are present is the evidence of the `axios` error. 

> serializers (object): an object containing functions for custom serialization of objects. These functions should return an JSONifiable object and they should never throw. When logging an object, each top-level property matching the exact key of a serializer will be serialized using the defined serializer.

How can I put the logic on top three of them together, not one by one? Let's say if only `config` is present - I want to log it as is.

I actually achieved it with `prettyPrint`, but when I tried to test this using unit-tests, and declared the custom write stream - pino init fails, since `prettyPrint` is not available if you are not using `process.stdout`.

What can be done about it?

Regards,

jsumners:
This is the same thing as https://github.com/pinojs/pino-http/pull/42 (https://github.com/pinojs/pino-http/pull/43). The solution is to split out the standard serializers into `pino-std-serializers`. This will also solve #265.

I'll start work on this.

PavelPolyakov:
@jsumners 
I might be wrong, but it seems that you are talking about placing the serializers aside, into the separate module. (how would it help? in `axios` error the property is called `request`)

While my question is more like - maybe we should have the control over the whole object serialisation. Like providing the option to register serializer called `*`  (or array of them) which will handle the whole object?

jsumners:
1. Pino would be updated to use the `pino-std-serializers` package to supply the fixed serializers.
2. Fastify would be updated to use the new version of Pino

The solution is merely centralizing this code so that we don't have to update it in 15 places every time someone finds a bug.

jsumners:
As for:

> While my question is more like - maybe we should have the control over the whole object serialisation. Like providing the option to register serializer called * (or array of them) which will handle the whole object?

[...]

Issue #355, opened 2018-02-08
Data property for stringified objects


grantcarthew:
Hi guys,

I have been reading the code and can not see a method for placing the serialized objects into a root log property.

As an example, let say I run the following:
```js

pino.debug({ a: 1, b:2 })

```

With the above log statement it will produce the following:

```js
{
  level: 20,
  pid: 1234,
  time: 1234,
  v: 1,
  hostname: 'host',
  msg: 'some message',
  name: 'some name',
  a: 1,
  b: 2
}
```

What I would like to achieve is this:

```js
{
  level: 20,
  pid: 1234,
  time: 1234,
  v: 1,
  hostname: 'host',
  msg: 'some message',
  name: 'some name',
  data: { a: 1,  b: 2 }
}
```

I know I could call the log function passing in an object with a data property each time however this seems a little verbose. I know I could also wrap the pino logger in a custom logger to achieve the same thing.

Is there a way of putting the JSON objects into a root property using pino? A catch all serializer for example?

davidmarkclements:
@grantcarthew no we currently don't support that, and it could be logged more performantly with a root key if implemented in pino core with string concatenation. It would require at least one conditional check for a different logic branch, which shouldn't affect performance in any significant way (unless it puts us over the complexity threshold). So if you'd like to PR you're more than welcome, I can see how this would be useful

grantcarthew:
Ok, thanks @davidmarkclements 
What would be the best solution?
A catch all serializer or an option?
I don't think I would be the only one interested in this in which case an option would be better?

grantcarthew:
I decided to go with an option. Just writing some tests. Will send a PR for initial review when I get time.

PR #365, opened 2018-02-23
introducing global serializer

PavelPolyakov:
Resolves
https://github.com/pinojs/pino/issues/362

Please, let me know if I missed anything

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -45,6 +45,8 @@
     of objects. These functions should return an JSONifiable object and they
     should never throw. When logging an object, each top-level property matching the exact key of a serializer
     will be serialized using the defined serializer.
+        
+    Alternatively, it is possible to register a serializer under the key `Symbol.for('pino.*')` which will act upon the complete log object, i.e. every property.
   * `timestamp` (boolean|function): Enables or disables the inclusion of a timestamp in the
     log message. If a function is supplied, it must synchronously return a JSON string
     representation of the time, e.g. `,"time":1493426328206 (which is the default).
--- a/pino.js
+++ b/pino.js
@@ -147,6 +147,10 @@ function asJson (obj, msg, num) {
     if (objError) {
       data += ',"type":"Error","stack":' + this.stringify(obj.stack)
     }
+    // if global serializer is set, call it first
+    if (this.serializers[Symbol.for('pino.*')]) {
+      obj = this.serializers[Symbol.for('pino.*')](obj)
+    }
     for (var key in obj) {
       value = obj[key]
       if ((notHasOwnProperty || obj.hasOwnProperty(key)) && value !== undefined) {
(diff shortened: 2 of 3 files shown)


docs/API.md
```diff
@@ -45,6 +45,9 @@
     of objects. These functions should return an JSONifiable object and they
     should never throw. When logging an object, each top-level property matching the exact key of a serializer
     will be serialized using the defined serializer.
+    
+    **USE ON YOUR OWN RISK**
```
jsumners: Why this warning?

docs/API.md
```diff
@@ -45,6 +45,9 @@
...
+    As an exception it is possible to register `Symbol.for('*')` serializer, which accepts the complete log object (`pid`, `hostname` and all custom properties) and returns the updated complete log object.
```
jsumners: "Alternatively, it is possible to register a serializer under the key `Symbol.for('*')` which will act upon the complete log object, i.e. every property."

PavelPolyakov: updated

davidmarkclements:
Symbol.for is a global namespace - over time collisions (especially against something like *) willl become likely. In general we should prefix Symbol.for namespaces with `pino.` - eg `Symbol.for('pino.*')`

Also could you please post before/after benchmark results here

PavelPolyakov:
I used `Symbol.for('*')` alike the [metadata one's usage](https://github.com/pinojs/pino/blob/master/docs/API.md#metadata) . I think probability that anyone would have `Symbol.for('*')` in the object they are going to serialise is the same as having `Symbol.for('pino.*')`.

Here are the results on my MacPro mid 2015:
`master`: https://pastebin.com/tzjKiwv7
`feature/global-serializer`: https://pastebin.com/2bDWnrmM

davidmarkclements:
Yeah see I disagree with metadata usage – I think it needs to be prefixed as well. 

@mcollina @jsumners – agree/disagree? 

Re benchmarks they look good!

jsumners:
I'm good with namespacing the symbols.

PavelPolyakov:
@davidmarkclements @jsumners 
I can bring the namespacing as part of this PR as well, but it'd mean that we need to bump the version (from 4 to 5), since it'll not be backward compatible.

PavelPolyakov:
if you want me to do it, please, approve that this can be done in scope of this PR.

We also can somehow make the `Symbol.for('needsMetadata')` usage deprecated, till the next version is not released. BUT I don't have any good idea how to do that, since I assume we can not write any deprecation message to the console. Maybe just deprecate it through the documentation?

mcollina:
I think we should leave `Symbol.for('needsMedata')` as is, and then namespace the new one. So we can release this as a minor. Also, open an issue about namespacing `Symbol.for('needsMedata')` so that we remember to do it for v5.

PavelPolyakov:
Good, I've updated the new `Symbol`

mcollina:
@jsumners are you good with this?

mcollina:
Thanks for all the hard work @PavelPolyakov!

Landing now.

grantcarthew:
For reference: this PR is a solution to this issue https://github.com/pinojs/pino/issues/355.

Issue #367, opened 2018-02-25
CI checking Node 7?


jsumners:
https://github.com/pinojs/pino/blob/04351416cd4d1723fbc872fa37c87e453731298d/.travis.yml#L3-L8

Is there a reason we continue to check against Node 7?

davidmarkclements:
no, I think it should be removed – as should 9 when 10 is released

mcollina:
I would keep it until it fails / or it become a problem.
I think we should drop both 7 and 9 when/if we bump the major.
Il giorno dom 25 feb 2018 alle 15:53 David Mark Clements <
notifications@github.com> ha scritto:

rimiti:
I'm totally agree to remove node 7 from CI. 
+1

jsumners:
Resolved by #370

PR #370, opened 2018-03-14
chore(.travis.yml) node 7 removed from CI, yaml indented

rimiti:
Hello,

Issue: #367 

**Changes:***

- Node 7 removed from Travis CI (only Node LTS version)
- Yaml indented

Regards,
Dimitri,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,12 @@
 language: node_js
 sudo: false
 node_js:
-- '4'
-- '6'
-- '7'
-- '8'
-- '9'
+  - 4
+  - 6
+  - 8
+  - 9
 script:
-- npm run ci
+  - npm run ci
 env:
   global:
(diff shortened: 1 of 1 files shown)


mcollina:
We will merge this when we release a new major version of pino.

jsumners:
I am unclear why we need to wait for a new major version. Node 7 has never been a production version of Node.

mcollina:
My logic is “a node version never gets removed if not in a major release”.

jsumners:
But can users of Node 7 really expect LTS-like support from modules?

davidmarkclements:
I mean... no - unless those modules state that.. and that might be a really nice thing to say about Pino

jsumners:
I think we need this document https://github.com/fastify/fastify/blob/master/docs/LTS.md

watson:
Though not LTS, I think it's very valuable to always test against the latest version of Node.js. If not there might be some nasty surprises in wait when v10 lands. And even when v10 lands it will not be LTS for a while.

jsumners:
@watson yes, but 7 is not 10.

rimiti:
@jsumners I'm totally agreed with.
If you see all biggest node projects, none of them test the version 7...

davidmarkclements:
No, I see @mcollina s logic. It's still more sensible to not break anything unless you go to major. If we stop testing 7 there is a possibility that something could break in 7. So whether people should be using it or not, if something breaks then we should be releasing a major.

watson:
@jsumners I was referring to v9 which I don't think should be removed from `.travis.yml`

jsumners:
Oh, I see. Yes, you are correct.

rimiti:
Hello, 

Sorry for the delay, I've added Node 9.
It can be merge now ;)

Dimitri,

davidmarkclements:
this will be merged prior to releasing our next major - thank you for the pr @rimiti !

rimiti:
You're welcome ;)

mcollina:
I've changed the base branch from master to next-major. So this can land now.

.travis.yml
```diff
@@ -1,13 +1,12 @@
 language: node_js
 sudo: false
 node_js:
-- '4'
-- '6'
-- '7'
-- '8'
-- '9'
+  - 4
```
mcollina: can you remove Node 4 as well? It won't be supported anymore in a month, so it's better dropping support from pino@5.

rimiti:
It's done ;)

PR #372, opened 2018-03-18
Issue #371, add print all nested props for errorProps flag

### LINKED ISSUE #371: Improving pino cli --errorProps
I really appreciate this new feature but I recognized that it does work only with non-nested properties.
If the additional property is an nested object the output is errors: [object Object].
It would be nice if this gets also parsed and prettyfied.

Also it would be quite nice if one could simply pass --errorProps * as argument to print any extra property. (In case I do not exactly know what library XY adds to the base Error object. )

ilyaztsv:
I've added `*` property for `errorProps` flag to resolve #371 

Glad to your feedback 😃 

### Result view

`log.log`:

```
{"level":50,"time":1457537229339,"msg":"Error message.","pid":44127,"hostname":"MacBook-Pro-3.home","type":"Error","stack":"Stack of the error","statusCode":500,"dataBaseSpecificError":{"errorType":"some database error type","erroMessage":"some database error message","evenMoreSpecificStuff":{"someErrorRelatedObject":"error"}},"v":1}
```

screen shot 2018-03-18 at 19 45 06

screen shot 2018-03-18 at 19 45 35

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -107,7 +107,7 @@ statusCode: 500
 
 In order to print all nested properties of `Error` objects, you can use `--errorProps` flag with `*` property.
 
-Please note, that you need to escape `*` character (`'*'` or `\*`), because `*` is a wildcard in Bash.
+Note: you must quote or escape the `*` (asterisk) to avoid shell expansion.
 
 `pino --errorProps '*'` will transform:
 
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -79,7 +79,7 @@ If an instance of `Error` is logged, Pino adds `"type":"Error"` to the logged JS
 Thus, when prettifying the output, Pino will transform the JSON:
 
 ```js
-{"level":50,"time":1457537229339,"msg":"Error message.","pid":44127,"hostname":"MacBook-Pro-3.home","type":"Error","stack":"Stack of the error","statusCode":500,"v":1}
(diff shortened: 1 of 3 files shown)


docs/cli.md
```diff
@@ -103,5 +103,29 @@ To:
...
     Stack of the error
 statusCode: 500
-v: 1
-```
\ No newline at end of file
+```
+
+In order to print all nested properties of `Error` objects, you can use `--errorProps` flag with `*` property.
+
+Please note, that you need to escape `*` character (`'*'` or `\*`), because `*` is a wildcard in Bash.
```
jsumners: I think this can be struck completely. Either that, or rephrase to: "Note: you must quote or escape the `*` (asterisk) to avoid shell expansion."

ilyaztsv: changed to 
> "Note: you must quote or escape the * (asterisk) to avoid shell expansion."

pretty.js
```diff
@@ -51,9 +51,14 @@ function withSpaces (value, eol) {
   return lines.join(eol)
 }
 
-function filter (value, messageKey, eol) {
+function filter (value, messageKey, eol, isExcludeStandardKeys) {
```
jsumners: `isExcludeStandardKeys` would make more sense as `excludeStandardKeys`.

ilyaztsv: rename to `excludeStandardKeys`

PR #377, opened 2018-03-27
std serializers api typo + deprecation unit tests

davidmarkclements:
this is a no brainer to fix, 

however: I'm assuming this is a patch release not a major, because even though we're technically changing the API - we should consider the current API broken

I doubt anyone is relying on the typo, we would have had a PR by now... surely

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,8 +20,8 @@
   ],
   "scripts": {
     "browser-test": "zuul tape test/browser*test.js --local",
-    "test": "standard | snazzy && tap --no-cov test/*test.js",
-    "ci": "standard | snazzy && tap --cov test/*test.js",
+    "test": "standard | snazzy && NODE_OPTIONS='--no-warnings' tap --no-cov test/*test.js",
+    "ci": "standard | snazzy &&  NODE_OPTIONS='--no-warnings' tap --cov test/*test.js",
     "bench-all": "node benchmarks/runbench all",
     "bench-basic": "node benchmarks/runbench basic",
     "bench-object": "node benchmarks/runbench object",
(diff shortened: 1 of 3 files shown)


davidmarkclements:
can we get another round, I've introduced an approach to deprecating with tests

davidmarkclements:
(this also increases coverage, tests the slowtime deprecation)

davidmarkclements:
failing in v4 because `emitWarning` isn't on `process`. 

That means technically we've been failing on v4 since slowtime was deprecated.

jsumners:
What do you guys want to do about v4? I would be okay with dropping support early since there is only a month of LTS left on it. But it's also possible to detect if `emitWarning` is unavailable and simply use `emit`.

davidmarkclements:
but we'd also have to do `process.on('warning', console.error)` 

we could just switch to using `util.deprecate` - which behaves differently in v4

it will make testing trickier but thats the least leaky approach in `pino.js`

jsumners:
Again, I'm okay with whichever route.

davidmarkclements:
I'd prefer to just drop support for v4

realistically, anyone on v4 will still be able to use current pino for at least the next month,
we just stop testing against it early

@mcollina what do you think?

davidmarkclements:
Although if we drop support for v4 it's a major release - and none of these current deprecation issues matter

mcollina:
https://nodejs.org/api/util.html#util_util_deprecate_function_string was added in Node 0.8. It's pretty safe to use.

davidmarkclements:
but much harder to test across versions because it behaves differently

However, it's probably the right approach, otherwise we have to start talking a major release

davidmarkclements:
ok, switched to util.deprecate tests are passing in every version

PR #385, opened 2018-04-04
added lastTime meta parameter

mcollina:
@jsumners this is needed to properly use the meta in `pino-pretty`.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -129,12 +129,12 @@ Object.defineProperty(
   {value: LOG_VERSION}
 )
 
-function asJson (obj, msg, num) {
+function asJson (obj, msg, num, time) {
   // to catch both null and undefined
   var hasObj = obj !== undefined && obj !== null
   var objError = hasObj && obj instanceof Error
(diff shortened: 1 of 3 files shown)


pino.js
```diff
@@ -208,13 +208,15 @@ Object.defineProperty(pinoPrototype, 'child', {
...
-  var s = this.asJson(obj, msg, num)
+  var t = this.time()
+  var s = this.asJson(obj, msg, num, t)
   var stream = this.stream
   if (this.cache === null) {
     if (stream[needsMetadata]) {
       stream.lastLevel = num
       stream.lastMsg = msg
       stream.lastObj = obj
+      stream.lastTime = t.slice(8)
```
davidmarkclements: is there a chance that `t` won't be a string (does `this.time` originate externally?)

mcollina: No, we are assuming it's a string everywhere anyway.

davidmarkclements: well then, LGTM

PR #386, opened 2018-04-05
CLI revamp

jsumners:
Here it is: the "revamp" of the CLI. Or rather, the complete removal of it. Notice that the target branch of this PR is the `next-major` branch ***not*** the `master` branch.

This PR:

1. Removes the CLI completely and reduces core to just a logger
2. Introduces the concept of Pino prettifier modules

## Prettifier Modules

A prettifier module is one like [`pino-pretty`](https://github.com/pinojs/pino-pretty). Such modules export a factory function that accepts a configuration object. That function returns a formatter function that can accept an ndJSON log line or a Pino log object. That function also includes an `asMetaWrapper(destStream)` method attached to it which returns a metadata API compatible stream object. This allows for users to do:

```js
const log = require('pino')({
  prettyPrint: {
    someFooPrettifierOption: 'and-value'
  },
  prettifier: 'foo-prettifier'
})
```

This means people like @lrlna could update their awesome modules to be used directly in-process. 🎉

## Semver Major 🚨🚨🚨 😱

I realize we would prefer not to issue a new major. But we have a few things queued up just waiting for us to be brave enough for a new major -- https://github.com/pinojs/pino/projects/1 . The majority of the recent issues deal with the prettifier and CLI. Those are not the problems Pino strives to solve. So by removing them from core and issuing a new major version we will be reducing Pino down to its purpose: generate logs. And I think the logger features are mature enough that v5.y.z will be with us for a *long* time: we just have to make sure the prettifier API is solid.


pino.js
```diff
@@ -272,6 +272,17 @@ Object.defineProperty(pinoPrototype, 'isLevelEnabled', {
   value: isLevelEnabled
 })
 
+function getPrettyStream (opts) {
+  var prettyFactory
+  var prettyModule = opts.prettifier || 'pino-pretty'
+  try {
+    prettyFactory = require(prettyModule)
+    return prettyFactory(opts).asMetaWrapper(process.stdout)
```
mcollina: I think we might as well include asMetaWrapper in here instead. I do not see a reason why a prettifier might not want that.

pino.js
```diff
@@ -272,6 +272,17 @@ Object.defineProperty(pinoPrototype, 'isLevelEnabled', {
   value: isLevelEnabled
 })
 
+function getPrettyStream (opts) {
+  var prettyFactory
+  var prettyModule = opts.prettifier || 'pino-pretty'
+  try {
+    prettyFactory = require(prettyModule)
```
mcollina: I would not require a string, but rather support a function (the factory) or true. If true, we can then require.

pino.js
jsumners: I don't understand this statement. Can you reword to clarify?

pino.js
jsumners: That makes sense.

davidmarkclements:
this is a great piece of work @jsumners 

I think we want to be careful not to implicitly direct users to opt for in-process prettifying unless they have no option but to do so

I think we also need: 

* a minimal CLI tool under the `pino` command name that does very minimal prettifying
* a  *gatekeeping/signposting* CLI under the `pino` command name that lists available prettifiers and installs on demand 
* a mixture of both

mcollina:
> a minimal CLI tool under the pino command name that does very minimal prettifying

I'm -1 to that.

I think we can just ship a `pino` executable with `pino-pretty`

jsumners:
@davidmarkclements I agree with @mcollina. The point here is that including the CLI in the core logger wasn't such a great plan.

The reason for maintaining the `prettyPrint` option is to reduce the breakage. I just opted to make it a bit more robust and turn it into a full fledged API for generic prettifiers. This is basically a "reporters" feature like with `tap` or https://github.com/nuxt/consola#reporters .

Also, users have to actually read the documentation to learn how to do this. And we know how well they do that 😆

davidmarkclements:
we *just* got people used to the idea of piping to the `pino` command... 

if we no longer bundle (and I'm not for bundling pino-pretty btw), then were setting up a learning flow where users will assume the right thing to do is use the programmatic API with pino-pretty or other prettifier.

mcollina:
> if we no longer bundle (and I'm not for bundling pino-pretty btw), then were setting up a learning flow where users will assume the right thing to do is use the programmatic API with pino-pretty or other prettifier.

I do not understand.

davidmarkclements:
I think we add a section right at the beginning, something to the effect of 

We never want to prettify in production. The best way to prettify is to do so in a separate process and pipe the log output to that process. This precludes any need to have separate development/production logger configurations, and familiarizes us with the concept of always processing log output in a separate process.
 
The recommended way to prettify is to install a prettifier, and pipe output to it. 
For instance, with pino-pretty we would achieve this like so:

```sh
npm i -g pino-pretty
```

```
node app.js | pino
```

Something to that effect, prioritizing command line based prettifying

davidmarkclements:
and probably put that in the main readme as well, then for programatic usage link to the pretty doc (which also enforces command line piping)

davidmarkclements:
@mcollina based on my follow up, does my prior comment make any more sense?

jsumners:
@davidmarkclements I just pushed up a change that, without even having read your suggestion 🙂, includes a primary example of piping. Give it a review. Also, feel free to push up your ideas to this branch.

docs/pretty.md
```diff
@@ -0,0 +1,77 @@
+# Pretty Printing
+
+By default, Pino log lines are newline delimited JSON. This is perfect for
+production usage and long term storage. It's not so great for development
+environments. Thus, Pino logs can be prettified by using a Pino prettifier
+module like [`pino-pretty`][pp]:
+
+```sh
+$ cat app.log | pino-pretty
```
davidmarkclements: are we going with `pino-pretty`, or `pino` since it's default prettifer?

davidmarkclements: add a line to the effect of "For almost all situations, this is the recommended way to log. The programmatic API is mostly for integration purposes of other CLI based prettifiers"

package.json
```diff
@@ -64,6 +61,7 @@
     "fresh-require": "^1.0.3",
     "log": "^1.4.0",
     "loglevel": "^1.6.1",
+    "pino-pretty": "^1.0.0",
```
mcollina: I think you can remove chalk from dependencies.

jsumners: Oops. Fixed.

jsumners:
`pino` bin added. Result:

```sh
% node example.js|pino
`pino` cli is deprecated. Use `pino-pretty` cli instead.
```

bin.js
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env node
+console.error('`pino` cli is deprecated. Use `pino-pretty` cli instead.')
```
marcbachmann: Doesn't **deprecated** mean that something is still working but it's discouraged to use?
I'd change the message in here to `'pino' cli got removed. Use 'pino-pretty' cli instead.`

A link to https://www.npmjs.com/package/pino-pretty could also be useful.

package.json
wtgtybhertgeghgtwtg: Ditto for `split2`.  But it still has to be a dev dependency, I think.

jsumners: Thanks. It's moved.

pino.js
```diff
@@ -272,6 +272,18 @@ Object.defineProperty(pinoPrototype, 'isLevelEnabled', {
   value: isLevelEnabled
 })
 
+function getPrettyStream (opts, prettifier) {
+  if (prettifier && typeof prettifier === 'function') {
+    return prettifier(opts).asMetaWrapper(process.stdout)
+  }
+  try {
+    var prettyFactory = require('pino-pretty')
+    return prettyFactory(opts).asMetaWrapper(process.stdout)
```
mcollina: Can you move `asMetaWrapper` in this repository? I don't think it would be needed at all in `pino-pretty` itself.

jsumners: I intend to when I get back to this later this week (*maybe* sometime today).

jsumners:
The latest commit moves `asMetaWrapper` into Pino as an internal API and updates the `doc/pretty.md` accordingly. How does this look @mcollina and @davidmarkclements ?

jsumners:
I think at this point @davidmarkclements issue has been addressed. If not, we can fix it before final v5.0.0 release. I'm merging this.

PR #391, opened 2018-04-13
Remove slowtime support

jsumners:
As per longstanding deprecation notice and https://github.com/pinojs/pino/projects/1, this PR removes the `slowtime` feature. This relegates time prettification to prettifiers. If people really want slow time, they can supply their own time formatting function via the `timestamp` option at logger instantiation (https://github.com/pinojs/pino/blob/a8d0b48408ca0262d398cd96309b9ea54cb1ec86/docs/API.md#parameters).

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -49,7 +49,7 @@
     Alternatively, it is possible to register a serializer under the key `Symbol.for('pino.*')` which will act upon the complete log object, i.e. every property.
   * `timestamp` (boolean|function): Enables or disables the inclusion of a timestamp in the
     log message. If a function is supplied, it must synchronously return a JSON string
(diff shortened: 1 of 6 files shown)


jsumners:
@mcollina we have been spitting out an error https://github.com/pinojs/pino/blob/a8d0b48408ca0262d398cd96309b9ea54cb1ec86/pino.js#L332

mcollina:
Thanks! I forgot! LGTM ;)

jsumners:
This should go in the release notes:

> + `slowtime` removed: if really want time formatted in-process, then you can supply an appropriate function via the `timestamp` option. However, we highly recommend formatting the time after the fact via a prettifier.

PR #392, opened 2018-04-13
Update fast-json-stringify

jsumners:
As per Project 1 (https://github.com/pinojs/pino/projects/1), this PR updates the `fast-json-stringify` dependency. Release notes should include:

> + Update `fast-json-stringify`: see https://github.com/davidmarkclements/fast-safe-stringify/blob/23fc3d1/CHANGELOG.md for notes on the breakages introduced by this.

@davidmarkclements: how about updating https://github.com/davidmarkclements/fast-safe-stringify/releases so that link isn't so ugly?

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
   },
   "dependencies": {
     "fast-json-parse": "^1.0.3",
-    "fast-safe-stringify": "^1.2.3",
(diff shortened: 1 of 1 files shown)


mcollina:
I think the best approach is to land this, get a prerelease out with a next tag, and then update noir and other packages.

PR #393, opened 2018-04-13
Use `sonic-boom` for extreme mode

jsumners:
This PR uses `sonic-boom` as the destination stream in extreme mode.

Currently, this is not working. I'm pushing up the broken code to get some guidance from @mcollina as to how it should really be implemented.


mcollina:
I'm currently seeing if I can find some time to work on it, I hope to get something going during this weekend.

davidmarkclements:
close in favour of #394

PR #404, opened 2018-04-28
Allow writing pretty text to any stream

jsumners:
This PR removes the prohibition on writing pretty text to any stream destination other than `process.stdout`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "fresh-require": "^1.0.3",
     "log": "^1.4.0",
     "loglevel": "^1.6.1",
-    "pino-pretty": "^1.0.0",
+    "pino-pretty": "^1.0.1",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.0.1",
     "snazzy": "^7.1.1",
--- a/pino.js
+++ b/pino.js
@@ -260,13 +260,13 @@ Object.defineProperty(pinoPrototype, 'isLevelEnabled', {
   value: isLevelEnabled
 })
 
-function getPrettyStream (opts, prettifier) {
+function getPrettyStream (opts, prettifier, dest) {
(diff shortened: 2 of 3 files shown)


mcollina:
Why? It seems a bad idea, even if it’s the most request feature.
Can you please add a test that verifies it works with a pino.destination?

jsumners:
@mcollina primarily so that I can write some tests in `pino-pretty`. But I also think with the revamped extreme mode it isn't necessary to impose the limitation.

This test never finishes because the `finish` event never gets fired:

```js
test('supports pino.destination stream', function (t) {
  t.plan(1)

  var tmpFile = path.join(os.tmpdir(), crypto.randomBytes(16).toString('hex'))
  var dest = pino.destination(tmpFile)

  dest.on('finish', () => {
    fs.readFile(tmpFile, (err, data) => {
      if (err) return t.threw(err)
      t.match(data.toString(), /^INFO.*foo\n$/)
      fs.unlink(tmpFile)
    })
  })

  dest.on('ready', () => {
    var log = pino({
      prettifier: require('pino-pretty'),
      prettyPrint: {
        levelFirst: true
      }
    }, dest)
    log.info('foo')
  })
})
```

jsumners:
I don't know of anywhere in the documentation that we discuss the limitation. I think it has always just been a runtime surprise (another reason to merge).

Do you have any ideas on getting the test in https://github.com/pinojs/pino/pull/404#issuecomment-385248354 to complete?

mcollina:
The problem is solved by https://github.com/mcollina/sonic-boom/pull/5.

mcollina:
Also, you'd need to call `destroy()` to have it emit finish.

jsumners:
Exactly. I think some documentation will be the easiest solution.

PR #406, opened 2018-04-30
Adding a line ending option

hibes:
Currently, `pino` appends a new line to every log line, and can be switched between `\n` and `\r\n`.

This pull request adds a separate boolean switch that would allow a user to opt out of line endings entirely.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -87,6 +87,7 @@
     * `hostname` (os.hostname)
     * `name` of logger if supplied as option
   * `crlf` (boolean): logs newline delimited JSON with `\r\n` instead of `\n`. Default: `false`.
+  * `appendLineEndings (boolean): controls whether or not newlines are appended to logs. Default: `true`.
 + `stream` (Writable): a writable stream where the logs will be written.
   It can also receive some log-line [metadata](#metadata), if the
   relative protocol is enabled. Default: `process.stdout`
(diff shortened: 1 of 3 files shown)


davidmarkclements:
I'm currently opposed to adding this, on the grounds that it's a footgun. Ndjson is far more efficient to parse

hibes:
The reason I'm looking to add this particular feature is that when pino is initialized with console as the write method, I am getting two output log lines.

```
require("pino")({}, { write: console.log }).info("Hello World");
```

Results in:

```
{"level":30,"time":1525120401082,"msg":"Hello World","pid":44135,"hostname":"x","v":1}

```

With whitespace as visible characters:
```
{"level":30,"time":1525120401082,"msg":"Hello World","pid":44135,"hostname":"x","v":1}¬
¬
```

jsumners:
This just sounds like misconfiguration. Supply `process.stdout` as the destination instead (who's is of course the default).

hibes:
I think I see how we got to this state of the code.  The last of the instructions on docs/API.md reads:

```
var instance = pino({}, {
  [Symbol.for('needsMetadata')]: true,
  write: function (chunk) {
    console.log('lastLevel', this.lastLevel)
    console.log('lastMsg', this.lastMsg)
    console.log('lastObj', this.lastObj)
    console.log('lastLogger', this.lastLogger)
    console.log('line', chunk)
  }
})
```

Our code is much closer to this code ^ including the `[Symbol.for('needsMetadata')]: true,` line.

We ended up working around the problem by trimming the chunk before passing it to console.log.
```
var instance = pino({}, {
  [Symbol.for('needsMetadata')]: true,
  write: function (chunk) {
    ...
    console.log('line', chunk.trim())
  }
})
```

It looks like we might be better off refactoring our code to solve this problem.

jsumners:
See the following for an example of proper usage of the metadata API:

https://github.com/pinojs/pino/blob/63b784acc9c6e1f10ecedd9c7e6abdeda00d0b6a/lib/tools.js#L30-L63

Issue #411, opened 2018-05-04
onTerminate without extreme mode


mcollina:
The `onTerminate` behavior is applied __also__ with `pino.destination(file)` or stdout.
We should apply just a limited version of this, just listening for `'exit'`.

jsumners:
I'm okay with a basic `exit` listener. Registering one of those does not alter the standard shutdown procedure like the others we listen for under extreme mode.

davidmarkclements:
Sgtm

jsumners:
Title says it should be for non-extreme mode. Didn't we only implement for extreme? I'm fine with it being closed if this is solved.

mcollina:
This is done ;).

PR #414, opened 2018-05-10
Only listen to all events if it is an extreme mode

mcollina:
Fixes https://github.com/pinojs/pino/issues/411

### DIFF
--- a/lib/events.js
+++ b/lib/events.js
@@ -27,10 +27,10 @@ module.exports = function (pinoInstance, internalExtremeHandler) {
     internalExtremeHandler()
   }
 
-  process.on('beforeExit', wrap('beforeExit'))
   process.on('exit', wrap('exit'))
 
   var handlers = {
+    beforeExit: wrap('beforeExit'),
     uncaughtException: wrap('uncaughtException'),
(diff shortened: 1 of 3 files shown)


lib/events.js
```diff
@@ -19,19 +27,26 @@ module.exports = function (pinoInstance, internalExtremeHandler) {
     internalExtremeHandler()
   }
 
+  process.on('beforeExit', wrap('beforeExit'))
```
jsumners: Do we need `beforeExit`? It's only called during normal shutdown not on an explicit shutdown. And the `exit` event is still going to fire after it.

test/events.test.js
```diff
@@ -18,7 +18,7 @@ test('no event loop logs successfully', function (t) {
   }))
 
   child.on('close', function () {
-    t.notEqual(output.match(/"msg":"h"/), null)
+    t.notEqual(null, output.match(/"msg":"h"/))
```
jsumners: If we are going to change all of these it might as well to be to using `t.match()`.

mcollina:
@jsumners @davidmarkclements PTAL. I've updated this according to feedbacks.

PR #415, opened 2018-05-12
Fast UTC time as timestamp 

davidmarkclements:
I've fixed up http://npm.im/fast-date so it doesn't use internals or deprecated API's, and it's still faster than `Date.now()`. 

Would we consider making UTC the default in pino v5? If not let's at least include it as an option

```
benchPinoEpochTime*10000: 929.780ms
benchPinoUnixTime*10000: 853.191ms
benchPinoUtcTime*10000: 817.595ms
benchPinoEpochTime*10000: 865.573ms
benchPinoUnixTime*10000: 861.808ms
benchPinoUtcTime*10000: 782.988ms
```

### DIFF
--- a/benchmarks/time.bench.js
+++ b/benchmarks/time.bench.js
@@ -1,0 +1,33 @@
+'use strict'
+
+var bench = require('fastbench')
+var pino = require('../')
+
+var epoch = pino({timestamp: pino.stdTimeFunctions.epochTime}, pino.destination('/dev/null'))
+var unix = pino({timestamp: pino.stdTimeFunctions.unixTime}, pino.destination('/dev/null'))
+var utc = pino({timestamp: pino.stdTimeFunctions.utcTime}, pino.destination('/dev/null'))
+
+var max = 100
+
+var run = bench([
+  function benchPinoEpochTime (cb) {
+    for (var i = 0; i < max; i++) {
+      epoch.info('hello world')
+    }
+    setImmediate(cb)
+  },
+  function benchPinoUnixTime (cb) {
+    for (var i = 0; i < max; i++) {
+      unix.info('hello world')
+    }
+    setImmediate(cb)
+  },
+  function benchPinoUtcTime (cb) {
+    for (var i = 0; i < max; i++) {
+      utc.info('hello world')
+    }
+    setImmediate(cb)
+  }
+], 10000)
+
+run(run)
--- a/benchmarks/time.bench.js
+++ b/benchmarks/time.bench.js
@@ -3,6 +3,7 @@
 var bench = require('fastbench')
 var pino = require('../')
 
+var iso = pino({timestamp: pino.stdTimeFunctions.isoTime}, pino.destination('/dev/null'))
 var epoch = pino({timestamp: pino.stdTimeFunctions.epochTime}, pino.destination('/dev/null'))
 var unix = pino({timestamp: pino.stdTimeFunctions.unixTime}, pino.destination('/dev/null'))
 var utc = pino({timestamp: pino.stdTimeFunctions.utcTime}, pino.destination('/dev/null'))
@@ -10,6 +11,12 @@ var utc = pino({timestamp: pino.stdTimeFunctions.utcTime}, pino.destination('/de
 var max = 100
 
 var run = bench([
+  function benchPinoIsoTime (cb) {
+    for (var i = 0; i < max; i++) {
+      iso.info('hello world')
+    }
+    setImmediate(cb)
+  },
   function benchPinoEpochTime (cb) {
     for (var i = 0; i < max; i++) {
       epoch.info('hello world')
--- a/lib/time.js
+++ b/lib/time.js
@@ -22,5 +22,6 @@ module.exports = {
   nullTime: nullTime,
   epochTime: epochTime,
   unixTime: unixTime,
-  utcTime: utcTime
+  utcTime: utcTime,
+  defaultTime: utcTime
 }
--- a/lib/time.js
+++ b/lib/time.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const fastDate = require('fast-date')
+
 function nullTime () {
   return ''
 }
@@ -12,8 +14,13 @@ function unixTime () {
   return ',"time":' + Math.round(Date.now() / 1000.0)
 }
 
+function utcTime () {
+  return ',"time": "' + fastDate() + '"'
+}
+
 module.exports = {
   nullTime: nullTime,
   epochTime: epochTime,
-  unixTime: unixTime
+  unixTime: unixTime,
+  utcTime: utcTime
 }
--- a/lib/time.js
+++ b/lib/time.js
@@ -10,12 +10,20 @@ function epochTime () {
   return ',"time":' + Date.now()
 }
 
-function unixTime () {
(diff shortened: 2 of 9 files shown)


mcollina:
Does changing it to UTC impacts perf (positively or negatively)?
Also, can we confirm the numbers on Node 8 and 10?

mcollina:
If we swap the default, we would have to bump the line format to v;1 (I think).

davidmarkclements:
Benchmarks show using fast date for utc positively impacts perf (eg it's faster) on 10. Need to run on 8

mcollina:
CI is red

davidmarkclements:
Yeah it's time stamp differences, time stamps are cached at 1s - will take
a while to fix, but we can still assess viability of it being default

davidmarkclements:
results on Node 8, fast-date is still 8% faster: 

```
benchPinoEpochTime*10000: 706.478ms
benchPinoUnixTime*10000: 672.067ms
benchPinoUtcTime*10000: 617.420ms
benchPinoEpochTime*10000: 658.277ms
benchPinoUnixTime*10000: 658.670ms
benchPinoUtcTime*10000: 598.660ms
```

davidmarkclements:
if the log is v1 with UTC timestamp, if someone opts for Epoch time stamp should the log then be v0?

jsumners:
I think that gets into a tricky problem. I believe it is best to simply say `v: 1` represents the delivered defaults. Once you alter the defaults, the version string doesn't really matter anymore and it's up to the user to know the difference. But if we change those defaults, then `v` must change and that can affect a great many people.

I think we should leave the default as it is and merely provide this as one of the alternatives.

davidmarkclements:
@mcollina CI is green

jsumners:
Technically, `unixTime` and `utcTime` should be reporting the exact same number:

> Unix time (also known as POSIX time[citation needed] or UNIX Epoch time[1]) is a system for describing a point in time, defined as the number of seconds that have elapsed since 00:00:00 Coordinated Universal Time (UTC), Thursday, 1 January 1970,[2].
https://en.wikipedia.org/wiki/Unix_time

If that is the case, I am fine with this being the default.

davidmarkclements:
@jsumners I'm half-inclined to agree with you - but I'll argue the counterpoint 

For pino, I think we should prioritize the most performant defaults whenever possible. In a minor/patch situation this wouldn't be possible, but because we're switching to a major, performance should take precedent over compatibility. 

We're also at a cross roads here, if we stay as is we are consistent with legacy pino, if we switch to UTC we are consistent with the logger ecosystem at large (bunyan, bole, winston, etc.)

On a side note, switching should also affect the benchmarks - we enable non-default "fast time" modes for other loggers, this should be switched off for a fair comparison if we offer UTC timestamps as default.

davidmarkclements:
@jsumners this is a good point, it might take some slight rearchitecting to do that but if consensus is that's important then I can make it work

jsumners:
@davidmarkclements is `fast-date` outputting seconds or milliseconds? If milliseconds, it should be equivalent to `Date.now` because `Date.now` is merely UTC in milliseconds.

I am not opposed to a `v: 2`, btw.

davidmarkclements:
@jsumners - seconds

jsumners:
That will cause issues with `pino-pretty` and the like. `new Date()` expects millisecond epochs.

davidmarkclements:
new Date can also take a utc string..

jsumners:
Can you test a version of fast-date that outputs milliseconds?

davidmarkclements:
Could do but unless an app is logging at nanosecond levels (unlikely if
there's other stuff going on) then there won't be any gains

jsumners:
The gain would be retaining compatibility with a speed increase (theoretically).

mcollina:
> The gain would be retaining compatibility with a speed increase (theoretically).

To be clear, we moved from ISO timestamps to epochs _for speed reasons_. I'm definitely +1 in moving back to __ISO__ timestamps.

However, this changes the default to unparseable `.toUTCString()` 

```js
d = new Date()
d.toUTCString() // 'Mon, 14 May 2018 11:49:53 GMT'
d.toISOString()  // '2018-05-14T11:49:53.808Z'
```


The latter includes _milliseconds_, in the example `808`.

davidmarkclements:
yes I've just come to a similar conclusion

davidmarkclements:
(didn't mean to close :/) 

You can't get a gain unless you're caching - and you can't cache on milliseconds because the cost of the setTimeout is usually a couple ms. 

We could offer ISO strings rounded to the nearest 10, 100 or 1000ms, but then it def shouldn't be the default

davidmarkclements:
utcTime could still be useful for HTTP servers who are interested in matching their logs to `Date` headers

jsumners:
1 node node 2018-05-14 08-00-30

davidmarkclements:
@jsumners is this reply to @mcollina regarding UTC time being unparsable? because I was just about to say, it is parsable

jsumners:
Correct.

davidmarkclements:
umm...  I just added `isoTime` to the benchmarks, and it's consistently faster than all other options in 8 and 10 

Node 8: 
```
benchPinoIsoTime*10000: 631.583ms
benchPinoEpochTime*10000: 764.734ms
benchPinoUnixTime*10000: 684.793ms
benchPinoUtcTime*10000: 637.815ms
benchPinoIsoTime*10000: 546.400ms
benchPinoEpochTime*10000: 697.360ms
benchPinoUnixTime*10000: 679.383ms
benchPinoUtcTime*10000: 625.108ms
```
Node 10: 
```
benchPinoIsoTime*10000: 761.630ms
benchPinoEpochTime*10000: 880.974ms
benchPinoUnixTime*10000: 850.540ms
benchPinoUtcTime*10000: 801.027ms
benchPinoIsoTime*10000: 766.795ms
benchPinoEpochTime*10000: 1288.589ms
benchPinoUnixTime*10000: 1297.384ms
benchPinoUtcTime*10000: 815.723ms
```

davidmarkclements:
if we offer ISO time rounded to the second we can get the same results as `utcTime`
we can also make unixTime faster

davidmarkclements:
Updated `fast-date` to support unix time (plus caching prefix/suffix)

Note that unix time is now the fastest (and more than 30% than before)

Node 8
```
benchPinoIsoTime*10000: 1402.495ms
benchPinoEpochTime*10000: 717.815ms
benchPinoUnixTime*10000: 619.282ms
benchPinoUtcTime*10000: 686.169ms
benchPinoIsoTime*10000: 1337.725ms
benchPinoEpochTime*10000: 723.801ms
benchPinoUnixTime*10000: 619.892ms
benchPinoUtcTime*10000: 645.761ms
```

Node 10 
```
benchPinoIsoTime*10000: 1565.109ms
benchPinoEpochTime*10000: 891.055ms
benchPinoUnixTime*10000: 825.385ms
benchPinoUtcTime*10000: 846.479ms
benchPinoIsoTime*10000: 1524.952ms
benchPinoEpochTime*10000: 886.523ms
benchPinoUnixTime*10000: 776.991ms
benchPinoUtcTime*10000: 822.611ms
```

mcollina:
Do we want to ship this in Pino@5? I'm somewhat -1 to change the default.

davidmarkclements:
I'm also -1 at this point

PR #417, opened 2018-05-22
Add badge to display install size

styfle:
This adds a badge to the `README.md` file to display the npm install size

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![banner](pino-banner.png)
 
-# pino&nbsp;&nbsp;[![Build Status](https://travis-ci.org/pinojs/pino.svg?branch=master)](https://travis-ci.org/pinojs/pino)&nbsp;[![Coverage Status](https://coveralls.io/repos/github/pinojs/pino/badge.svg?branch=master)](https://coveralls.io/github/pinojs/pino?branch=master) [![TypeScript definitions on DefinitelyTyped](http://definitelytyped.org/badges/standard.svg)](http://definitelytyped.org)
+# pino&nbsp;&nbsp;[![Build Status](https://travis-ci.org/pinojs/pino.svg?branch=master)](https://travis-ci.org/pinojs/pino)&nbsp;[![Coverage Status](https://coveralls.io/repos/github/pinojs/pino/badge.svg?branch=master)](https://coveralls.io/github/pinojs/pino?branch=master) [![Install Size](https://packagephobia.now.sh/badge?p=pino)](https://packagephobia.now.sh/result?p=pino) [![TypeScript definitions on DefinitelyTyped](http://definitelytyped.org/badges/standard.svg)](http://definitelytyped.org)
 
 [Extremely fast](#benchmarks) node.js logger, inspired by Bunyan.
 It also includes a shell utility to pretty-print its log files.


jsumners:
I really don't understand the need.

davidmarkclements:
this is only important for the browser, and I assume these measurements are for pino server (the index.js file) rather than pino browser (the browser.js) file

styfle:
I think this information is still relevant.

The `micro` team was able to make their package smaller after discovering the npm publish size was much larger than expected: https://github.com/zeit/micro/issues/234

The same with the `typescript` team here: https://github.com/Microsoft/TypeScript/issues/23339

styfle:
I can add another badge with the gzip size of [browser.js](https://cdn.jsdelivr.net/npm/pino@4.16.1/browser.js) if you would like.

jsumners:
What does this accomplish? Does it indicate the disk space necessary when installing with `--production`? Or does it show the space when installing all possible dependencies? Why is this information even necessary? Are we all moving back to IBM PS/2s?

jsumners:
I think it is disingenuous to advertise the install size of the package in development mode. Such information should really only be relevant to production deployments.

Also, if you really care about your disk space, use [pnpm](https://pnpm.js.org/).

styfle:
This is **not** using development mode....this reports the size of someone installing `pino` as a dependency in their own project.

For example:

```
mkdir myapp
cd myapp
npm init -y
npm install --save pino
du -sh .
```

jsumners:
Which is different from `npm install --save --production pino`.

styfle:
@jsumners With or without the `--production` flag reports 1.2 MB

Try it yourself:

```
mkdir myapp
cd myapp
npm init -y
npm install --save pino
du -sh . # 1.2 M
cd ..
mkdir myapp-prd
cd myapp-prd
npm init -y
npm install --save --production pino
du -sh . # 1.2 M
```

jsumners:
Regardless, I don't understand the need. I leave it up to @davidmarkclements and @mcollina to decide on this.

davidmarkclements:
@styfle I think we're closing for now, but good luck

on a side note, knowing the npm install for the browser version of a module is not nearly as helpful as knowing the size of the browserified/webpacked dep tree  (which you won't know from just npm install pino)

rictorres:
as i commented here: https://github.com/fastify/fastify/pull/1147

> i just want to point out the package size is a major performance bottleneck on the serverless land: lambda, azure functions, etc due to cold start times.

> and i think developers focused on serverless architectures would really benefit from having that information (size) upfront.

davidmarkclements:
interesting point, but instead of a correlative metric why not add an initialization time overhead metric?

mcollina:
Even if there is supporting evidence, there are a lot of super-effective techniques to fix this. As an example, webpacking your function in a single file/bundle. The above has way more effect than fighting to reduce the size of any module.

There is evidence that the more JS files are loaded, the more time it takes for node to start up. Solving it by demonizing modules is the wrong answer, as we are aware of the problem and working on it on Node.js itself.

The advantage of having tests, docs and examples in a module increases the development experience, and it’s something that is very valuable to me.

As such, the download size of a module is something we should not optimize for.

rictorres:
thanks a lot for your replies :)

i think we're missing the goal of the badge though. it's as informative as showing _100% code coverage_, _definitely typed_ - they are nice but don't really mean a lot.

it's not to say that we should only care about package size. and in my opinion, given the popularity of _serverless_, we should at least consider it before adding a bloated dependency.

also the start up delay is not really about the nodejs issue mentioned by @mcollina, serverless providers need to boot up your container on first load (cold start) and its size matters.

rictorres:
> interesting point, but instead of a correlative metric why not add an initialization time overhead metric?

@davidmarkclements that would be ideal however serverless providers are always changing/tweaking these so it would be probably inaccurate to measure. it also depends on many factors like amount of cpu and memory, noisy neighbors, etc.

PR #420, opened 2018-05-28
Add redact API as first class citizen + faster redaction than pino-noir

davidmarkclements:
This integrates pino with [fast-redact](http://github.com/davidmarkclements/fast-redact) which is a generic version of `pino-noir` but adds 16x less over head to serialization (28% overhead with `pino-noir` vs 2% overhead with `fast-redact`)

`fast-redact`  will only work with Node 6+ because of its use of `Proxy` to validate paths, this is why it's appropriate for pino v5.

The proposal is:

* Redaction is a first class citizen in Pino v5
* In Pino v4, `pino-noir` remains as the method of redaction
* This also enables (fast) intermediate wildcards (e.g. a.*.c) which `pino-noir` doesn't have

These factors together with the speed improvements should further incentivize upgrading to v5.

The PR adds `redact` and `censor` options (`censor` defaults to `[Redacted]`)

```js
require('pino')({
  redact: ['req.headers.cookie'],
  censor: 'hidden :D'
})
```

This API solves leaky abstraction in pino-noir (having to pass in the standard serializers), 
and side steps the issue of needing to tag a serializer that.. well.. actually serializes. After spending some time thinking about it, imnho it's the cleanest and fastest way to integrate (we don't have to worry about whether `safe` is false etc etc)

## Redact benchmarks

Overhead of redaction is very low, sometimes redaction is faster – I chose a run where it's not faster. At any rate, redaction is within noise.

```sh
node benchmarks/redact.bench
```

```
benchPinoNoRedact*10000: 349.224ms
benchPinoRedact*10000: 350.569ms
benchPinoUnsafeNoRedact*10000: 327.524ms
benchPinoUnsafeRedact*10000: 330.655ms
benchPinoExtremeNoRedact*10000: 205.040ms
benchPinoExtremeRedact*10000: 215.273ms
benchPinoUnsafeExtremeNoRedact*10000: 193.195ms
benchPinoUnsafeExtremeRedact*10000: 193.174ms
```

## Server overhead

When linking this branch into `pino-http` and then redacting a Cookie header the overhead is within noise, it's hard to verify exactly without using an isolated machine - but  the redacting server is frequently faster by a few %. This is probably because the non-redacting server because it's serializing and writing fewer bytes but also, this redaction may be triggering further V8 optimizations (see benchmarks). Even setting the `censor` to the same content as the Cookie header yields either more or similar req/s as having no redaction.

This is the redacting server:
```js
var http = require('http')
var server = http.createServer(handle)

var logger = require('../')({
  redact: ['req.headers.cookie']
})

function handle (req, res) {
  logger(req, res)
  res.end('hello world')
}

server.listen(3000)
```

This is the `autocannon` command (using higher duration and connections to try to reduce the noise to signal ratio):

```sh
autocannon -d60 -c50 -H Cookie="SESSID=298zf09hf012fh2; csrftoken=u32t4o3tb3gg43; _gat=1;" localhost:3000
```

Results for the redacting server are:
```
Stat         Avg     Stdev  Max
Latency (ms) 2.63    0.77   39.75
Req/Sec      15985.8 987.5  18209
Bytes/Sec    1.77 MB 111 kB 2.02 MB
```

If we take redaction out, results are:

```
Latency (ms) 2.75     0.73    28.55
Req/Sec      15728.87 744.53  18668
Bytes/Sec    1.75 MB  86.1 kB 2.07 MB
```

Over several runs results fluctuate, essentially adding redaction to a server is within noise.

## Comparison Benchmarks

The global cost of adding redaction is adding a `stringifiers` object and checking if `stringifiers[key]` exists before using `this.stringify`. 

`benchmarks/object.bench.js` for just the pino benchmarks

### This branch:

```
benchPinoObj*10000: 299.598ms
benchPinoUnsafeObj*10000: 265.580ms
benchPinoExtremeObj*10000: 143.180ms
benchPinoUnsafeExtremeObj*10000: 140.752ms
```

### next-major

```
benchPinoObj*10000: 301.574ms
benchPinoUnsafeObj*10000: 275.551ms
benchPinoExtremeObj*10000: 158.657ms
benchPinoUnsafeExtremeObj*10000: 145.294ms
```

The cost of this change is approximately zero.

## Tests and Docs

I'll add tests and docs once we're in agreement on API

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -23,8 +23,7 @@ var defaultOptions = {
   safe: true,
   name: undefined,
   serializers: {},
-  redact: [],
-  censor: '[Redacted]',
+  redact: null,
   timestamp: time.epochTime,
   level: 'info',
   levelVal: undefined,
@@ -300,12 +299,8 @@ function pino (opts, stream) {
 
   // internal options
   iopts.stringify = iopts.safe ? stringifySafe : JSON.stringify
-  iopts.stringifiers = (iopts.redact.length > 0) ? redact({
-    paths: iopts.redact,
-    censor: iopts.censor,
-    serialize: iopts.stringify
-  }) : {}
-  iopts.formatOpts = (iopts.redact.length > 0)
+  iopts.stringifiers = iopts.redact ? redact(iopts.redact, iopts.stringify) : {}
+  iopts.formatOpts = iopts.redact
     ? {stringify: iopts.stringifiers[redact.format]}
     : {stringify: iopts.stringify}
   iopts.messageKeyString = `,"${iopts.messageKey}":`
--- a/pino.js
+++ b/pino.js
@@ -9,6 +9,7 @@ var flatstr = require('flatstr')
 var SonicBoom = require('sonic-boom')
 var events = require('./lib/events')
 var levels = require('./lib/levels')
+var redact = require('./lib/redact')
 var tools = require('./lib/tools')
 var time = require('./lib/time')
 var needsMetadata = Symbol.for('needsMetadata')
@@ -22,6 +23,8 @@ var defaultOptions = {
   safe: true,
   name: undefined,
   serializers: {},
+  redact: [],
+  censor: '[Redacted]',
   timestamp: time.epochTime,
   level: 'info',
   levelVal: undefined,
@@ -151,7 +154,7 @@ function asJson (obj, msg, num, time) {
     for (var key in obj) {
       value = obj[key]
       if ((notHasOwnProperty || obj.hasOwnProperty(key)) && value !== undefined) {
-        value = this.stringify(this.serializers[key] ? this.serializers[key](value) : value)
+        value = (this.stringifiers[key] || this.stringify)(this.serializers[key] ? this.serializers[key](value) : value)
         if (value !== undefined) {
           data += ',"' + key + '":' + value
         }
@@ -179,7 +182,7 @@ function asChindings (that, bindings) {
     value = bindings[key]
     if (key !== 'level' && key !== 'serializers' && bindings.hasOwnProperty(key) && value !== undefined) {
       value = that.serializers[key] ? that.serializers[key](value) : value
-      data += ',"' + key + '":' + that.stringify(value)
+      data += ',"' + key + '":' + (that.stringifiers[key] || that.stringify)(value)
     }
   }
   return data
@@ -296,6 +299,11 @@ function pino (opts, stream) {
 
   // internal options
   iopts.stringify = iopts.safe ? stringifySafe : JSON.stringify
+  iopts.stringifiers = (iopts.redact.length > 0) ? redact({
+    paths: iopts.redact,
+    censor: iopts.censor,
+    serialize: iopts.stringify
+  }) : {}
   iopts.formatOpts = {lowres: true}
   iopts.messageKeyString = `,"${iopts.messageKey}":`
   iopts.end = ',"v":' + LOG_VERSION + '}' + (iopts.crlf ? '\r\n' : '\n')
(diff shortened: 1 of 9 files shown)


davidmarkclements:
@mcollina happy to add tests and docs - if we're all cool with the API (cc @jsumners )

mcollina:
I would prefer if we had:

```
require('pino')({
  redact: {
    paths: ['req.headers.cookie'],
    censor: 'hidden :D'
  } // this goes to fast-redact as is.
})
```

davidmarkclements:
I thought you might say that, we can go that way

It won't go to `fast-redact` as-is though, it has to be processed because of our object name space approach 

if you want the options passed directly you need

```
require('pino')({
  redact: {
    req: {
      paths: ['headers.cookie'],
      censor: 'hidden :D'
    }
  } 
})
```

Which is too verbose for the standard case imo, however it could be supported to allow different censors for different namespaces – this could be useful where on some namespaces you want to remove properties but for others you want to redact

mcollina:
I'm +1 with the current approach. the other one is too confusing.

davidmarkclements:
added tests, also updated `quick-format-unescaped` to take a stringifier and released v2 

this solves two problems: 

* keeping the fast-safe-stringify synced, and
* making the format function use the redactor to serialize

There is an implication for browser pino: it will no longer use fast safe stringifier via quick-format-unescaped for interpolated values – instead if will do a try/catch `JSON.stringify` (as per quick-format-unescaped default). We can put this back in, but I'm of the opinion that this is beneficial, it decreases payload size of browser side logging without losing much benefit. We can always put fast-safe-stringify back to `browser.js` instead of course.

test/redact.test.js
```diff
@@ -0,0 +1,119 @@
...
+//       headers: {
+//         host: 'localhost:3000',
+//         connection: 'keep-alive',
+//         cookie: 'SESSID=298zf09hf012fh2; csrftoken=u32t4o3tb3gg43; _gat=1;'
+//       },
+//       remoteAddress: '::ffff:127.0.0.1',
+//       remotePort: 58022
+//     }
+//   })
+// })
```
mcollina: why is this commented?

pino.js
```diff
@@ -22,6 +23,8 @@ var defaultOptions = {
   safe: true,
   name: undefined,
   serializers: {},
+  redact: [],
+  censor: '[Redacted]',
```
jsumners: How about `censorWith`?

jsumners: Or `redactWith`.

jsumners:
I rather like the array or object config option. It allows flexibility. I'm very onboard with this.

test/redact.test.js
davidmarkclements: accidental

pino.js
davidmarkclements: given the context it seems to make the same amount of sense a `censor` does, 
I think it's self explanatory - what else would `censor: <String>` mean for a logger?

jsumners: To me, `censor` implies a boolean that would enable or disable censoring.

mcollina:
In the documentation, can you add a caveat section like in https://github.com/davidmarkclements/fast-redact#caveat? We need to note that passing user input in the redact option _is not safe_.

davidmarkclements:
@mcollina totally agree 
I've updated the api so we have a basic case, `redact: Array` and an advanced case `redact: Object` – this means we drop the top level `censor` option in favour of `redact.censor` and allow for an additional `redact.remove` option

```js
pino({redact: ['a.b.c']})
```

and

```
pino({redact: {
  paths: ['a.b.c'],
  censor: '[Redacted]', //default
  remove: false // default
})
```

Final todo is the docs

davidmarkclements:
rebased

davidmarkclements:
@mcollina Docs have been added

This is no longer a WIP – once we have agreement on approval this can be merged.

🎉

davidmarkclements:
v5.0.0-rc.3 published

PR #421, opened 2018-05-28
Added Node 10 to .travis.yml

mcollina:


jsumners:
Do we want to target Node 10 with Pino v4?

mcollina:
Yes I think so. Node 10 is out, and our dependencies are already using pino@4 with it.

PR #423, opened 2018-05-29
feat: flush & flushSync

ronag:
When in extreme mode you want to be able to flush to log every second but without blocking the event thread. The only time you usually want to do a sync flush is on shutdown.

e.g.

```js
setInterval(() => logger.flush(), 1000)
function terminate(code) {
   logger.flushSync()
   process.terminate(code)
}
```

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -369,6 +369,13 @@ parameters will be reduced accodringly.
 Flushes the content of the buffer in [extreme mode](extreme.md). It has no effect if
 extreme mode is not enabled.
 
+<a id="flushSync"></a>
+## .flushSync()
+
+### Discussion:
+Synchronously flushes the content of the buffer in [extreme mode](extreme.md). It has no effect if
+extreme mode is not enabled.
+
 <a id="addLevel"></a>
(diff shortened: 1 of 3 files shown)


mcollina:
Can you please add tests for flushSync? Also docs?

docs/API.md
```diff
@@ -369,6 +369,13 @@ parameters will be reduced accodringly.
 Flushes the content of the buffer in [extreme mode](extreme.md). It has no effect if
 extreme mode is not enabled.
 
+<a id="flushSync"></a>
+## .flushSync()
+
+### Discussion:
```
davidmarkclements: I don't believe "Discussion:" is a common idiom in the documentation - can you please remove this for consistency

ronag: I'm confused... "Discussion" is all over the place. I just copy pasted the `flush` section and slightly modified it.

jsumners: @davidmarkclements it's the only thing I could think to title the subsection for each part of the API.

davidmarkclements:
supersceded yb #452

PR #426, opened 2018-05-30
Updates from master into next-major

mcollina:
As titled

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lib"
   ],
   "scripts": {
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -16,7 +16,7 @@
     "lib"
   ],
   "scripts": {
-    "browser-test": "zuul tape test/browser*test.js --local",
+    "browser-test": "airtap --local 8080 test/browser*test.js",
```
davidmarkclements: if we're switching out zuul, then I assume `zuul.yml` can be removed as well?

PR #427, opened 2018-06-01
30% basic benchmark improvement

mcollina:
This does two things:

1. use a faster way to encode messages.
2. do not flatstr if we are using SonicBoom, it will do it for us.

Before:

```
benchBunyan*10000: 693.108ms
benchWinston*10000: 574.107ms
benchBole*10000: 537.312ms
benchDebug*10000: 336.824ms
benchLogLevel*10000: 445.080ms
benchPino*10000: 310.882ms
benchPinoDestination*10000: 239.628ms
benchPinoExtreme*10000: 177.962ms
benchBunyan*10000: 632.296ms
benchWinston*10000: 500.196ms
benchBole*10000: 328.447ms
benchDebug*10000: 317.202ms
benchLogLevel*10000: 364.723ms
benchPino*10000: 299.117ms
benchPinoDestination*10000: 234.706ms
benchPinoExtreme*10000: 163.296ms
```

After:

```
benchBunyan*10000: 660.848ms
benchWinston*10000: 540.111ms
benchBole*10000: 531.759ms
benchDebug*10000: 335.192ms
benchLogLevel*10000: 385.685ms
benchPino*10000: 312.994ms
benchPinoDestination*10000: 202.055ms
benchPinoExtreme*10000: 132.033ms
benchBunyan*10000: 645.676ms
benchWinston*10000: 469.996ms
benchBole*10000: 300.862ms
benchDebug*10000: 324.070ms
benchLogLevel*10000: 357.076ms
benchPino*10000: 265.018ms
benchPinoDestination*10000: 185.741ms
benchPinoExtreme*10000: 119.442ms
```

These were taken from Node 10, but we can appreciate a similar improvement on Node 8.

cc @BridgeAR

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -125,6 +125,34 @@ Object.defineProperty(
   {value: LOG_VERSION}
 )
 
+// magically escape strings for json
+// relying on their charCodeAt
+// everything below 32 needs JSON.stringify()
+// 34 and 92 happens all the time, so we
+// have a fast case for them
+function asString (str) {
+  var result = ''
+  var last = 0
+  var l = str.length
+  var point = 255
+  if (l > 42) {
+    return JSON.stringify(str)
+  }
+  for (var i = 0; i < l && point >= 32; i++) {
+    point = str.charCodeAt(i)
+    if (point === 34 || point === 92) {
+      result += str.slice(last, i) + '\\'
+      last = i
+    }
+  }
+  if (last === 0) {
+    result = str
+  } else {
+    result += str.slice(last)
+  }
+  return point < 32 ? JSON.stringify(str) : '"' + result + '"'
+}
+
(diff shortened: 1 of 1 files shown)


pino.js
```diff
@@ -214,7 +241,11 @@ function pinoWrite (obj, msg, num) {
     stream.lastTime = t.slice(8)
     stream.lastLogger = this // for child loggers
   }
-  stream.write(flatstr(s))
+  if (stream instanceof SonicBoom) {
```
davidmarkclements: Faster way than instanceof? (Maybe if sonic boom adds flag to stream?)

mcollina: I cannot appreciate any difference here in not having an `instanceof` check in this block if it's a SonicBoom instance. This specific `instanceof` gets optimized away.

PR #428, opened 2018-06-02
v5: Docs merge howtos and faq

davidmarkclements:
FAQ and Howto's are very similar ("How to" vs "How do I") – I think they should be the same doc

Also this updates the "How do I use Pino with Koa?" section


davidmarkclements:
ah yes, thanks - doing it now

davidmarkclements:
@jsumners updated 👍

PR #438, opened 2018-06-15
Modernize test syntax [WIP – Do not merge]

davidmarkclements:
To match the code refactor, this updates the tests syntax.

The goal is to make tests very clean and clear. 

This introduces async/await syntax to tests - support for async/await tests on Node 6 is made possible with [`qodaa`](http://github.com/davidmarkclements/qodaa) 

There's still a lot to do, this is just to get early feedback - also I'm seeing if CI works


mcollina:
I’m kind of -1 with the approach because 1) transpiling and 2) messy streams support fir async/await.

What you really want is the same logic we have for the AsyncIterators in Node 10. It would get into readable-stream asap, but it won’t remove the need for transpiling.

Tests also serves as docs/example and the way the helper is hooked up (stream.next) is really not something I can agree with.

mcollina:
I’m way more concerned about that stream.next() than of the transpiling.

davidmarkclements:
@mcollina I'm going to replace the `stream.next` thing with a more normative `.once('data')` approach (promisified). I think that will solve the non-conventionality issue with `stream.next`. As for AsyncIterators.. we only need the first piece of data for 90% of cases.. I don't see them as being that relevant atm. 

For Node 6 I agree with @jsumners – I think we should either drop it, or minimally transpile the tests

davidmarkclements:
@mcollina 

How's this:

```js
test('can add a custom level via constructor', async ({is}) => {
  const stream = sink()
  const instance = pino({level: 'foo', levelVal: 35}, stream)
  is(typeof instance.foo, 'function')
  instance.foo('bar')
  const { msg } = await once(stream, 'data')
  is(msg, 'bar')
})
```

mcollina:
I just think all of this is well ahead of time. Wrapping streams with promises is something that is going to bite us one day or another.

davidmarkclements:
@mcollina I agree - so that's why I'm saying let's treat it as an event emitter. That's all we need in the tests, a single data event in almost all cases.

test/helper.js
```diff
@@ -1,28 +1,30 @@
...
+const os = require('os')
+const writer = require('flush-write-stream')
+const split = require('split2')
+const pid = process.pid
+const hostname = os.hostname()
+const v = 1
+
+function once (emitter, name) {
+  return new Promise((resolve) => emitter.once(name, resolve))
+}
```
mcollina: Can you please have it listen also to ‘error’ until ‘data’ is emitted?

davidmarkclements: 👍

lib/events.js
```diff
@@ -35,6 +35,8 @@ function events (instance) {
     }
 
     function theWorldIsBurning (err) {
+      if (theWorldIsBurning.called === true) return
```
jsumners: This was to prevent the method from running multiple times on shutdown (due to different shutdown events). Is this no longer needed? I'm not clear on this change.

package.json
```diff
@@ -17,8 +17,8 @@
   ],
   "scripts": {
     "browser-test": "airtap --local 8080 test/browser*test.js",
-    "test": "standard | snazzy && NODE_OPTIONS='--no-warnings' tap -j 4 --no-cov test/*test.js",
-    "ci": "standard | snazzy && NODE_OPTIONS='--no-warnings' tap -j 4 --cov test/*test.js",
+    "test": "standard | snazzy && NODE_OPTIONS='--no-warnings -r qodaa' tap -j 4 --no-cov test/*test.js",
```
jsumners: I'm not fond of this. My favorite thing about `tap` is that you don't have to use a runner to run the tests; you can simple `node the-test-file.js`. This change breaks that functionality without explaining it to potential new contributors.

I understand this should only be necessary on Node 6. I think we should just drop Node 6 support for pino@5 and prevent this hack.

davidmarkclements:
We can also just require qodaa at the top of each test file

davidmarkclements:
No wait, we can't. There would have to be an entry point which requires the
test. Hmm that will be too disruptive

package.json
davidmarkclements: This is a good point, but anyone who is actively developing with Node 6, for whatever reason - has already bought into the fact that newer code may not work or may take extra steps. I think that is going to effect few to no people

lib/events.js
davidmarkclements: I added this to prevent it running multiple times on shutdown? 
Maybe I took it out by accident during the refactor and the tests have highlighted it?

jsumners: I swear it looked like a removal change. Ignore this.

PR #444, opened 2018-07-08
Rename .pino to .version

davidmarkclements:
Due to it's self referential nature `.pino` is difficult to talk about, because you have to explicitly signify that you're talking about the version property not pino itself. This leads to confusing documentation.

This PR suggests changing the property to `version` while we have the chance prior to v5 release, and also adding a `version` property to the exported function.


davidmarkclements:
full docs will be coming as separate PR

PR #447, opened 2018-07-10
V5 docs

davidmarkclements:
Apologies if the following explanation of changes/goals is dense, my tired brain has started to give up:

* brings documentation inline with any api changes
* bitesize reading through subject isolation
* configure docsify with pino shades of green & other minor aesthetic adjustments
* add pino tree to side bar 
* restructure api.md for (hopefully) improved clarity
* use headers for API for easier linking to specific areas
* restructure sidebar according to content changes
* split out FAQ into Web Frameworks and Help
* make the readme as focused as possible, navigation consistent with sidebar
* reduce explanation footprints by removing slightly-off subject content, and/or replacing content with references using internal linking
* focus on content ordering that makes most sense for quick referencing *and* learning through linear reading
* consistent formatting
* consistent pseudo-code notation (e.g. `fn([param])` vs `fn(param?)` etc)
* consistent markdown syntax 

## Todo

- [x] "Migration & Legacy" doc
- [x] re-read and apply corrections
- [x] make sure all links are working
- [x] make sure all external links are tagged with  ⇗
- [x] make sure all heading levels are correct
- [x] probably go over the statics section again as it was done at the high tiredness levels
- [x] update benchmarks doc

image

image

image

image

Issue #446, opened 2018-07-10
v5 proposal: adding `levels` option, removing implicit `level`+`levelVar` behaviour


davidmarkclements:
I think we should allow a new `levels` option:

```js
pino({levels: {label: 'foo', value: 100}})
```

And remove the following implicit behaviour: 
```
  pino({level: 'foo', levelVal: 100})`
```

I also think `levelVal` should be a static property, instead of a getter/setter

The benefits:

* It's explicit
* Allows for multiple levels to be added
* It's much easier to document (less conflation of concepts)
* Will allow for less code,
* **could** lead to faster child logger creation (possible removal of additional `isStandardLevel` and `isStandardLevelVal` checks)

mcollina:
I'm not necessarily sure what you are asking. What would this do?

```js
pino({levels: {label: 'foo', value: 100}})
```

Does it create a custom level, or set the default level?

`pino({ level: 'info' })` set the default level.

davidmarkclements:
`pino({ level: 'info' })` sets the default level, but `pino({ level: 'foo', levelVal: 100 })`  *creates a new level*. This is documented behaviour in v4. 

I don't think this in intuitive and instead suggest we allow for a `levels` option that allows for additional custom levels. I've messed up the syntax, it should be `{foo: 100}`.
 Maybe we should call it `customLevels`:

```
const logger = pino({
customLevels: {
   foo: 100, 
   bar: 110
  }
})

logger.foo('hi')
logger.bar('hi')
logger.info('hi')
```

jsumners:
Go for it. I'm going to defend the current implementation as a backward compatible implementation. Yeah, that's the ticket.

mcollina:
Go for it, but use `customLevels`.

davidmarkclements:
done

PR #449, opened 2018-07-13
v5: custom levels

davidmarkclements:
* Adds `customLevels` config property as per #446 
* `customLevels` also supported in child bindings
* remove `levelVal` from config object (it's only purpose was for adding a customLevel)
* enforces levelVal property as read only (will throw on attempt to write)
* Docs to be added separately (as part of #447)

### API

```js
const logger = pino({
  customLevels: {
    foo: 35,
    bar: 45,
  }
})
logger.foo('hi')
const childLogger = logger.child({})
childLogger.foo('hi')
```

```js
const logger = pino()

const childLogger = logger.child({
  customLevels: {
    foo: 35,
    bar: 45,
  }
})

childLogger.foo('hi')
```

#### Outcome: Zero overhead.

```sh
$ node benchmarks/internal/custom-levels.js
```

```
benchPinoNoCustomLevel*10000: 691.204ms
benchPinoCustomLevel*10000: 693.940ms
benchChildNoCustomLevel*10000: 684.271ms
benchPinoChildCustomLevel*10000: 687.863ms
benchPinoChildInheritedCustomLevel*10000: 686.649ms
```

#### next-major

```
benchPinoChild*10000: 500.976ms
benchPinoChild*10000: 439.672ms
```

#### custom-levels

```
benchPinoChild*10000: 499.220ms
benchPinoChild*10000: 439.289ms
```

#### next-major

```
benchPinoCreation*10000: 231.800ms
benchPinoCreation*10000: 189.613ms
```

#### custom-levels

```
benchPinoCreation*10000: 220.555ms
benchPinoCreation*10000: 189.862ms
```


davidmarkclements:
#### Child Creation with Custom Levels
```
benchPinoChildCreation*10000: 839.401ms
benchPinoChildCreationCustomLevel*10000: 1107.336ms
```

Need to look into this, but it's likely because we're not reusing levelMethods for custom functions.

jsumners:
Will this still work?

https://github.com/pinojs/pino-arborsculpture/blob/03ca7415713bf1448c9d25b6911df74173c44e10/sculpt.js#L19

benchmarks/child-creation.bench.js
```diff
@@ -15,13 +14,13 @@ const blog = bunyan.createLogger({
   name: 'myapp',
   streams: [{
     level: 'trace',
-    stream: dest
+    stream: fs.createWriteStream('/dev/null')
```
jsumners: I think we have had questions about running tests and such on Windows. Will this work there?

davidmarkclements: oh I'm not sure - but this isn't anything new – I just updated to `pino.destination` because I missed it before

davidmarkclements:
Child creation has just gotten slower because I've come across an unintentional state sharing bug in `addLevel` – this bug will also be in v4. 

The `levels` property is inherited, `addLevel` checks for `hasOwnProperty` but on a prototype property - the `levels` property is unintentionally shared. 

In v4 the bug is like so:

```
const child = logger.child({})
child.addLevel('foo', 25) // true
const child2 = logger.child({})
child2.addLevel('foo', 35) // false
```

In commit https://github.com/pinojs/pino/pull/449/commits/164120042c4f256d3a2c75090ef6d4fcc0634cc6 the bug became apparent because it throws instead 

I'll see what I can do to optimize child creation again and then backport a fix to v4 after that

davidmarkclements:
@jsumners  yes definitely `level` getter/setter is still supported

rjharmon:
Is there any important reason the browser module shouldn't custom levels?

mcollina:
nobody implemented it, PRs are welcome.
Il 21 dic 2019, 04:29 +0100, Randall Harmon <notifications@github.com>, ha scritto:

PR #452, opened 2018-07-15
feat: flush & flushSync

davidmarkclements:
Due to massive amounts of changes this re-implements #423 - however I started with @ronag s
commit and built on there to ensure @ronag is also listed as a contributor for this

The docs for this will go in as part of the docs pr, #447

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -9,7 +9,7 @@ const time = require('./lib/time')
 const proto = require('./lib/proto')
 const symbols = require('./lib/symbols')
 const { mappings, genLsCache, assertNoLevelCollisions } = require('./lib/levels')
-const { createArgsNormalizer, asChindings } = require('./lib/tools')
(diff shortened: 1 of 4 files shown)


davidmarkclements:
strange - ci is red – tests are passing locally - need to look into this but out of time for tonight

pino.js
```diff
@@ -82,10 +82,13 @@ function pino (...args) {
   const time = (timestamp instanceof Function)
     ? timestamp : (timestamp ? epochTime : nullTime)
 
+  const extreme = stream.minLength === 4096
```
mcollina: I think this make sense if stream is a SonicBoom instance instead.

davidmarkclements: then there's no need for noop - which is a little less explicit  - what if both? 

```
const extreme = stream instanceof SonicBoom && stream.minLength === 4096
```

mcollina: right, we need to check both

davidmarkclements:
supsceded by #454

PR #454, opened 2018-07-18
pino.final

davidmarkclements:
* replaces `onTerminated` so we're not overriding global handlers
* is useful for `pino.destination` and `pino.extreme` so writes are *always* guaranteed on exit
* much easier to explain and document, non-leaky abstraction
* adds the flush method (correctly)
* precludes the need to a flushSync method


docs/api.md
```diff
@@ -647,6 +628,29 @@ module.
...
+
+Since final log writes cannot be gauranteed with normal Node.js streams, 
+if the `destination` parameter of the `logger` supplied to `pino.final`
+is a Node.js stream `pino.final` will throw. It's highly recommended
+for both performance and safety to use `pino.destination`/`pino.extreme`
+destinations.
+
+* See [`destination` parameter](#destination)
+* See [Exit logging help](/docs/help.md#exit-logging)
+* See [Extreme mode ⇗](/docs/extreme.md)
```
mcollina: Can you please add an example here, or link to the "log loss prevention" section?

docs/extreme.md
```diff
@@ -37,36 +43,52 @@ This has a couple of important caveats:
...
+
+The following strategy can be used to minimize log loss:
+
+```js
 const pino = require('pino')
 const dest = pino.extreme() // no arguments
 const logger = pino(dest)
 
+// flush every 10 seconds to keep the buffer empty 
+// in periods of low activity
```
mcollina: we should note that `flush` is async.

docs/help.md
```diff
@@ -8,6 +9,28 @@
...
+in cases of error. 
+
+Writing to a Node.js stream on exit is not necessarily guaranteed, and naively writing
+to an Extreme Mode logger on exit will definitely lead to lost logs. 
+
+To write logs in an exit handler, create the handler with `pino.final`:
+
+```js
+process.on('uncaughtException', pino.final(logger, (err, finalLogger) => {
+  finalLogger.error(err)
```
mcollina: can you add a `process.exit(1)` here? Also an example with `'unhandledRejection'` would be nice.

docs/help.md
```diff
@@ -8,6 +9,28 @@
...
+}))
+```
+
+The `finalLogger` is a special logger instance that will synchronously and reliably 
+flush every log line. This is important in exit handlers, since no more asynchronous
+activity may be scheduled.
```
mcollina: This should like to the final logger section.

test/final.test.js
```diff
@@ -0,0 +1,149 @@
...
+test('listener function immediately sync flushes when fired', async ({pass, fail}) => {
+  const dest = pino.destination('/dev/null')
+  var passed = false
+  dest.flushSync = () => {
+    passed = true
+    pass('flushSync called')
+  }
+  pino.final(pino(dest), () => {})()
+  await sleep(10)
+  if (passed === false) fail('flushSync not called')
```
mcollina: Can you verify that the destination is actually written? I trust SonicBoom tests up a certain point.

docs/extreme.md
```diff
@@ -37,36 +43,52 @@ This has a couple of important caveats:
...
+  finalLogger.info(`${evt} caught`)
+})
+// catch all the ways node might exit
+process.on('beforeExit', () => handler(null, 'beforeExit'))
+process.on('exit', () => handler(null, 'exit'))
+process.on('uncaughtException', (err) => handler(err, 'uncaughtException'))
+process.on('SIGHUP', () => handler(null, 'SIGHUP'))
+process.on('SIGINT', () => handler(null, 'SIGINT'))
+process.on('SIGQUIT', () => handler(null, 'SIGQUIT'))
+process.on('SIGTERM', () => handler(null, 'SIGTERM'))
```
mcollina: This example is not totally correct. On some of those event we would like to exit with an error code (or not), in some other we want to just flush.

Specifically with `SIGHUP`, we were just flushing asynchronously. I think that behavior should be documented as well.

davidmarkclements: I'll remove SIGHUP if it's a special case, we can add something to the help section about it

test/final.test.js
davidmarkclements: it's verified here: 

https://github.com/pinojs/pino/pull/454/files#diff-21810649b385f6e84ec1653845cc5221R6

davidmarkclements: the extreme mode test would fail without it

docs/help.md
davidmarkclements: there isn't a final logger section - but I linked the `pino.final` in the line above to the pino.final section in api

davidmarkclements: https://github.com/pinojs/pino/pull/454/files#diff-cd2af9a0c135fa45588493397b947c05R22

Issue #382, opened 2018-03-30
PrettyPrint fails when error's stack is null


grbr:
Hello!

I try to log an object like this:
```javascript
// just an object with property 'error'
var obj = { data: 'bla', error: { message: 'bla error', stack: null } }
prettyPino.info(obj)
```
and i get an error:
```
TypeError: Cannot read property 'length' of null
    at filter (node_modules/pino/pretty.js:87:23)
    at DestroyableTransform.mapLine [as mapper] (node_modules/pino/pretty.js:251:15)
    at DestroyableTransform.transform [as _transform] (node_modules/split2/index.js:33:21)
    at DestroyableTransform.Transform._read (node_modules/readable-stream/lib/_stream_transform.js:184:10)
    at DestroyableTransform.Transform._write (node_modules/readable-stream/lib/_stream_transform.js:172:83)
    at doWrite (node_modules/readable-stream/lib/_stream_writable.js:418:64)
    at writeOrBuffer (node_modules/readable-stream/lib/_stream_writable.js:407:5)
    at DestroyableTransform.Writable.write (node_modules/readable-stream/lib/_stream_writable.js:334:11)
    at EventEmitter.pinoWrite (node_modules/pino/pino.js:220:12)
    at EventEmitter.LOG (node_modules/pino/lib/tools.js:93:10)
...
```
In pretty.js null is a matches for 
```javascript
var matches = /^(\s*"stack":)\s*"(.*)",?$/.exec(line)
```
'length' of null is
```javascript
if (matches.length === 3) {
```

Please add null check
Best regards

jsumners:
Did this work correctly in 4.14.0?

davidmarkclements:
just checked - works correctly in 4.14.0 - error introduced in 4.15.0

jsumners:
Ugh, I need to get the CLI revamp done.

mcollina:
Anyway, this can be easily fixed with a small patch, @grbr would you like to do the honors?

jsumners:
FYI, I have this accounted for in the upcoming separate prettifier:

https://github.com/pinojs/pino-pretty/blob/95823e330698319b567b92754894dc564b8351db/index.js#L215
https://github.com/pinojs/pino-pretty/blob/95823e330698319b567b92754894dc564b8351db/test/errorProps.test.js#L46-L60

grbr:
i've added pull request

PR #457, opened 2018-07-20
fix: support logging errors that do not contain a stack

### LINKED ISSUE #456: Invalid error JSON logged if stack is undefined
To reproduce:

```js
const logger = require('pino')();
const err = new Error();
delete err.stack;
logger.error(err);

/*
{"level":50,"time":1532087320790,"msg":"","pid":20631,"hostname":"dev.local","type":"Error","stack":undefined,"v":1}
*/
```
Note the section `"stack":undefined` which is not valid JSON.

Real world example: Apollo GraphQL has a custom error class called ApolloError that does not hold `stack` property. When logging an ApolloError we see the issue above which breaks pino's pretty printer among other things.

Pino version: 4.17.3

disintegrator:
Fixes #456 

Custom errors may not have their stack property set.
Pino used to serialize undefined stack which created invalid JSON output

Given the following contrived example:
```js
const logger = require('pino')();
const err = new Error();
delete err.stack;
logger.error(err);
```
Would output:
```
{"level":50,"time":1532087320790,"msg":"","pid":20631,"hostname":"dev.local","type":"Error","stack":undefined,"v":1}
```
This is invalid JSON due to the `"stack":undefined` at the end of that output line.

This PR adds a fix to avoid logging the stack if it is not set. So the above code now outputs:

```
{"level":50,"time":1532094347667,"msg":"","pid":25624,"hostname":"dev.local","type":"Error","v":1}
```

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -146,7 +146,8 @@ function asJson (obj, msg, num, time) {
   if (hasObj) {
     var notHasOwnProperty = obj.hasOwnProperty === undefined
     if (objError) {
-      data += ',"type":"Error","stack":' + this.stringify(obj.stack)
+      data += ',"type":"Error"'
+      data += obj.stack ? ',"stack":' + this.stringify(obj.stack) : ''
     }
     // if global serializer is set, call it first
     if (this.serializers[Symbol.for('pino.*')]) {
--- a/pretty.js
+++ b/pretty.js
@@ -217,7 +217,9 @@ function pretty (opts) {
     line += eol
 
     if (value.type === 'Error') {
-      line += '    ' + withSpaces(value.stack, eol) + eol
+      line += value.stack
+        ? '    ' + withSpaces(value.stack, eol) + eol
+        : ''
 
       var propsForPrint
       if (errorProps && errorProps.length > 0) {
--- a/test/error.test.js
+++ b/test/error.test.js
@@ -128,3 +128,18 @@ test('an error with statusCode property is not confused for a http response', fu
   instance.level = name
   instance[name](err)
 })
+
+test('stack is omitted if it is not set on err', t => {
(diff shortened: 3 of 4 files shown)


jsumners:
Since `null` is valid JSON, wouldn't the more appropriate fix be to serialize it as `null` if `error.stack === undefined`?

mcollina:
I’d prefer to keep it as a string, so that users would always decode a
string. Errors should really have the stack property.
Il giorno sab 21 lug 2018 alle 14:46 James Sumners <notifications@github.com>
ha scritto:

jsumners:
See https://github.com/pinojs/pino/issues/382

davidmarkclements:
... I think we've made a mistake here.

For the current implementation, due to coercion, there are seven ways in which there may not be a stack

* delete
* null
* undefined value
* false
* empty string
* 0 
* NaN

Let's remove the silly ones and go to the ones where a user may be attempting something real:

* delete
* null
* undefined value
* false
* empty string

What's the (likely) expectations in each case?

* delete - I want to remove stack. In this case there should be no stack property
* null - I want stack to be null (as per #382). In this case it *should* be null. That's a valid JSON property.
* undefined value - I want stack to have the value `undefined`. This can't be done with JSON, in this case *only* we might consider making it an empty string. But we could also just allow it to be removed and educate anyone who doesn't know this part of the JSON spec. 
* false - I want to stack to be `false`. Then it should be false.
* empty string - I want stack to be an empty string. This will work anyway, but with extra unnecessary steps. 

I should have spent more time reviewing this. I'm not sure what to do with this in v4, but in v5 we need to be more thoughtful in how we approach this.

jsumners:
You understand the problem now :)

PR #462, opened 2018-07-25
destination reopening docs

davidmarkclements:
adds information on how to reopen log files if they are moved by a log rotater - e.g. if copytruncate isn't being used

cf https://github.com/pinojs/pino-tee/issues/12


docsify/sidebar.md
```diff
@@ -11,11 +11,12 @@
...
-  * [debug](/docs/help.md#debug)
+  * [Exit logging](#exit-logging)
+  * [Log rotation](#rotate)
+  * [Reopening log files](#reopening)
+  * [Saving to multiple files](#multiple)
+  * [Log filtering](#filter-logs)
+  * [Transports and systemd](#transport-systemd)
+  * [Duplicate keys](#dupe-keys)
+  * [Log levels as labels instead of numbers](#level-string)
+  * [Pino with `debug`](#debug)
```
mcollina: are you sure this is ok?

docs/help.md
```diff
@@ -63,6 +64,25 @@ We would rotate our log files with logrotate, by adding the following to `/etc/l
...
+reopens the log file destination.
+
+```js
+const dest = pino.destination('/log/file') // pino.extreme will also work
+const logger = require('pino')(dest)
+process.on('SIGUSR2', () => dest.reopen())
+```
+
+The log rotation tool can then be configured to send this signal to the process 
+after a log rotation has occurred.
```
mcollina: I think we should also add a config for logrotate to help with this.

docsify/sidebar.md
davidmarkclements: woops, fixing

PR #463, opened 2018-07-26
optimizing deep objects

davidmarkclements:
Our deep objects have become slow (in both v4 and v5, node 8 and 10). This PR is an on-going attempt to address that, but it's going to mean making hard  trade off decisions – which we'll need consensus on.

There's actually two optimizations here, one general and one that creates a fast path for objects that have no serializable properties.

The first wraps a try/catch around JSON.stringify, this gives around a ~15% improvement on deep objects. The trade-off of this approach is it opens up an edge case, if an object mutates itself in toJSON methods but has a circular reference that causes a throw it may have mutated parts of itself before throwing, and the following reserialization after a catch could give undesired results. This is why we had to not do this, since `pino-noir` uses that exact approach. However in v5 we're using fast-redact, which doesn't - and that approach is something we should recommend against, and is unlikely to be used outside of pino-noir. 

The second optimization checks for keys in the object that match serializers, if there are no serializers it will use JSON.stringify on the entire object, and then chop of the starting `{` and ending `}`. 
This gives a ~25% improvement over current deep object benchmarks. 

However... the trade off of this second optimization is that it **adds** ~25% onto our small object benchmarks. Which are still faster than everything even with that.

Let's discuss.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -1,5 +1,5 @@
 'use strict'
-
+const stringify = require('fast-safe-stringify')
 const format = require('quick-format-unescaped')
 const { mapHttpRequest, mapHttpResponse } = require('pino-std-serializers')
 const SonicBoom = require('sonic-boom')
@@ -16,7 +16,8 @@ const {
   stringifySym,
   needsMetadataGsym,
   wildcardGsym,
-  streamSym
+  streamSym,
+  redactSym
 } = require('./symbols')
 
 function noop () {}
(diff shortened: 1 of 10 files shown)


lib/tools.js
```diff
@@ -82,26 +83,41 @@ function asJson (obj, msg, num, time) {
...
+
+  // if global serializer is set, call it first
+  if (serializers[wildcardGsym]) {
+    obj = serializers[wildcardGsym](obj)
+  }
+
+  var hasSerializableKeys = false
+  for (var szKey in serializers) {
+    if (szKey in obj) hasSerializableKeys = true
+  }
```
mcollina: Instead of iterating over serializers, can you compile a function with all the set serializers names?

davidmarkclements:
The fast path *causes* the small object slow down. It's a trade off. We need to be +1 on both of neither

lib/tools.js
davidmarkclements: can try it

davidmarkclements: if we do that it will down the creation of slow child loggers with serializers  - they'll have to compile a funciton. that's going to be pretty expensive...

davidmarkclements: of course, doing that will negatively impact http logging

davidmarkclements:
I've gotten `asJson` to inline and now the hit on small objects is only 10% vs 25% faster on deep objects

lib/tools.js
mcollina: There is no need to do that for every child instantiation. The vast majority of the time child loggers do not have serializers, so we can reuse the function that was already there.

lib/tools.js
```diff
@@ -82,26 +89,38 @@ function asJson (obj, msg, num, time) {
...
+    data += ',"type":"Error","stack":' + stringify(obj.stack)
+  }
+
+  // if global serializer is set, call it first
+  if (serializers[wildcardGsym]) {
+    obj = serializers[wildcardGsym](obj)
+  }
+
+
+  if (serializersInObject(serializers, obj) === false) {
```
mcollina: this returns a boolean, there is no need for `=== false`.

davidmarkclements:
```
if (serializersInObject(serializers, obj) === false) {
```
is more explicit and easier to read than 
```
if (!serializersInObject(serializers, obj)) {
```

Explicit checking is also optimized for in turbofan..

jsumners:
I think that people adding `toJSON` methods should be smart enough to handle their own issues when their method is hit. We should only concern ourselves with the contract, i.e. objects can be serialized to JSON. So if the `try/catch` method has no impact on other types of objects ("small" ones), then I think that is the approach to take.

davidmarkclements:
yeah agree @jsumners  - the try/catch has no impact on small objects.

The no serializers fast path does impact our object benchmark... so to be specific, what we know for sure is the no serializers fast path is it impacts objects with a single key and a string value.

mcollina:
I’ve come to the conclusion that this PR just move the optimization of a benchmark vs another. I do not see any comparable difference when benchmarking pino-http for example. Our typical use case has a) some serializer set and b) some text message. We should really have a benchmark for that.

I’m fine in closing this for now.

davidmarkclements:
Closing in favour of #466

PR #464, opened 2018-07-27
Do not cast to string Error#stack.

mcollina:
Fixes https://github.com/pinojs/pino-pretty/issues/16.

@jsumners @davidmarkclements what do you think?

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -147,7 +147,9 @@ function asJson (obj, msg, num, time) {
     var notHasOwnProperty = obj.hasOwnProperty === undefined
     if (objError) {
       data += ',"type":"Error"'
-      data += obj.stack ? ',"stack":' + this.stringify(obj.stack) : ''
+      if (obj.stack !== undefined) {
+        data += ',"stack":' + this.stringify(obj.stack)
+      }
     }
     // if global serializer is set, call it first
     if (this.serializers[Symbol.for('pino.*')]) {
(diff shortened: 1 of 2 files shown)


davidmarkclements:
yes I agree - but this is now a breaking change, so should be in v5 but not v4

davidmarkclements:
We also need to discuss one case:

If I  `delete err.stack` then "stack" should not show.

However if I assign: `err.stack = undefined`, then some make the argument that they'd like to see, in some way, that it's been set to `undefined`. Which is how we got here.

So, should we set stack to an empty string when the `stack` is present but the value is `undefined`?

mcollina:
> yes I agree - but this is now a breaking change, so should be in v5 but not v4

I don't think so. We landed a botched commit that we shouldn't. Let's restore proper behavior.

> However if I assign: err.stack = undefined, then some make the argument that they'd like to see, in some way, that it's been set to undefined. Which is how we got here.

No, you do not want to see it. Mainly because:

```js
pino.info({ hello: undefined })
```

Will not show the `'hello'` property. `undefined` is invisible in JSON, and it is equal to `delete`.

mcollina:
I do not know how to run that, and there are no instructions. However it does fix the pino-pretty test, so I think we are good to go.

davidmarkclements:
To run just go to travis and restart build for master - prior to release. We don't have a way to run for pr's atm

PR #467, opened 2018-08-04
inline asJson

davidmarkclements:
```sh
 node --trace-turbo-inlining benchmarks/basic.bench.js | grep asJson
```
```
Inlining asString into asJson
  - size:473, name: asJson
  - size:473, name: asJson
  - size:473, name: asJson
  - size:473, name: asJson
Inlining asJson into write
  - size:473, name: asJson
  - size:473, name: asJson
Inlining asJson into LOG
  - size:473, name: asJson
  - size:473, name: asJson
  - size:473, name: asJson
  - size:473, name: asJson
Inlining asString into asJson
  - size:473, name: asJson
Inlining asJson into benchPinoNodeStream
  - size:473, name: asJson
Inlining asJson into LOG
```

* This seems to make no difference to benchmarks
* If the destination is a Node Stream `asJson` will not inline – no idea why

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -98,24 +98,19 @@ function asJson (obj, msg, num, time) {
         value = serializers[key] ? serializers[key](value) : value
 
         switch (typeof value) {
-          case 'undefined':
-            break
-          case 'string':
-            value = (stringifiers[key] || asString)(value)
-            if (value !== undefined) {
-              data += ',"' + key + '":' + value
-            }
-            break
+          case 'undefined': continue
           case 'number':
           case 'boolean':
             data += ',"' + key + '":' + value
+            continue
+          case 'string':
+            value = (stringifiers[key] || asString)(value)
             break
           default:
             value = (stringifiers[key] || stringify)(value)
-            if (value !== undefined) {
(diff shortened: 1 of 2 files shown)


lib/tools.js
```diff
@@ -98,24 +98,19 @@ function asJson (obj, msg, num, time) {
...
-            value = (stringifiers[key] || asString)(value)
-            if (value !== undefined) {
-              data += ',"' + key + '":' + value
-            }
-            break
+          case 'undefined': continue
           case 'number':
           case 'boolean':
             data += ',"' + key + '":' + value
+            continue
```
mcollina: I was wondering if we should call the stringifiers for booleans and numbers as well. What do you think?

davidmarkclements: yeah this is a good point, if there is a stringifier it may be a redactor 
and numbers definitely have a need to be redacted, and why not booleans as well,
particular in removal instead of redaction case

it might screw up inlining again though 🙈 I'll check

davidmarkclements: @mcollina - updated and still inlines, also tests added

PR #470, opened 2018-08-06
Removed safe option

### LINKED ISSUE #468: remove unsafe option in v5
After #466, I do not think the `unsafe`  option is needed anymore.

@davidmarkclements what do you think?

mcollina:
Fixes #468


mcollina:
@davidmarkclements done.

PR #472, opened 2018-08-06
Support automatic destination with string.

mcollina:
This should simplify the integration with pino-http, fastify, etc..

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -178,7 +178,7 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
 * See [Browser API ⇗](/docs/browser.md)
 
 <a id="destination"></a>
-### `destination` (SonicBoom | WritableStream)
+### `destination` (SonicBoom | WritableStream | String)
 
 Default: `pino.destination(1)` (STDOUT)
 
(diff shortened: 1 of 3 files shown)


mcollina:
I think that's _advanced_, and possibly users should avoid doing that. We are already picking up the fd from stdout and stderr anyway.

mcollina:
PTAL, I just added docs and fixed a bug.

lib/tools.js
```diff
@@ -193,6 +193,15 @@ function prettifierMetaWrapper (pretty, dest) {
 
 function createArgsNormalizer (defaultOptions) {
   return function normalizeArgs (opts = {}, stream) {
+    // support stream as a string
+    if (typeof opts === 'string') {
```
davidmarkclements: this could be one block with an ||

mcollina: It can't, or I would need to add one if inside, making it harder to read. I need to pass either opts or stream to `new SonicBoom`.

docs/api.md
```diff
@@ -178,7 +178,7 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
 * See [Browser API ⇗](/docs/browser.md)
 
 <a id="destination"></a>
-### `destination` (SonicBoom | WritableStream)
+### `destination` (SonicBoom | WritableStream | String)
 
 Default: `pino.destination(1)` (STDOUT)
```
jsumners: If `FileDescriptor` isn't going to be in the types supported, then this should read: `pino.destination(process.stdout)`.

mcollina: I think it's fine as it is right now, this is equivalent to:

```js
pino(pino.destination(1))
```

What `pino.destination(1)`  returns is an instance of Sonic Boom.

PR #473, opened 2018-08-08
Bump sonic-boom dependency

mcollina:


mcollina:
nothing, for some reason we didn't bump when we added `reopen()`.

PR #475, opened 2018-08-08
More fixes to pino-pretty, now it applies serializers as expected

mcollina:


mcollina:
No, it's matching a string `'object'`.

PR #486, opened 2018-08-14
Forward to pretty an error object passed as first arg

mcollina:
Fixes https://github.com/pinojs/pino-pretty/issues/23
Fixes https://github.com/pinojs/pino-pretty/issues/24


mcollina:
Landing as I don't have much time later today to land this, and I think this is fairly urgent to land.

PR #487, opened 2018-08-14
Symbol.for('pino.serializers') is part of the public API

mcollina:
Fixes https://github.com/pinojs/hapi-pino/issues/62

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -15,6 +15,7 @@
   * [logger.level](#level)
   * [logger.isLevelEnabled()](#islevelenabled)
   * [logger.levels](#levels)
+  * [Logger\[Symbol.for('pino.serializers')\]](#serializers)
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -543,6 +544,13 @@ $ node -p "require('pino')().levels"
 
 * See [`logger.level`](#level)
 
+<a id="serializers"></a>
+### logger\[Symbol.for('pino.serializers')\]
+
+Returns the serializers for the logger instance, as seen by the logger
+itself. Child loggers might reuse the same instance of their parent, if
+there are no changes.
```
davidmarkclements: Returns the serializers as applied to the current logger instance. If a child logger did not
register it's own serializer upon instantiation the serializers of the parent will be returned.

Issue #488, opened 2018-08-15
On demand log flushing?


thomasmichaelwallace:
Hi!

I'm trying to use pino (not in extreme mode, just the default destination) with AWS Lambda.

With Lambda AWS freezes the process after the (handler) function has completed and then _might_ thaw it later.

Periodically this causes a problem that pino doesn't write the log until after the process is unfrozen, or even loses logs if the process is killed after freezing.

I've seen some discussions- but I couldn't work out if there's a method for me to explicitly _make_ pino (`destination(1)`) flush that I could then call at the end of the function.

Apologies if this _is_ documented- I'm not that experienced with node streams.

mcollina:
What is very interesting is that it does not happen with node streams. They do very similar buffering, but maybe AWS lambda does something special that we are sidestepping.

pino.destination() has a flush method. It’s an instance of https://github.com/mcollina/sonic-boom.

By any chance, can you confirm that pino@4 worked ok?

Also, can you do a console.log(process.stdout) and paste what’s in there?

thomasmichaelwallace:
So here's the output from `console.log(process.stdout)` which I did just before returning (freezing) on a function that 'lost' two lines of logging:

```Socket {
connecting: false,
_hadError: false,
_handle: 
Pipe {
writeQueueSize: 0,
owner: [Circular],
onread: [Function: onread] },
_parent: null,
_host: null,
_readableState: 
ReadableState {
objectMode: false,
highWaterMark: 16384,
buffer: BufferList { head: null, tail: null, length: 0 },
length: 0,
pipes: null,
pipesCount: 0,
flowing: null,
ended: false,
endEmitted: false,
reading: false,
sync: true,
needReadable: false,
emittedReadable: false,
readableListening: false,
resumeScheduled: false,
destroyed: false,
defaultEncoding: 'utf8',
awaitDrain: 0,
readingMore: false,
decoder: null,
encoding: null },
readable: false,
domain: null,
_events: 
{ end: { [Function: bound onceWrapper] listener: [Function: onend] },
finish: [Function: onSocketFinish],
_socketEnd: [Function: onSocketEnd] },
_eventsCount: 3,
_maxListeners: undefined,
_writableState: 
WritableState {
objectMode: false,
highWaterMark: 16384,
finalCalled: false,
needDrain: false,
ending: false,
ended: false,
finished: false,
destroyed: false,
decodeStrings: false,
defaultEncoding: 'utf8',
length: 0,
writing: false,
[...]

PR #489, opened 2018-08-16
Do not use SonicBoom for stdout on Lambda

mcollina:
Fixes https://github.com/pinojs/pino/issues/488

@thomasmichaelwallace can you please verify that this PR fixes the problem?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -203,6 +203,8 @@ const stderrLogger = require('pino')(opts, pino.destination(2))
 const fileLogger = require('pino')('/log/path')
 ```
 
+On AWS Lambda the default destination is `process.stdout` instead.
+
 * See [`pino.destination`](#pino-destination) 
 * See [`pino.extreme`](#pino-extreme) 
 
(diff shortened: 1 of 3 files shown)


davidmarkclements:
I'm not sure about this, perhaps we should instead have a pino-lamda module - that calls pino with process.stdout , and uses process.nextTick to flushSync at the end of the function

mcollina:
You can’t pass in a serverless-specific config, because we overwrite the stream if it’s stdout. That’s part of what this PR fixes as well.

davidmarkclements:
ah yes.. what about an option: `useNodeSTDIO`

mcollina:
I would prefer if it worked by default on Lambda.
Il giorno gio 16 ago 2018 alle 16:31 David Mark Clements <
notifications@github.com> ha scritto:

jsumners:
I agree with @mcollina. Now that I do everything in AWS (Lambda, ECS, CloudWatch, all of that) there are things here that I think we need to be more lenient about to make the library useful in an increasingly popular environment / usage pattern (more on it later). The short of it is that someone using Pino in Lambda shouldn't have to think about it; it should just work.

Also, regarding the pipes, that's definitely an issue. It's definitely possible to freeze a process in a regular *nix environment -- http://web.mit.edu/gnu/doc/html/features_5.html

davidmarkclements:
ok, fair enough - let's go with it

davidmarkclements:
we still should think of a more graceful way around the flushSync situation

Issue #279, opened 2017-08-04
An option to have level be a string instead of a number


robinjoseph08:
Is it at all possible to have pino output the level string (`info`, `warn`, `error`, etc) instead of the number? The log viewer we're using can autodetect `"level":"error"` and highlight it red for us, but it doesn't do it for `"level":50`. Could there possibly be a option to toggle between integer and string?

Thanks!

mcollina:
No, write a transport if you need to convert these. Using numbers provide a very easy and straightforward way to do filtering. Which logviewer are you using?

robinjoseph08:
Is there a way we can hook up a transport that's within the same node process then? In our setup, we are aggregating logs through a docker log driver and then forwarding them to LogDNA (our log viewer). Because it's just pulling from docker's stdout logs, there's not a great way to pipe those through the transport.

mcollina:
Check out the [metadata protocol](https://github.com/pinojs/pino/blob/master/docs/API.md#metadata) and [pino-multi-stream](https://www.npmjs.com/package/pino-multi-stream).

You might be able to solve your issues. I would pass in a custom object, and in its write I would do: `line.replace(/"level":30/, "level":"info")`. With the metadata protocol you will know which numbered level to replace.

jsumners:
Closing because I think the issue has been addressed with appropriate suggestions. Please re-open if you have further questions.

thanpolas:
A natural evolution in logging is to normalize all logging formatting across your entire organization, regardless of platform or stack. To that end, an organization-wide Logging Schema is needed and from what research I did I came up with the [Log Event JSON Schema](https://github.com/timberio/log-event-json-schema/issues/37) which provides a solid schema to work on.

In that schema, log levels are strings, for good reasons as it makes filtering and processing log messages easier. What works for Node.js might not be the case for a Java application, so the schema needs to be, and actually is, platform agnostic.

A custom transport is a workaround, rather than a solution to this problem.

I'd ask the maintainers to reconsider their choices and allow Pino to be fully configurable to allow for string format in `level` propagation.

davidmarkclements:
I really like the idea of a universal cross-platform JSON logging schema – if such a schema was a standard spec I would absolutely start making changes to pino core with the emphasis being on the most performant implementation according to the spec. 

Since this schema is more of an independent effort, it reduces the strength of the argument for level strings in core vs the arguments against.

[...]

PR #492, opened 2018-08-18
Add useLevelLabels option

jsumners:
I am sure we have had more issues than just issue #279 regarding this, but I cannot find them easily. And I know I have argued our stance against this in the past. However, I have had to evolve my opinion:

1. The labels are just as parseable
1. We claim to follow [Bunyan](https://www.npmjs.com/package/bunyan): Bunyan outputs labels
1. We claim to be a Log4j style logger: Log4j defaults to outputting labels -- https://www.mkyong.com/logging/log4j-hello-world-example/

But the ultimate reason is selfish: I want to use Pino in my work at Knock.com and using transports is not feasible. Our platform is built upon AWS. Thus:

1. AWS Lambda doesn't support transports
1. AWS ECS could potentially support transports, but it would unnecessarily complicate deployments

Additionally, we are a polyglot environment. I need to be able to conform whatever logger I use to our internal log format specification. Using labels instead of integers makes this easier across multiple languages.

Initially I attempted to solve this locally through the global serializer, but that serializer doesn't receive the final log object. Maybe I could do this through the meta API, but both of these would just be hacks. I think we should be more flexible by:

1. Asserting a strong opinion about the usage of Pino via documentation and instance defaults
1. Recognizing that others have use cases outside of what we recommend and allowing them to make the choices they need to make

This feature, along with the `endSym` feature in Pino@5, allows me to accomplish that in my environment.

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -1,6 +1,6 @@
 'use strict'
 const flatstr = require('flatstr')
-const { lsCacheSym, levelValSym } = require('./symbols')
+const { lsCacheSym, levelValSym, useLevelLabelsSym } = require('./symbols')
 const { noop, genLog } = require('./tools')
 
 const levels = {
@@ -33,7 +33,9 @@ const initialLsCache = Object.keys(nums).reduce((o, k) => {
 
 function genLsCache (instance) {
   instance[lsCacheSym] = Object.keys(instance.levels.labels).reduce((o, k) => {
-    if (!(k in o)) o[k] = flatstr('{"level":' + Number(k))
+    o[k] = instance[useLevelLabelsSym]
+      ? `{"level":"${instance.levels.labels[k]}"`
+      : flatstr('{"level":' + Number(k))
     return o
   }, instance[lsCacheSym])
   return instance
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,4 +1,4 @@
-# API 
+# API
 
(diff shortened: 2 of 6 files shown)


mcollina:
Can you clarify to role of `endSym` here?

lib/levels.js
```diff
@@ -33,7 +33,9 @@ const initialLsCache = Object.keys(nums).reduce((o, k) => {
 
 function genLsCache (instance) {
   instance[lsCacheSym] = Object.keys(instance.levels.labels).reduce((o, k) => {
-    if (!(k in o)) o[k] = flatstr('{"level":' + Number(k))
+    o[k] = instance[useLevelLabelsSym]
+      ? `{"level":"${instance.levels.labels[k]}"`
+      : flatstr('{"level":' + Number(k))
```
mcollina: by any chance, can we avoid adding another symbol and passing another argument to `getLsCache`?

jsumners:
`endSym` doesn't have anything to do with this PR. I am merely stating that it is another feature that, in addition to this proposed one, helps me accomplish what I need to accomplish.

lib/levels.js
jsumners: If I remove the symbol then I will have to pass the `useLevelLabels` option around to various parts of the code so that child loggers will have access to it.

mcollina: Ok!

PR #493, opened 2018-08-19
Protect from hacks

### LINKED ISSUE #491: Doesn't work with pm2?
I can't see anything from pino when running through pm2 & pm2 logs?

mcollina:
Fixes #491


docs/api.md
```diff
@@ -203,6 +203,8 @@ const stderrLogger = require('pino')(opts, pino.destination(2))
 const fileLogger = require('pino')('/log/path')
 ```
 
+In case some other libraries monkey-patched `process.stdout.write`, that
```
jsumners: Let's replace these two sentences with:

> However, there are some special instances where `pino.destination` is not used as the default:
>
> + When something, e.g a process manager, has monkey-patched `process.stdout.write`.
> + When running on AWS Lambda.
>
>  In these cases `process.stdout` is used instead.

jsumners:
image

Issue #490, opened 2018-08-16
How to use pino.final with avvio


gyszalai:
I have an avvio application with a few onClose handlers that run when the application is
stopped when the node process receives a SIGTERM signal.

I know that `pino.final()` is made for this purpose, but how can I use it in an onClose handler?

jsumners:
Are you sure you don't want http://getpino.io/#/docs/api?id=flush ? Otherwise, I think the documentation is pretty clear http://getpino.io/#/docs/api?id=pino-final . What is unclear to you?

gyszalai:
the flush() method is async so it doesn't work in my case, i.e. the last log entries are not flushed to the destination before the VM process exits.

Yes, the documentation for pino.final() is clear, but this call is for creating a callback that you can pass as an event handler. In my case I just need the finalLogger in my onClose handlers (quite a few of them). Actually I did the following to get get the final logger:
```js
const finalLogger = pino.final(logger, (_, finalLogger) => finalLogger)()
```
which is not really nice IMHO

davidmarkclements:
We can adapt the api so that if a callback isn't supplied the final logger is returned - this would be a simple PR if you're up for contributing

gyszalai:
Okay, I'll file a PR soon. Already working on it. It'll break the API though, as now `pino.final` it throws an Error if no handler is passed to it.

davidmarkclements:
this isn't a breakage really, it's an api extension, since the PR has been merged and 5.3.0 has been released we're good to close

thanks @gyszalai  !

PR #496, opened 2018-08-21
Modify pino.final to handle calls without a handler

gyszalai:
Modifies pino.final to be able to return the finalLogger directly. See #490

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -662,7 +662,7 @@ of each function execution to avoid losing data.
 
 <a id="pino-final"></a>
 
-### `pino.final(logger, [handler]) => Function | Logger`
+### `pino.final(logger, [handler]) => Function | FinalLogger`
 
 The `pino.final` method can be used to acquire a final logger instance 
 or create an exit listener function.
--- a/docs/api.md
+++ b/docs/api.md
@@ -694,7 +694,7 @@ process.on('uncaughtException', pino.final(logger, (err, finalLogger) => {
 }))
 ```
 
-#### `pino.final(logger) => Logger`
+#### `pino.final(logger) => FinalLogger`
 
 In this case the `pino.final` method returns a finalLogger instance.
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -661,15 +661,12 @@ of each function execution to avoid losing data.
 * See [Reopening log files](/docs/help.md#reopening)
 
 <a id="pino-final"></a>
-### `pino.final(logger, handler) => Function`
 
-The `pino.final` method supplies an exit listener function that can be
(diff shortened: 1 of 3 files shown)


lib/tools.js
```diff
@@ -275,14 +275,31 @@ function final (logger, handler) {
   if (typeof logger === 'undefined' || typeof logger.child !== 'function') {
     throw Error('expected a pino logger instance')
   }
-  if (typeof handler !== 'function') {
+  var hasHandler = (typeof handler !== 'undefined')
```
jsumners: Why `var` instead of `const`?

docs/api.md
```diff
@@ -661,15 +661,12 @@ of each function execution to avoid losing data.
 * See [Reopening log files](/docs/help.md#reopening)
 
 <a id="pino-final"></a>
-### `pino.final(logger, handler) => Function`
 
-The `pino.final` method supplies an exit listener function that can be
-supplied to process exit events such as `exit`, `uncaughtException`,
-`SIGHUP` and so on.
+### `pino.final(logger, [handler]) => Function | Logger`
```
davidmarkclements: can we make it `FinalLogger`

lib/tools.js
```diff
@@ -275,14 +275,31 @@ function final (logger, handler) {
   if (typeof logger === 'undefined' || typeof logger.child !== 'function') {
     throw Error('expected a pino logger instance')
   }
-  if (typeof handler !== 'function') {
+  const hasHandler = (typeof handler !== 'undefined')
+  if (hasHandler && typeof handler !== 'function') {
```
davidmarkclements: can we update the error message: "if supplied, the handler parameter should be a function"

test/final.test.js
```diff
@@ -147,3 +138,22 @@ test('also instruments custom log methods', async ({pass, fail, error}) => {
   await sleep(10)
   if (passed === false) fail('flushSync not called')
 })
+
+function testLoggers (logger, finalLogger, is, isNot) {
```
davidmarkclements: I would much prefer that the assertions are repeated inline

for testing repetition > indirection - keep things declarative

docs/api.md
```diff
@@ -680,13 +677,32 @@ is a Node.js stream `pino.final` will throw. It's highly recommended
...
+by any additional arguments the `handler` may be called with.
+
 ```js
 process.on('uncaughtException', pino.final(logger, (err, finalLogger) => {
   finalLogger.error(err, 'uncaughtException')
   process.exit(1)
 }))
 ```
 
+#### `pino.final(logger) => Logger`
```
davidmarkclements: also, FinalLogger.

gyszalai: oops, sorry

Issue #494, opened 2018-08-19
Setting `base: null` removes logger name


jsumners:
http://getpino.io/#/docs/api?id=base-object

The docs claim that setting `base: null` will only remove the pid and the hostname from the output log but it is also removing the "name" field:

![image](https://user-images.githubusercontent.com/321201/44313226-5c316200-a3d2-11e8-9305-97b1c86090f2.png)

Is this a bug in the implementation or the documentation? I claim it to be a bug in the implementation.

mcollina:
Considering why I’ve added that, it’s a problem in the docs IMHO.

jsumners:
:(

jsumners:
Fair enough. I can solve locally via `endSym`.

PR #499, opened 2018-08-23
Base option readme fix

chrisandrews7:
Fix to API docs to correct the missing `name` property in base option.

Fixes https://github.com/pinojs/pino/issues/494


mcollina:
Thanks!

PR #503, opened 2018-08-24
Add ability to change level name when creating the instance

### LINKED ISSUE #495: Custom label for level
At first glance it looks like https://github.com/pinojs/pino/blob/master/lib/levels.js#L29-L42 is where the labels are set.

Some of my destinations use the label `severity` instead of `level`. Is there any way at the moment to configure this? Is a PR to add this feature welcome?

Arun4rangan:
Solution for issue: Custom label for level fix #495

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -179,6 +179,20 @@ Default: `false`
 Enables printing of level labels instead of level values in the printed logs.
 Warning: this option may not be supported by downstream transports.
 
+<a id="changeLevelName"></a>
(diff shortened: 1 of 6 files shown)


Arun4rangan:
Done! Let me know if you want me to change anything else!

Arun4rangan:
@jsumners updated the tests and updated the logic to use the `initialLsCache` if  `changeLevelName` is the default value.

PR #513, opened 2018-09-03
Do not override opts if the destination is a string

### LINKED ISSUE #512: Options are ignored when logging to a file
Some pino options are ignored when the destination is a file. Also, `useLevelLabels` seems to be able to change after a logger is already created.

Repro code:

```js
const pino = require('pino')

const opts = {
  base: null,
  timestamp: () => ',"time":"none"',
  useLevelLabels: true
}
const logger = pino(opts, './my.log')

logger.info('first')
pino(opts).info('second') // stdout
logger.info('third')
```

Contents of `my.log`:

```json
{"level":30,"time":1535948917084,"msg":"first","pid":34728,"hostname":"ubuntu","v":1}
{"level":"info","time":1535948917085,"msg":"third","pid":34728,"hostname":"ubuntu","v":1}
```

Terminal output:

```json
{"level":"info","time":"none","msg":"second","v":1}
```

Notice how the `base` and `timestamp` options have no effect on the logger that writes to `./my.log`.
Also notice how `useLevelLabels` is ignored at first, but then works after the second logger is created that writes to `stdout`.

Pino: v5.4.0
Node: v10.9.0
OS: Ubuntu 18.04, Windows 10

mcollina:
Fixes #512.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -242,12 +242,9 @@ function createArgsNormalizer (defaultOptions) {
     if (typeof opts === 'string') {
       stream = new SonicBoom(opts)
       opts = {}
-    }
-    if (typeof stream === 'string') {
+    } else if (typeof stream === 'string') {
       stream = new SonicBoom(stream)
-      opts = {}
-    }
-    if (opts instanceof SonicBoom || opts.writable || opts._writableState) {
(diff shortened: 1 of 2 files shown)


mcollina:
@nwoltman please confirm this works for you.

nwoltman:
Yes this fix worked. Thanks @mcollina!

Issue #507, opened 2018-08-29
Error: levels cannot be overridden


kilianc:
I was wondering what was the reasoning behind this. We're trying to configure the logger with the [stackdriver levels](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry).

`info` and `debug` are at the moment not configurable.

cc/ @Arun4rangan

mcollina:
I do not understand. Can you please make an example with code with what would you like to do, and what is currently throwing?

kilianc:
@Arun4rangan do you mind jumping in with the snippet?

kilianc:
@mcollina this currently throws:

```js
new Pino({
  default: 0,
  debug: 100,
  info: 200,
  notice: 300,
  warning: 400,
  error: 500,
  critical: 600,
  alert: 700,
  emergency: 800
})
```

mcollina:
What is it throwing, and what are you expecting that to do?

kilianc:
```
..../node_modules/pino/lib/levels.js:119
      throw Error('levels cannot be overridden')
      ^

Error: levels cannot be overridden
    at assertNoLevelCollisions (..../node_modules/pino/lib/levels.js:119:13)
    at new pino (..../node_modules/pino/pino.js:69:3)
    at Object.<anonymous> (..../lib/logger.js:6:18)
```

Expected behavior is to use my levels and discard default ones.

mcollina:
That is not possible. You might want to send a PR for that and add a new option.

mcollina:
Would you like to send a PR with a fix?

cberg-zalando:
@mcollina Actually, I already tried this morning, but it didn't work out. So I could open a WIP PR and you might have an idea why my the code I wrote is actually not doing what it should. If this approach is fine with you.

mcollina:
go for it

Il giorno ven 28 set 2018 alle 13:38 Christoph Berg <
notifications@github.com> ha scritto:

cberg-zalando:
@kilianc Do the changes done in #515 and #522 should suffice to consider this issue as done?

davidmarkclements:
closes by #515 and #522

PR #515, opened 2018-09-07
Allow custom levels to override default and allow to remove default levels

Arun4rangan:
Solves:

- Allows custom levels to override default levels provided by Pino 
- If needed default values from Pino can be removed 

Reference : #507

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -73,6 +73,27 @@ const logger = pino({
 logger.foo('hi')
 ```
 
+<a id=opt-useOnlyCustomLevels></a>
+#### `useOnlyCustomLevels` (Boolean)
+
+Default: `false`
+
+Use this option to only use defined `customLevels` and omit Pino's levels.
+Logger's default `level` must be changed to a value in `customLevels` in order to use `useOnlyCustomLevels`
+Warning: this option may not be supported by downstream transports.
+
+```js
+const logger = pino({
+  customLevels: {
+    foo: 35
+  },
+  useOnlyCustomLevels: true,
+  level: 'foo'
+})
+logger.foo('hi')
+logger.info('hello') // Will throw an error saying info in not found in logger object
+```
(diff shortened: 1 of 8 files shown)


mcollina:
Can you add this option to the docs?

Arun4rangan:
@mcollina sure!

Arun4rangan:
@mcollina Added to docs, let me know if you have any other requests!

mcollina:
CI is failing. It seems Object.values is not available.

Arun4rangan:
@mcollina I added additional tests as well, I am not sure why CI is failing. Its failing because there isn't enough tests for a function initialising line. Should I write more tests for it?

mcollina:
On Node 6: https://travis-ci.org/pinojs/pino/jobs/426109727. You are using some
syntax that is not available there.

mcollina:
You can use `npm run cov-ui` to generate the code coverage report. There is a condition in 1-2 branches that it is not covered.

Arun4rangan:
@mcollina I figured out which branches wasn't covered. All the tests are passing, let me know if you have any comments on the implementation!

docs/api.md
```diff
@@ -73,6 +73,26 @@ const logger = pino({
 logger.foo('hi')
 ```
 
+<a id=opt-useOnlyCustomLevels></a>
+#### `useOnlyCustomLevels` (Boolean)
```
jsumners: For all options that drastically change the output we should be adding a warning that using the option can break compatibility with downstream transports. See https://github.com/pinojs/pino/blob/7dd98cc8c2c41e1458e9eb6e0de07c18abb2ac9e/docs/api.md#uselevellabels-boolean for an example.

Sorry for the late review. Been very busy.

Arun4rangan:
@jsumners  added the warning!

Issue #516, opened 2018-09-17
Cannot set default level to a predefined value


husticdf:
Hi, a recent commit broke how I integrate your package into our project.

var pino = require("pino")({level: 30}) // previously worked, now breaks

https://github.com/pinojs/pino/commit/e870ff3811acdb5d0b2e0875892e5c3a527b7872#r30530279

mcollina:
Fixed in #517

PR #517, opened 2018-09-17
Fix for issue 516

husticdf:
The problem was described in https://github.com/pinojs/pino/issues/516


mcollina:
Can you add a unit test?

husticdf:
Added a test that fails without the code modification and succeeds after.

PR #520, opened 2018-09-24
Convert -Infinity to null

nwoltman:
Building on top of #519, this adds support for converting all invalid number values to `null` (really the only value left to handle was `-Infinity`, but this change does it all with one function call).


mcollina:
good catch and thanks!

PR #522, opened 2018-09-28
Don't use default levels if useOnlyCustomLevels is specified

cberg-zalando:
As said in #507, I gave it a try to allow to override the standard levels, but for some reason the priority in the test is `null`.

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -1,6 +1,6 @@
 'use strict'
 const flatstr = require('flatstr')
-const { lsCacheSym, levelValSym, useLevelLabelsSym, changeLevelNameSym } = require('./symbols')
+const { lsCacheSym, levelValSym, useLevelLabelsSym, changeLevelNameSym, useOnlyCustomLevelsSym } = require('./symbols')
 const { noop, genLog } = require('./tools')
 
 const levels = {
@@ -42,7 +42,11 @@ function genLsCache (instance) {
   return instance
 }
 
-function isStandardLevel (level) {
+function isStandardLevel (level, useOnlyCustomLevels) {
+  if (useOnlyCustomLevels) {
+    return false
+  }
+
   switch (level) {
     case 'fatal':
     case 'error':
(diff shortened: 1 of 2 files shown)


lib/levels.js
```diff
@@ -65,13 +69,14 @@ function setLevel (level) {
   if (values[level] === undefined) throw Error('unknown level ' + level)
   const preLevelVal = this[levelValSym]
   const levelVal = this[levelValSym] = values[level]
+  const useOnlyCustomLevelsVal = this[useOnlyCustomLevelsSym]
```
mcollina: have you verifyies this property is defined and of the correct value here?

cberg-zalando: I had put a `console.log` in the code and for my test its `true` and for the other tests where it is actually defined. I'm just confused that the test actually returns `null` for priority. That's really what I don't understand.

test/custom-levels.test.js
```diff
@@ -244,3 +244,20 @@ test('custom level does not affect changeLevelName', async ({ is }) => {
...
+      info: 35,
+      debug: 45
+    },
+    useOnlyCustomLevels: true
+  }, stream)
+
+  is(logger.hasOwnProperty('info'), true)
+
+  logger.info('test')
+  const { priority } = await once(stream, 'data')
```
mcollina: This is missing the `changeLevelName` option.

cberg-zalando: Ah… Then I misunderstood the `changeLevelName` option. And as I only want to test that my level setting is taken over, I will just use the standard name.

cberg-zalando:
What is still missing here? The property check you mentioned in your first comment?

cberg-zalando:
As I mentioned on #507, I was not sure why my unit test was failing, but as you pointed out, I just used the wrong property to compare with. So much for copy & pasting. So from my point of view, I'd say it does what it should: Allowing to redefine standard levels.

Issue #525, opened 2018-10-02
Invalid JSON logged when toString function returns undefined


NiXXeD:
Similar to https://github.com/pinojs/pino/issues/456

We found a situation where a some errors we were logging had the output contain `"toString":[object Undefined]`. This isn't valid JSON and was causing our log aggregator parser to break.

In playing with it, I am able to reproduce this by:
```javascript
const pino = require('pino')()
const err = new Error('Blah')
err.toString = () => undefined
pino.error(err, 'Something bad happened.')
```

The full output then is this, which is not valid JSON.
```json
{
  "level": "error",
  "time": 1538513586038,
  "msg": "Something bad happened.",
  "pid": 4086,
  "hostname": "dev.local",
  "type": "Error",
  "stack": "...",
  "toString": [object Undefined],
  "v": 1
}
```

I'm guessing this isn't exactly how it's produced, but this is *a* way to reproduce the same situation.

NiXXeD:
Yup this is fixed as of 5.6.3. Thank you!

PR #527, opened 2018-10-03
The serializer object should not have a prototype

mcollina:
https://github.com/pinojs/pino/issues/525

@NiXXeD can you please confirm that this fixes your issue?


NiXXeD:
Yes, validating it now.

PR #528, opened 2018-10-04
Makes final warn that is not supported instead of throwing

mcollina:
Fixes https://github.com/pinojs/pino-pretty/issues/37

### DIFF
--- a/test/final.test.js
+++ b/test/final.test.js
@@ -24,33 +24,18 @@ test('throws if the supplied handler is not a function', async ({ throws }) => {
   }, Error('if supplied, the handler parameter should be a function'))
 })
 
-test('logs as warn if not supplied logger with pino.destination or pino.extreme instance', async ({ plan, is }) => {
-  plan(2)
+test('throws if not supplied logger with pino.destination or pino.extreme instance', async ({ throws, doesNotThrow }) => {
+  throws(() => {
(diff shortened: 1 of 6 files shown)


davidmarkclements:
I've read up on the issue - not sure I agree that warning is better than throwing. 

Throwing is the right thing because you shouldn't use pino.final at all with streams 

I'd rather pino-pretty catches the error and then prints a warning itself

davidmarkclements:
yeah I think it caters to the two targeted cases - 

1. output a warning when using pretty print with pino.final
2. throw when using any other stream with `pino.final`

mcollina:
PTAL

test/final.test.js
```diff
@@ -27,7 +27,7 @@ test('throws if the supplied handler is not a function', async ({ throws }) => {
 test('throws if not supplied logger with pino.destination or pino.extreme instance', async ({ throws, doesNotThrow }) => {
   throws(() => {
     pino.final(pino(fs.createWriteStream('/dev/null')), () => {})
-  }, Error('only compatible with loggers with pino.destination and pino.extreme'))
+  }, Error('inal requires a stream that has a flushSync method, such as pino.destination and pino.extreme'))
```
davidmarkclements: typo: `inal` -> `final`

davidmarkclements:
other than the typo - LGTM

PR #529, opened 2018-10-14
Fix response headers in tests

JamesKyburz:
Fix the response tests since [this commit](https://github.com/pinojs/pino-std-serializers/commit/e2ea83bd2e3057dd1f5dbabafc520cca4110a233) was added to pino-std-serializers.

### DIFF
--- a/test/http.test.js
+++ b/test/http.test.js
@@ -136,7 +136,7 @@ test('http response support', async ({ ok, same, error, teardown }) => {
       v: 1,
       res: {
         statusCode: originalRes.statusCode,
-        header: originalRes._header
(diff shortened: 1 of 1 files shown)


jsumners:
I am undecided about the classification of this. I want to say patch, but it could certainly be argued for greater.

JamesKyburz:
@jsumners The tests referencing res._header both fail in the `.same` assertions.

This is because when pino-std-serializers@2.3.0 is installed headers are serialized, prior to the fix res.header was null.

This `PR` just reflects the changed made when serializing an http response, and adds test assertions for headers.

Issue #539, opened 2018-11-05
Error: EPIPE: broken pipe, write | pino 5.8.0


debianw:
NodeJS: v10.13
OS: MacOS 10.13.6
pino: 5.8.0

I'm doing a small transport test but after a benchmark I get this `broken pipe` exception.

Please take a look to this small demo where you can reproduce the issue: https://github.com/debianw/testing-pino

Just clone the repo and do

```
npm install
npm run start
autocannon -c 5 -p 5 -d 5 localhost:4001
```

You'll get this exception:

![screen shot 2018-11-05 at 1 11 50 pm](https://user-images.githubusercontent.com/1002407/48020906-c073e000-e0fc-11e8-9000-0343fd3ee4a6.png)

See the JSON it produce before the crash:
```
{
  "level": 30,
  "time": 1541445073817,
  "msg": "processing request",
  "pid": 9052,
  "hostname": "Walters-Mac-mini.{"level":30,"time":1541445073820,"msg":"processing request","pid":9052,"hostname":"Walters-Mac-mini.local","v":1}
```
That's not a valid JSON

jsumners:
![testing-pino zsh 2018-11-05 19-10-23](https://user-images.githubusercontent.com/321201/48034755-a4893200-e12e-11e8-9e86-1393dc94b5f1.png)

I am unable to reproduce it with your sample repo.

debianw:
@mcollina service in not crashing now in 5.8.1 but still the exception above before the crash is happening and logs stop working. #541

Issue #535, opened 2018-10-23
SonicBoom crashing the logs after updating pino to v5


andersonleite:
I'm facing an issue after updating pino from "pino": "^4.16.1" to "pino": "^5.8.0"

```
Error: EPIPE: broken pipe, write
Emitted 'error' event at:
    at SonicBoom.release (/app/node_modules/sonic-boom/index.js:62:12)
    at FSReqWrap.wrapper [as oncomplete] (fs.js:643:5)

```
Looks like pino > 5 started using SonicBoom and the logs are crashing on the server for me.
Is this a good place to search for help on this topic?

Thanks!

mcollina:
Can you please provide an example to reproduce?

andersonleite:
> is pino-tee staying alive?

No.

mcollina:
Does this happens on high throughput? How much does it happens?

It seems the destination is crashing because the pipe between the to processes is shut down.

debianw:
hi I'm having the same issue. @mcollina I prepare a repo with a small app where you can reproduce the issue. #ref 539

Please take a look to my repo here: https://github.com/debianw/testing-pino

PR #540, opened 2018-11-06
do not crash on broken pipe, just redirect to /dev/null

mcollina:
Fixes https://github.com/pinojs/pino/issues/539
Fixes https://github.com/pinojs/pino/issues/535

### DIFF
--- a/test/fixtures/broken-pipe/basic.js
+++ b/test/fixtures/broken-pipe/basic.js
@@ -1,0 +1,9 @@
+'use strict'
+
+global.process = { __proto__: process, pid: 123456 }
+Date.now = function () { return 1459875739796 }
+require('os').hostname = function () { return 'abcdefghijklmnopqr' }
+
+const pino = require('../../..')()
+
+pino.info('hello world')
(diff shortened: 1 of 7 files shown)


jsumners:
Question: I was just looking into the issue myself. When I modify the provided example service to use `process.stdout` (`pino({ destination: process.stdout })`), the error still occurs. This PR seems to imply the error only happens if the destination is a SonicBoom destination. Is this accurate?

test/fixtures/broken-pipe/basic.js
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+global.process = { __proto__: process, pid: 123456 }
+Date.now = function () { return 1459875739796 }
+require('os').hostname = function () { return 'abcdefghijklmnopqr' }
+
+const pino = require('../../..')()
```
jsumners: How is this test different than the `destination.js` test? Isn't the default value `pino.destination()`?

mcollina: This uses a different code path. We were creating SonicBoom instances on several places.

PR #550, opened 2018-11-22
Simplify switch statement

nwoltman:
Both the `'undefined'` and `'function'` cases can be handled in the same way.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -108,7 +108,9 @@ function asJson (obj, msg, num, time) {
         value = serializers[key] ? serializers[key](value) : value
 
         switch (typeof value) {
-          case 'undefined': continue
+          case 'undefined':
+          case 'function':
+            continue
           case 'number':
             /* eslint no-fallthrough: "off" */
             if (Number.isFinite(value) === false) {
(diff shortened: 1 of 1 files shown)


lib/tools.js
```diff
@@ -108,7 +108,9 @@ function asJson (obj, msg, num, time) {
         value = serializers[key] ? serializers[key](value) : value
 
         switch (typeof value) {
-          case 'undefined': continue
+          case 'undefined':
+          case 'function':
```
jsumners: I don't understand. Prior to this PR the "function" case re-assigned `value` to `undefined`. In this PR, that statement does not exist. How is this equivalent?

nwoltman: This is why the original code was confusing. Setting `value` to `undefined` just caused the `continue` statement to be triggered here:
https://github.com/pinojs/pino/blob/220e3d019fe22167a59cfe26260f6c2b4d0ea22b/lib/tools.js#L132

Here's the full context:
https://github.com/pinojs/pino/blob/220e3d019fe22167a59cfe26260f6c2b4d0ea22b/lib/tools.js#L110-L132

PR #551, opened 2018-11-22
Fix failing test - Allow censor function

nwoltman:
`fast-redact` allows the `censor` option to be a function starting in version [`1.4.0`](https://github.com/davidmarkclements/fast-redact/releases).


mcollina:
Thanks! code coverage for this is dropping slightly, specifically you have to remove https://github.com/pinojs/pino/blob/master/lib/redaction.js#L8.

PR #553, opened 2018-11-22
Added docs about the new censor function.

mcollina:
We missed to update the docs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -105,7 +105,7 @@ Each path must be a string using a syntax which corresponds to JavaScript dot an
 
 If an object is supplied, three options can be specified:
   * `paths` (array): Required. An array of paths. See [redaction - Path Syntax ⇗](/docs/redaction.md#paths) for specifics.
-  * `censor` (String|Function): Optional. A value to overwrite keys which are to be redacted.
+  * `censor` (String|Function|Undefined): Optional. When supplied as a String the `censor` option will overwrite keys which are to be redacted. When set to `undefined` the the key will be removed entirely from the object.
     In case a function is specified, it will be called with the value to be redacted and it must return the replacement
     value synchronously.
     Default: `'[Redacted]'`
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -105,7 +105,9 @@ Each path must be a string using a syntax which corresponds to JavaScript dot an
 
 If an object is supplied, three options can be specified:
   * `paths` (array): Required. An array of paths. See [redaction - Path Syntax ⇗](/docs/redaction.md#paths) for specifics.
-  * `censor` (String): Optional. A value to overwrite key which are to be redacted. Default: `'[Redacted]'`
+  * `censor` (String|Function): Optional. A value to overwrite key which are to be redacted.
```
jsumners: “key” => “keys”

mcollina:
PTAL

docs/api.md
```diff
@@ -105,7 +105,10 @@ Each path must be a string using a syntax which corresponds to JavaScript dot an
 
 If an object is supplied, three options can be specified:
   * `paths` (array): Required. An array of paths. See [redaction - Path Syntax ⇗](/docs/redaction.md#paths) for specifics.
-  * `censor` (String): Optional. A value to overwrite key which are to be redacted. Default: `'[Redacted]'`
+  * `censor` (String|Function): Optional. A value to overwrite keys which are to be redacted.
```
davidmarkclements:
```suggestion
  * `censor` (String|Function|Undefined): Optional. When supplied as a String the `censor` option will overwrite keys which are to be redacted. When set to `undefined` the the key will be removed entirely from the object.
```

docs/api.md
```diff
@@ -105,7 +105,10 @@ Each path must be a string using a syntax which corresponds to JavaScript dot an
...
+    In case a function is specified, it will be called with the value to be redacted and it must return the replacement
```
davidmarkclements:
```suggestion
    The `censor` option may also be a mapping function. The (synchronous) mapping function is called with the unredacted value. The value returned from the mapping function becomes the applied censor value. Default: `'[Redacted]'`
```

Issue #534, opened 2018-10-23
Partial Prettyfier


rubenhak:
Just like available in bunyan and winston, I would like to have ability to prettify the console output only, but keep the output to the file as json. Currently I'm using the following way:

```
const Pino = require('pino');
const multi = require('multi-write-stream');

var logFileStream = fs.createWriteStream('the-path');
var stream = multi([
    process.stdout,
    logFileStream
])

 var pinoOptions = {
    name: this._name,
    prettyPrint: {
        levelFirst: false,
        translateTime: true,
        colorize: true
    },
    prettifier: require('pino-pretty')
}

var logger = Pino(pinoOptions, stream);
```

P.S. I'm aware of the out-of-process prettyfier. In this particular case need to get console prettification in process.

rubenhak:
It was suggested to move the request here. [https://github.com/pinojs/pino-pretty/issues/31](https://github.com/pinojs/pino-pretty/issues/31)

mcollina:
This might require exposing some of the internals, but it should be doable.

It would need a combination of [getPrettyStream](https://github.com/pinojs/pino/blob/master/lib/tools.js#L168) with [pino-multi-stream](https://github.com/pinojs/pino-multi-stream). I think this would be a nice use case, and something that we should document.

Would you mind checking if the approach works, and in case we'll add `getPrettyStream` to our public API.

rubenhak:
Are you referring to something like this?

```
const Pino = require('pino');
const PinoPretty = require('pino-pretty');
const PinoGetPrettyStream = require('pino/lib/tools').getPrettyStream;
const multi = require('multi-write-stream');

var prettyStream = PinoGetPrettyStream({
        levelFirst: false,
        translateTime: true,
        colorize: true
    },
    PinoPretty,
    process.stdout);

var logFileStream = fs.createWriteStream('the-path');

var stream = multi([
    prettyStream,
    logFileStream
])

 var pinoOptions = {
    name: this._name
}

var logger = Pino(pinoOptions, stream);
```

mcollina:
Yes exactly, does it work?

Il giorno mar 23 ott 2018 alle 21:31 Ruben <notifications@github.com> ha
scritto:

rubenhak:
Something not working right. There is some magic inside `prettifierMetaWrapper` that expects `this`.

```
const fs = require('fs');
const Pino = require('pino');
const PinoPretty = require('pino-pretty');
[...]

Issue #530, opened 2018-10-18
How use v5 with prettyPrint and pino-multi-stream? 


sheldhur:
`
 const outStream = pino.pretty();
 outStream.pipe(process.stdout);
 streams.push({ level: options.level, stream: outStream });
 ...
 const logger = pino(options, streams && streams.length ? pinoms.multistream(streams) : process.stdout);
`
I want 3 streams for dev: pretty stdout, elasticsearch and rotating-file-stream.
It's will work in v4 becouse `pino` and `pino.pretty()` return stream. 

I'm trying get steram as
`
  const outStream = pino({prettyString: true});
  streams.push({ level: options.level, stream: outStream[pino.symbols.streamSym] });
`
but got an exception
`
(node:7822) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'match' of undefined
    at Object.write (%projectname%/node_modules/pino/lib/tools.js:225:16)
    at Object.write (%projectname%/node_modules/pino-multi-stream/multistream.js:59:16)
    at Pino.write (%projectname%/node_modules/pino/lib/proto.js:99:15)
    at Pino.LOG (%projectname%/node_modules/pino/lib/tools.js:39:21)
    at catch.e (%projectname%/src/index.js:11:10)
    at process._tickCallback (internal/process/next_tick.js:68:7)
    at Function.Module.runMain (internal/modules/cjs/loader.js:745:11)
    at Object.<anonymous> (%projectname%/node_modules/babel-cli/lib/_babel-node.js:154:22)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)
    at startup (internal/bootstrap/node.js:266:19)
    at bootstrapNodeJSCore (internal/bootstrap/node.js:596:3)
`

How I can do it now with v5?

jimmiehansson:
No prior experience from Pino, although it seems like a great project to use.
A quick search in the codebase seems that pretty adheres to `prettyPrint` and not `prettyString`, is this correct?
https://github.com/pinojs/pino-pretty#options

jsumners:
[...]

PR #555, opened 2018-11-30
Fix time in prettifier

sheldhur:
#530 
#534

### DIFF
--- a/test/pretty.test.js
+++ b/test/pretty.test.js
@@ -83,8 +83,15 @@ test('can send pretty print to custom stream', async ({ is }) => {
   log.info('foo')
 })
 
-test('multi-stream with pretty print', async ({ doesNotThrow }) => {
+test('multi-stream with pretty print', async ({ is, doesNotThrow }) => {
   doesNotThrow(() => {
+    const dest = new Writable({
+      objectMode: true,
+      write (formatted, enc) {
+        is(/^INFO.*foo\n$/.test(formatted), true)
+      }
+    })
+
     const { multistream } = require('pino-multi-stream')
     const outStream = pino({
       prettifier: require('pino-pretty'),
@@ -92,7 +99,7 @@ test('multi-stream with pretty print', async ({ doesNotThrow }) => {
         levelFirst: true,
         colorize: false
       }
-    })
+    }, dest)
     const log = pino({}, multistream([
       { stream: outStream[pino.symbols.streamSym] }
     ]))


lib/tools.js
```diff
@@ -220,7 +220,9 @@ function prettifierMetaWrapper (pretty, dest) {
 
       var time = this.lastTime
 
-      if (time.match(/^\d+/)) {
+      if (!time) {
```
jimmiehansson: It would be a good idea at this point to check that time actually has `match` in its string prototype for type restrictions before attempting to invoke it. This would reduce the nature of of "breaking" errors and malice code in the stream overflowing.

jimmiehansson:
> This is original code, not my :) Can you show situation example for "breaking"?

Sure! This type of scenario might be indigenous to other units of code, but it still presents possibilities to adhere error prone behavior when not properly checking the value being expressed.

Use case 1:
```js
      var time = 0x100 // Buffer cast to string hex disguise
      if (!time) { // Condition bypass, string eval to true
        time = JSON.parse(chunk).time // Ignored
      } else if (time.match(/^\d+/)) { // Attempts to find match prototype of string literal, fails
        time = parseInt(time) // Ignored
      }

// Exception thrown ...
```

Use case 2:
```js
      var time = 0 // Epoch fail, returns false, convert to string
      if (!time) { // Meets condition and assumes to parse
        time = JSON.parse(chunk).time // Steps in, assumes to default and parses from the chunk
      } else if (time.match(/^\d+/)) { // Condition met but never reached, double/unit, short circuits before
        time = parseInt(time) // Ignored
      }

// Time is now coheres to time from chunk, this is fine, if this is the semantics we want 
```

sheldhur:
Thx, but I ask example "breaking" executed via `pino`, not hypothetically possible.

sheldhur:
@mcollina pino-pretty with pino-multi-stream?

mcollina:
We need some tests that just use code in this repo. Also note that this is breaking our unit tests.

sheldhur:
I'm started tests before make changes and got exactly the same errors
```
test/pretty.test.js ................................. 28/34 10s
final works with pretty
```

jimmiehansson:
@sheldhur I don't necessarily agree, hypothetically possible is something completely different, what I pointed out is a state of fundamental awareness to the side effects that changes and the introduction of new code can produce. Evaluating the values to mitigate errors lead to stability, not probability. But I respect your opinion in that case. 
In either case its up to the maintainers to make that decision.

test/pretty.test.js
```diff
@@ -83,6 +83,24 @@ test('can send pretty print to custom stream', async ({ is }) => {
...
+      prettyPrint: {
+        levelFirst: true,
+        colorize: false
+      }
+    })
+    const log = pino({}, multistream([
+      { stream: outStream[pino.symbols.streamSym] }
+    ]))
+
+    log.info('foo')
```
jsumners: This is writing a log line to `stdout` when running the tests. Test output should be silent so that it does not interfere with the test runner's output.

lib/tools.js
```diff
@@ -220,7 +220,9 @@ function prettifierMetaWrapper (pretty, dest) {
 
       var time = this.lastTime
 
-      if (time.match(/^\d+/)) {
+      if (time == null) {
+        time = JSON.parse(chunk).time
```
mcollina: I think this is actually a bug in pino-multi-stream. It should set `lastTime`, See https://github.com/pinojs/pino-multi-stream/blob/28443984d06c43ffaf332165c1bd5395936bda25/multistream.js#L57.

sheldhur: @mcollina yeap, that is right.

sheldhur:
@mcollina open PR for pino-multi-stream
https://github.com/pinojs/pino-multi-stream/pull/19

mcollina:
I think this could be closed. Wdyt?

jsumners:
@sheldhur thank you for working on this and bringing bugs in tests to attention.

PR #556, opened 2018-11-30
ISSUE-555 Fix failing pretty tests for null return on match

jimmiehansson:
**Description**
Attempts to address problem with failing tests mentioned in #555 

See discussion in thread for ref.


jimmiehansson:
Closed see ref #557

PR #557, opened 2018-12-01
Fix pretty tests

jsumners:
This fixes the errors preventing #555 from passing validation. It's a result of https://github.com/pinojs/pino-pretty/pull/47 changing the formatting for log levels. I have also shifted to the use of the `execa` package in these tests because it actually returns the data in a predictable manner. Debugging this with `child_process.fork` was not working for me; the `actual` variable was an empty string when my debugger statement was hit.

### DIFF
--- a/test/pretty.test.js
+++ b/test/pretty.test.js
@@ -3,14 +3,14 @@
 const { Writable } = require('stream')
 const { test } = require('tap')
 const { join } = require('path')
-const { fork } = require('child_process')
+const execa = require('execa')
 const writer = require('flush-write-stream')
 const { once } = require('./helper')
 const pino = require('../')
 
 test('can be enabled via exported pino function', async ({ isNot }) => {
   var actual = ''
-  const child = fork(join(__dirname, 'fixtures', 'pretty', 'basic.js'), { silent: true })
+  const child = execa(process.argv[0], [join(__dirname, 'fixtures', 'pretty', 'basic.js')])
 
   child.stdout.pipe(writer((s, enc, cb) => {
     actual += s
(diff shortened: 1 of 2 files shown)


jsumners:
I leave it up to you @mcollina and @davidmarkclements which to merge: this one or #556. If #556, then it should be updated to account for the extra space after "WARN" and "INFO" instead of simply removing the check for those prefixes.

test/pretty.test.js
```diff
@@ -205,39 +205,39 @@ test('errors with props', async ({ isNot }) => {
...
+  const child = execa(process.argv[0], [join(__dirname, 'fixtures', 'pretty', 'final-no-log-before.js')])
 
   child.stdout.pipe(writer((s, enc, cb) => {
     actual += s
     cb()
   }))
   await once(child, 'close')
-  isNot(actual.match(/WARN: pino.final with prettyPrint does not support flushing/), null)
-  isNot(actual.match(/INFO \(123456 on abcdefghijklmnopqr\): beforeExit/), null)
+  isNot(actual.match(/WARN\s+: pino.final with prettyPrint does not support flushing/), null)
```
sheldhur: `WARN\s*` will be better, because space symbol is bug in pretty-print. It's can be fixed and test will be failed

jsumners:
I’m in the middle of packing for a house move over the next week. If we get the PR on `pino-pretty` then I should be able to make a quick update here at some point. But feel free to update this PR as needed if someone else has time before I do.

jimmiehansson:
I think @jsumners PR approaches better, its not degrading the test coverage of formatting or levels.
Maybe there could be a better way to extract the actual testing of expected levels vs. messages returned moreover. What is the purpose of the additional spacing? delim?

mcollina:
Having the mesaages aligned.

test/pretty.test.js
sheldhur: @mcollina maybe it was conceived? 
https://github.com/pinojs/pino-pretty/commit/a2a6821a5a556001d0ab04786296037ce3777f0c

jsumners:
@sheldhur see the PR for `pino-pretty` that I linked in the opening statement of this PR. That space was explicitly added to `pino-pretty` to format the messages. What seems to be happening is when `pino` is configured to _not_ show the pid and hostname then we get `WARN :` instead of `WARN (pid on hostname):`. If there is a bug, that is it.

sheldhur:
@jsumners yep, saw later

jsumners:
So do you still want to see the expression changed in this PR?

jimmiehansson:
All good

PR #561, opened 2018-12-05
Helper to match log output against keywords in tests

jimmiehansson:
Haven't had any time this week at all except a small amount this evening to come to any contributions whatsoever as I'm in middle of finishing up a job.

**Disclaimer**
I'd say this is as bare of a helper as they get, I really wish I could set aside more time for it, but PTL if its of use. :boom: 

**Description**
(Edge-case? the current tests are never implicit).
Attempts to make more sense in finding specific keywords/elements of choice in the assertions made for the log output, completely optional and non-obtrusive.

**Edge case anyway**
Assert if a certain literal word occurs in the combination of another due side-effects of the output

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -7,6 +7,43 @@ const pid = process.pid
 const hostname = os.hostname()
 const v = 1
 
+function expected (str) {
+  var memo = new WeakMap()
+
+  var valid = {
+    string: n => n !== void 0 && n.length > 0 && typeof n === 'string',
+    array: n => Array.isArray(n) && n.length > 0,
+    object: n => n !== null && typeof n === 'object'
+  }
+  var cursor = {
+    set: (k, v) => (valid.object(k)) ? memo.set(k, v) : void 0,
+    get: (k) => (valid.object(k)) ? memo.get(k) : void 0,
+    has: (k) => (valid.object(k)) ? memo.has(k) : void 0
+  }
+
+  var shake = ({ s_, ndl_ }) => {
+    var delim = ' '
+    /* eslint-disable */
+    var sane = /([\s()[\[^\]]:*)/mg
+    /* eslint-enable */
+
+    if (!valid.string(s_) || !valid.array(ndl_)) return void 0
+
+    var tree = s_.replace(sane, ' ').split(delim).filter(o => ndl_.indexOf(o) !== -1)
+    return tree.length !== 0
+  }
+
+  return {
+    output: function output$ ({ has }) {
+      if (!valid.string(str)) return void 0
+      var prev = { str: str, has: has }
+      var tree = shake({ s_: str, ndl_: has })
+      if (!cursor.has(prev)) cursor.set(prev, tree)
+      return cursor.get(prev)
+    }
+  }
+}
+
 function once (emitter, name) {
   return new Promise((resolve, reject) => {
     if (name !== 'error') emitter.once('error', reject)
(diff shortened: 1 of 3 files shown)


test/helper.js
```diff
@@ -7,6 +7,40 @@ const pid = process.pid
...
+  return {
+    output: function output$ ({ has }) {
+      if (!valid.string(str)) return void 0
+      var prev = { str: str, has: has }
+      var tree = shake({ s_: str, ndl_: has })
+      if (!cursor.has(prev)) cursor.set(prev, tree)
+      return cursor.get(prev)
+    }
+  }
+}
```
mcollina: Can you please add some docs about this?

jimmiehansson:
> Do you plan to use the utility in all tests?
No, I don't think its suitable for all tests, this is to extract relevant output from the assertions, as a proposal its not certain that it is needed be more than that to find a suitable path in decoupling the state of explicit matching of the output that may have unanticipated side effects further on, moreover I think it is better to see if this is something that can be of positive case for the current tests, and if not then it should be dismissed as so :+1:

test/helper.js
jimmiehansson: What conventions for inline documentation is in project?

jimmiehansson: Minor documentation added inline to clarify the purpose and usage

davidmarkclements:
I.. don't see the value here.. Why are we adding test helper util libraries code for one specific case? Isn't this over-abstraction and unnecessary indirection?

jsumners:
I'm with @davidmarkclements but for a different reason. I haven't had time to study this to figure it out. It really just confuses the tests for me.

jimmiehansson:
Fair enough :100:  
This PR is intended as a proposal helper (with little found time, unfortunately) with the reason to diverge some of the pain points that I have been seeing in the current test strategy where inconsistencies in the asserted code has been too literate and hardwired to result rather than functionality which is fine if acceptance comes from result of behavior.

There's a couple of areas that surface in the current tests that I found to be cumbersome to the integrity of the code. 

1. The current tests assert literal outcome and not expectation or behavior, this has a tendency to diverge the integrity of units of code actually needed to pass acceptance and make the tests more difficult for external contributors to understand. (E.g. Why the requirement of 2 spaces/tabs explicitly are needed in a lot of the cases which is not obvious to the developer) - Separation of this would make more sense from an integration point, for instance, to run 1 test that verifies this rather than 4 or more.
2. Edge-case or not, when if necessary the current format will have to be asserted against newer formats, the paradigm of the current test strategy requires a refactor, which is a direct anti pattern of having to change the test suite to meet the implementation specs rather than the other way around.
3. Abstraction and separation helps the entirety of the code base remain in a healthy and progressive state of development. I'm not claiming this barely 30 minutes of code would make any significant difference to the project as-is, but rather to introduce a few areas of which a change or enhancement could have significant result in the long run.

Apart from this I also agree, if value can't be seen at this point and this would be considered as an ambiguous entity to the current working-state, then I simply confirm why its submitted as a proposal

davidmarkclements:
@jimmiehansson thanks for this, however even if you're willing to refactor all tests to demonstrate your philosophy (which would be needed I think), it's still going to be a hard sell. I'm open to debate (as long as actual code is present) but I think at least two of the core contributors differ currently on philosophy. I'm going to close this PR for now but feel free to come back with more code to further exemplify the approach you are championing.

PR #566, opened 2018-12-09
Make pino.destination sync

### LINKED ISSUE #542: Pino 5.8.1 fails to flush log on process exit
If I use process exit after a log the log entry is missing -
```js
const pino = require('pino');
const logger = pino(pino.destination(1));

try {
	logger.info('start', pino.version);
	throw new Error('oops');
} catch (e) {
	logger.error(e);
	process.exit(1);
}
logger.info('end');
```
Expected Output -
```shell
{"level":30,"time":1541564755224,"msg":"start 5.8.1","pid":9237,"hostname":"dev","v":1}
{"level":50,"time":1541564755225,"msg":"oops","pid":9237,"hostname":"dev","type":"Error","stack":"Error: oops\n    at Object.<anonymous> (/test.js:7:8)\n    at Module._compile (internal/modules/cjs/loader.js:689:30)\n    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)\n    at Module.load (internal/modules/cjs/loader.js:599:32)\n    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)\n    at Function.Module._load (internal/modules/cjs/loader.js:530:3)\n    at Function.Module.runMain (internal/modules/cjs/loader.js:742:12)\n    at startup (internal/bootstrap/node.js:279:19)\n    at bootstrapNodeJSCore (internal/bootstrap/node.js:696:3)","v":1}
```

Actual output - 
```shell
{"level":30,"time":1541564543106,"msg":"start 5.8.1","pid":9214,"hostname":"dev","v":1}
```

I also tried without setting the destination but result is the same.
I think this might be a regression of https://github.com/pinojs/pino/issues/458 though I'm not sure if it worked on previous v5 versions

mcollina:
Fixes #542

### DIFF
--- a/test/destination-exit.test.js
+++ b/test/destination-exit.test.js
@@ -1,0 +1,23 @@
+'use strict'
+
+const { test } = require('tap')
+const { join } = require('path')
+const execa = require('execa')
+const writer = require('flush-write-stream')
+const { once } = require('./helper')
+
+// https://github.com/pinojs/pino/issues/542
+test('log everything when calling process.exit(0)', async ({ isNot }) => {
+  var actual = ''
+  const child = execa(process.argv[0], [join(__dirname, 'fixtures', 'destination-exit.js')])
+
+  child.stdout.pipe(writer((s, enc, cb) => {
+    actual += s
+    cb()
+  }))
+
+  await once(child, 'close')
+
+  isNot(actual.match(/hello/), null)
+  isNot(actual.match(/world/), null)
+})
--- a/test/final.test.js
+++ b/test/final.test.js
@@ -24,13 +24,13 @@ test('throws if the supplied handler is not a function', async ({ throws }) => {
   }, Error('if supplied, the handler parameter should be a function'))
 })
 
-test('throws if not supplied logger with pino.destination or pino.extreme instance', async ({ throws, doesNotThrow }) => {
+test('throws if not supplied logger with pino.extreme or pino.extreme instance', async ({ throws, doesNotThrow }) => {
   throws(() => {
     pino.final(pino(fs.createWriteStream('/dev/null')), () => {})
   }, Error('final requires a stream that has a flushSync method, such as pino.destination and pino.extreme'))
 
   doesNotThrow(() => {
-    pino.final(pino(pino.destination()), () => {})
+    pino.final(pino(pino.extreme()), () => {})
   })
 
   doesNotThrow(() => {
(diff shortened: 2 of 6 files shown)


mcollina:
@yonjah @brandondoran @pies @zyf0330 please test!

zyf0330:
With this fix, I don't must need `pino.final` or `pino.destination` to output complete log before `process.exit()`, just normal pino logger.

yonjah:
I did a few basic tests and it seem to be working with no issues

test/final.test.js
```diff
@@ -24,13 +24,13 @@ test('throws if the supplied handler is not a function', async ({ throws }) => {
   }, Error('if supplied, the handler parameter should be a function'))
 })
 
-test('throws if not supplied logger with pino.destination or pino.extreme instance', async ({ throws, doesNotThrow }) => {
+test('throws if not supplied logger with pino.extreme or pino.extreme instance', async ({ throws, doesNotThrow }) => {
```
davidmarkclements:
```suggestion
test('throws if not supplied logger with pino.extreme instance', async ({ throws, doesNotThrow }) => {
```

davidmarkclements: even though it's not necessary, can we add a smoke test to ensure that `pino.destination` it at least compatible with `pino.final`

test/destination-exit.test.js
```diff
@@ -0,0 +1,23 @@
+'use strict'
```
davidmarkclements: should there be a complementary `extreme-exit.test.js`

davidmarkclements:
1. I think this means we can remove (or at least deprecate) `destination.flushSync`
2. In the docs we can remove https://github.com/pinojs/pino/blame/master/docs/api.md#L665-L666
3. need to adjust pino.final docs (starting: https://github.com/pinojs/pino/blame/master/docs/api.md#L703) with an intro saying ONLY FOR `pino.extreme` MODE and then remove all references to `pino.destination`

mcollina:
> I think this means we can remove (or at least deprecate) destination.flushSync

destination and extreme are in fact the same thing, configured differently. We need it for extreme, so we get it in destination as well.

> In the docs we can remove https://github.com/pinojs/pino/blame/master/docs/api.md#L665-L666

Definitely.

> need to adjust pino.final docs (starting: https://github.com/pinojs/pino/blame/master/docs/api.md#L703) with an intro saying ONLY FOR pino.extreme MODE and then remove all references to pino.destination

I'll do that.

mcollina:
@davidmarkclements @jsumners PTAL, it should be ready to land.

I've removed the lambda hack, as it should not be needed anymore.

PR #567, opened 2018-12-17
Update api.md

turakvlad:
Hey,

Please take a look at these changes. The first change is just a cosmetic one. As for the default logging methods I think `error` should be there too.

Thanks.


davidmarkclements:
thank you!

PR #574, opened 2018-12-23
Remove unused dependencies

trygve-lie:
I noticed that `fast-json-parse` are listed as a dependency but not used in any of the code and that `pump` is listed under dependencies but only used in the benchmarks. This removes `fast-json-parse` as a dependency and moved `pump` to be a devDependency so its less deps to install for users of the lib.

This also eliminated a security warning when installing the dependencies in this module:

```sh
npm WARN notice [SECURITY] static-eval has the following vulnerability: 1 moderate. Go here for more details: https://nodesecurity.io/advisories?search=static-eval&version=0.1.1 - Run `npm i npm@latest -g` to upgrade your npm version, and then `npm audit` to get more info.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,15 +80,14 @@
     "tap": "^12.1.0",
     "tape": "^4.9.0",
     "through2": "^3.0.0",
-    "winston": "^3.1.0"
+    "winston": "^3.1.0",
+    "pump": "^3.0.0"
   },
   "dependencies": {
-    "fast-json-parse": "^1.0.3",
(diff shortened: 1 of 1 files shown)


trygve-lie:
I'm really not sure. 

Pre changing these dependencies I got that warning when installing dependencies on a fresh checkout of the repo. Then I changed to whats done in this PR, removed `node_modules` to do a fresh install for making sure tests run against new dependency tree and the warning was gone.

Issue #560, opened 2018-12-04
method to get a child loggers bindings


abuzarhamza:
Hi 

I am using pino logger . Sometimes we need to get the attr we have to create an audit record.
Here my scenario 

```
let pino = require('pino');

let p = pino({}).child({class: 'hello'});
p.info('hello');
console.log(p)
```

Is there any function to get the values of `class` we set for the child logger?

mcollina:
I think we should refactor https://github.com/pinojs/pino/blob/master/lib/tools.js#L198-L216 so that every logger has a new function `chindings()` to retrieve these. Would you like to send us a PR?

abuzarhamza:
@mcollina : one small question I do not find any documentation about `chindings` funciton and its usage. I can see it attached to pino object

davidmarkclements:
@abuzarhamza matteo is proposing a chindings function be added, whereas the `Symbol(pino.chindings)` property is private (as are all symbol properties) so does not require documentation.

If you do want to create a PR around this, please make sure benchmarks aren't negatively impacted (and post results)

yellowbrickc:
Hi @davidmarkclements @mcollina 

I have some free time and would like to spend it by helping out in open source projects. Before I create an unfinished PR I have some questions:
- Do I get it right, you propose a simple refactoring making the private chindings(). In tools.js there is already a method called asChindings https://github.com/pinojs/pino/blob/master/lib/tools.js#L137-L65 
I think this could be confusing.
- Should this property be available in pino.js as a property? 
- what about proto.js (there is a funny comment, I'm not sure it is still used)

If you have something better for me to do just tell :)

davidmarkclements:
@yellowbrickc cool!

* we’re talking about exposing a chindings method on every logger instance - the code that Matteo references is specific to the prettifier but could be repurposed and placed on all logger instances. asChindings isn’t exposed, and returns a string instead of an object. I could be renamed to getChindingsAsString but I think it’s fine as is. 
* it should be exposed on every logger instance - the pattern would be to place it in tools.js and export, include in proto.js to expose as a public method 
* it’s absolutely used.. are you referring to the comment about use of class syntax being satirical?

yellowbrickc:
Yes :)

davidmarkclements:
So that’s explained here https://github.com/pinojs/pino/pull/433#discussion_r194711807 (gh is opening the comment, open up the resolved discussion)

[...]

PR #588, opened 2019-01-20
add `bindings` method to logger instances

yellowbrickc:
Step 1:
Expose tools.getChindings()
Register a symbol for the method
Reference the method in prototype.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -11,7 +11,7 @@
   * [logger.error()](#error)
   * [logger.fatal()](#fatal)
   * [logger.child()](#child)
-  * [logger.childBindings()](#childBindings)
+  * [logger.bindings()](#bindings)
   * [logger.flush()](#flush)
   * [logger.level](#level)
   * [logger.isLevelEnabled()](#islevelenabled)
@@ -498,16 +498,16 @@ child.info({test: 'will be overwritten'})
 * See [`serializers` option](#opt-serializers)
 * See [pino.stdSerializers](#pino-stdSerializers)
 
-<a id="childBindings"></a>
-### `logger.childBindings()`
+<a id="bindings"></a>
+### `logger.bindings()`
 
 Returns an object containing all the current bindings, cloned from the ones passed in via `logger.child()`.
 ```js
 const child = logger.child({ foo: 'bar' })
-console.log(child.childBindings())
+console.log(child.bindings())
 // { foo: 'bar' }
 const anotherChild = child.child({ MIX: { IN: 'always' } })
-console.log(child.childBindings())
+console.log(child.bindings())
 // { foo: 'bar', MIX: { IN: 'always' } }
 ```
 
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -40,7 +40,7 @@ const constructor = class Pino {}
 const prototype = {
   constructor,
   child,
-  childBindings,
+  bindings,
   flush,
   isLevelEnabled,
   version,
@@ -86,13 +86,13 @@ function child (bindings) {
   return instance
 }
 
-function childBindings () {
+function bindings () {
   const chindings = this[chindingsSym]
   var chindingsJson = `{${chindings.substr(1)}}` // at least contains ,"pid":7068,"hostname":"myMac"
-  var bindings = JSON.parse(chindingsJson)
-  delete bindings['pid']
-  delete bindings['hostname']
-  return bindings
+  var bindingsFromJson = JSON.parse(chindingsJson)
+  delete bindingsFromJson['pid']
+  delete bindingsFromJson['hostname']
+  return bindingsFromJson
 }
 
 function write (obj, msg, num) {
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -24,14 +24,14 @@ test('child instance exposes pino version', async ({ is }) => {
   is(child.version, version)
 })
 
-test('childBindings are exposed on every instance', async ({ same }) => {
+test('bindings are exposed on every instance', async ({ same }) => {
   const instance = pino()
-  same(instance.childBindings(), {})
+  same(instance.bindings(), {})
 })
 
-test('childBindings contain the name and the child bindings', async ({ same }) => {
+test('bindings contain the name and the child bindings', async ({ same }) => {
   const instance = pino({ name: 'basicTest', level: 'info' }).child({ foo: 'bar' }).child({ a: 2 })
-  same(instance.childBindings(), { name: 'basicTest', foo: 'bar', a: 2 })
+  same(instance.bindings(), { name: 'basicTest', foo: 'bar', a: 2 })
 })
 
 function levelTest (name, level) {


mcollina:
+1 to that proposal @yellowbrickc! I would add that we should cache that value as it's not going to change after usage.

I like it a lot for two reasons:

1. it ensures that the data is fully cloned and serialized before accessing it back.
2. it does not touch a hot path in this library, which is still painful to opimize.

yellowbrickc:
Ok, than I will do this.
Regarding the caching, it could be dangerous. With this solution
```javascript
var bindings = null
function childBindings () {
  if (!bindings) {
    const chindings = this[chindingsSym]
    var chindingsObject = (chindings.indexOf(',') === 0) ? `{${chindings.substr(1)}}` : `{${chindings}}`
    bindings = JSON.parse(chindingsObject) || {}
    // eventually strip props as hostname and pid
  }
  return bindings
}
```
will the second of these tests fail, the method still returns the object w/o {foo: 'bar'}
```javascript
test('WIP childBindings are exposed on instance', async ({ isA }) => {
  const instance = pino()
  isA(instance.childBindings(), 'object')
})

test('WIP childBindings are exposed on every instance', async ({ match }) => {
  const instance = pino().child({ foo: 'bar' })
  match(instance.childBindings(), { foo: 'bar' })
})
```

mcollina:
Let's leave caching to a later time :).

davidmarkclements:
Quick check though: would a user expect the bindings to be cloned, or would a user expect to have the same object that was passed to child (referntially)?

mcollina:
100% cloned.

davidmarkclements:
Given an input API: `logger.child(bindings)`
And and output API `logger.childBindings() => bindings`

I kind of would expect the same object back, however we definitely should clone it because we need to uphold/promote the assumption that once set, the bindings are immutable. So we'll need to make docs clear on this.

@yellowbrickc if it makes sense to clone directly from the passed in `binding` object, (because that may make caching easier), use http://npm.im/rfdc for cloning

yellowbrickc:
Sorry, I didn't see your last comments.
I just finished what we discussed (and discovered the LsCache, I suppose, this is what you meant earlier), but I didn't touch this subject.
I have reread the #560 and I tried to fulfill his issues/wishes. I didn't returned the whole object (neither the level, nor any serializers) just the "current state of the binding props for child loggers". 

I tried but I cannot imagine any situation where I would need something else from a childlogger. But it is your decision :)

lib/proto.js
```diff
@@ -85,6 +86,19 @@ function child (bindings) {
   return instance
 }
 
+function childBindings () {
+  const chindings = this[chindingsSym]
+  var chindingsObject = (chindings.indexOf(',') === 0) ? `{${chindings.substr(1)}}` : `{${chindings}}`
```
davidmarkclements: nitpick: it's not an object yet, it's a string, maybe call it, .. `chindingsStr` or.. `chindingsJSON`

lib/proto.js
```diff
@@ -85,6 +86,19 @@ function child (bindings) {
...
+  try {
+    var bindings = JSON.parse(chindingsObject) || {}
+    delete bindings['pid']
```
davidmarkclements: avoid square bracket notation unless require - e.g. use `bindings.pid`  (and same below)

lib/proto.js
```diff
@@ -85,6 +86,19 @@ function child (bindings) {
...
+    delete bindings['hostname']
```
davidmarkclements: I wonder can there be slice out of the string before it gets parsed, this will mean that JSON.parse has less to process, and will return an object that has been deoptimized out of fast properties (https://v8.dev/blog/fast-properties) – `delete` is bad for objects

davidmarkclements:
I think this also has to be intergrated with pretty wrapper as well right? we need to remove that code from the pretty wrapper and use this instead?

lib/proto.js
yellowbrickc: yes, I also thought about this, but I have no idea how the string looks like. My current examples (via tests) are 
```js
// ,"pid":1180,"hostname":"myYoga"
// or
// ,"pid":1180,"hostname":"myYoga","name":"basicTest","foo":"bar","a":2
```
If I could be sure that pid and hostname are always near each other, in this order, than I could just strip them away with a regex. But that could be dangerous and if I make sure that nothing else will be removed the code will be of a slow performance.

I have read about delete being bad an the alternative is to set the properties to undefined. What do you think about that?

yellowbrickc:
The private method in the prettyWrapper does not remove any props, adds "v:1" to the output and is based on the lastLogger which is created/overwritten during the life cycle of the logger instance. It is used only for writing the logs.
The new method is a simple nice-to-have "reader" method, without side effects but with a try-catch (which I wouldn't us if errors wouldn't be allowed like in the other method) .
Also if you ask me than no, I wouldn't replace the other method.

docs/api.md
```diff
@@ -497,6 +498,19 @@ child.info({test: 'will be overwritten'})
 * See [`serializers` option](#opt-serializers)
 * See [pino.stdSerializers](#pino-stdSerializers)
 
+<a id="childBindings"></a>
+### `logger.childBindings()`
+
+Returns an object containing all the current bindings.
```
mcollina:
```suggestion
Returns an object containing all the current bindings, cloned from the ones passed in via `logger.child()`.
```

lib/proto.js
mcollina: I think `delete`  is fine here, let's leave optimizations to the next steps. This should not sit in a hot path, and in case we'll have a cache and a fast properties conversion in place.

yellowbrickc:
Finally I passed your 100% code coverage requirement 😆 
I cannot find any open conversations, I think @davidmarkclements you must accept them maybe?

mcollina:
Thanks for keeping up with us @yellowbrickc! Good work!

davidmarkclements:
sorry for the delay, I'm just thinking about the name, 

if the logger is a child, then the returned object isn't child bindings, it's just bindings

I think we should rename `childBindings` to `bindings`

yellowbrickc:
Done.

davidmarkclements:
amazing thanks @yellowbrickc

davidmarkclements:
v5.11.0 released 👍

yellowbrickc:
It was a pleasure.
It would be very helpful for me, if we could have short "retro" ;-)  I want to gather more experience in working on OSS.

1. At the end this was a very small PR (less then 20 lines) but we still had a lot of reworking and a lot of communication. My question is: this is the usual way or you would like more or less planing before?

2. Is this anything else what you would have to run somehow different?

jsumners:
I suspect some of the back and forth was due to a small language barrier. Regardless if that is true or not, it’s not the ultimate delta in code size that dictates the conversation around its implementation. The goal of the conversation is to make sure the right code gets checked in.

yellowbrickc:
I have no problem at all with back and forth communication. I rather build something what fulfills some needs than something else. This is why I liked the last change request from @davidmarkclements very much because it shows that you really care not only about coding but also about naming.

Maybe I can give my answers:
1. I don't know if this is the usual way in OSS but this is the usual way for me to build the right features.
2. I like the way how this PR evolved. 
3. Just feedback from me regarding the repo: I made very good experience with a separate test file for edge cases. This is where I would start with recreating bugs and it would be not a unit test but a behavior test set.  

BTW: thank you for the "fireworks" on twitter, I was really flattered - because we know that it wasn't that much what I did.

mcollina:
Pino does not receive many contributions, and this was exceptionally well done.
This PR is an example of why OSS is awesome 👏 .

PR #590, opened 2019-01-21
fix: browser-transmit messages array when asObject is true

### LINKED ISSUE #582: Browser API: logEvent.messages cuts off first/main message
Thank you for building this universal solution. Very nice.

However I'm running into an issue when logging in a browser. The goal is to use the `send` method to post the message to a backend service.

Here is my code:

```js
const logger = require('pino')({
  browser: {
    asObject: true,
    transmit: {
      level: 'warn',
      send(level, logEvent) {
        console.log('the event', level, logEvent.messages);
      }
    }
  }
});

logger.info('hello world');
logger.warn('look out world');
logger.error('goodbye world', 'something else');
```

...and the pino logger output when run both client and server-side (ignoring `pid`/`hostname`/`v`):

```bash
{"level":30,"time":1547591483164,"msg":"hello world"}
{"level":40,"time":1547591483164,"msg":"look out world"}
{"level":50,"time":1547591483164,"msg":"goodbye world something else"}
```
Great!

Also, the `send` function is correctly only running for `warn` and `error` (given the minimum level in the `transmit` configuration). However, the `send` method is not holding the right messages. Here is the `console.log` output:

```
the event warn []
the event error ["something else"]
```

This gives me the idea that the first argument passed into `logger.[verb]()` is ignored. However, the [documentation](https://getpino.io/#/docs/browser?id=transmit-object) seems to say otherwise.

Am I missing something? As this seems like a relatively straightforward implementation. Is there a better way to send the log message to a backend service? We don't necessarily need it as an array but it seems like the correct way to retrieve and send it. We may need to end up making a custom transport but wanted to be sure I wasn't missing anything simple.

Using webpack with default configuration. Node 8.9.x, Chrome latest. I can spin up a demo repo if needed.

Thanks for your time.

justinhelmer:
Fixes #582 

When `asObject` is `true`, and logging with _just_ strings, the `messages` array does not contain all of the messages.

`args.slice()` creates a clone of `args`, but it was only happening in the case of log messages with objects.

This fixes the issue and expands the test case to include all 4 permutations.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -203,18 +203,18 @@ function wrap (opts, logger, level) {
 
 function asObject (logger, level, args, ts) {
   if (logger._serialize) applySerializers(args, logger._serialize, logger.serializers, logger._stdErrSerialize)
-  var msg = args[0]
+  var argsCloned = args.slice()
+  var msg = argsCloned[0]
   var o = { time: ts, level: pino.levels.values[level] }
   var lvl = (logger._childLevel | 0) + 1
   if (lvl < 1) lvl = 1
   // deliberate, catching objects, arrays
   if (msg !== null && typeof msg === 'object') {
-    args = args.slice()
-    while (lvl-- && typeof args[0] === 'object') {
-      Object.assign(o, args.shift())
+    while (lvl-- && typeof argsCloned[0] === 'object') {
+      Object.assign(o, argsCloned.shift())
     }
-    msg = args.length ? format(args.shift(), args) : undefined
-  } else if (typeof msg === 'string') msg = format(args.shift(), args)
+    msg = argsCloned.length ? format(argsCloned.shift(), argsCloned) : undefined
+  } else if (typeof msg === 'string') msg = format(argsCloned.shift(), argsCloned)
   if (msg !== undefined) o.msg = msg
   return o
 }
--- a/test/browser-transmit.test.js
+++ b/test/browser-transmit.test.js
@@ -50,7 +50,38 @@ test('passes send function the logged level', ({ end, is }) => {
 })
 
 test('passes send function messages in logEvent object', ({ end, same, is }) => {
-  const logger = pino({
+  // in order to cover all possible scenarios, need to test messages with/without objects, and with/without the `asObject` option set
+
+  const logger1 = pino({
+    browser: {
+      write: noop,
+      transmit: {
+        send (level, { messages }) {
+          is(messages[0], 'test')
+          is(messages[1], 'another test')
+        }
+      }
+    }
+  })
+
+  logger1.fatal('test', 'another test')
+
+  const logger2 = pino({
+    browser: {
+      asObject: true,
+      write: noop,
+      transmit: {
+        send (level, { messages }) {
+          is(messages[0], 'test')
+          is(messages[1], 'another test')
+        }
+      }
+    }
+  })
+
+  logger2.fatal('test', 'another test')
+
+  const logger3 = pino({
     browser: {
       write: noop,
       transmit: {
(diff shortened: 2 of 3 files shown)


test/browser-transmit.test.js
```diff
@@ -50,7 +50,38 @@ test('passes send function the logged level', ({ end, is }) => {
 })
 
 test('passes send function messages in logEvent object', ({ end, same, is }) => {
-  const logger = pino({
+  // in order to cover all possible scenarios, need to test messages with/without objects, and with/without the `asObject` option set
```
davidmarkclements: can you please break these into separate tests

browser.js
```diff
@@ -203,18 +203,18 @@ function wrap (opts, logger, level) {
 
 function asObject (logger, level, args, ts) {
   if (logger._serialize) applySerializers(args, logger._serialize, logger.serializers, logger._stdErrSerialize)
-  var msg = args[0]
+  var argsCloned = args.slice()
```
davidmarkclements: we need to check the `transmit` functionality alongside `asObject` - because it takes in the `args` array (after modification) - that might be a bug which you've fixed or you might have introduced a bug by cloning.. I'm not sure without checking. Would you mind adding a test for transmit + asObject true with these changes?

test/browser-transmit.test.js
justinhelmer: yes I will do that now

browser.js
justinhelmer: Sorry im not sure what you mean here. 

`transmit` is an object key, but `send` is the function. The three test cases I added cover all the permutations of `send` being called with the correct messages. Are you thinking we need to expand the test cases for `write`? Because we have cases covering that `send` actually gets called with the correct info under the different conditions.

justinhelmer: Perhaps you mean you want to make sure that `write` is being called correctly in the `asObject` scenario?

```js
if (opts.asObject) write.call(proto, asObject(this, level, args, ts))
else write.apply(proto, args)
```

.i.e. `asObject` is returning the correct formatted data?

justinhelmer: let me know if this is what you are looking for, or something different: https://github.com/pinojs/pino/pull/590/commits/139266aebfdc95b73ecca4755bbdc752a08c72f8

justinhelmer: based on your description, that's precisely what I covered in the original set of tests I added, because thats what was **broken** prior to this change:

```js
test('passes send function message strings in logEvent object when asObject is set', ({ end, same, is }) => {
  const logger = pino({
    browser: {
      asObject: true,
      write: noop,
      transmit: {
        send (level, { messages }) {
          is(messages[0], 'test')
          is(messages[1], 'another test')
        }
      }
    }
  })

  logger.fatal('test', 'another test')

  end()
})

test('passes send function message objects in logEvent object when asObject is set', ({ end, same, is }) => {
  const logger = pino({
    browser: {
      asObject: true,
      write: noop,
      transmit: {
        send (level, { messages }) {
          same(messages[0], { test: 'test' })
          is(messages[1], 'another test')
        }
      }
    }
  })

  logger.fatal({ test: 'test' }, 'another test')

  end()
})
```
In these scenarios, both `transmit` and `asObject` are set, and that is what was breaking before this PR (in the case of just strings).

What am I missing here?

davidmarkclements: you're not missing anything, it's me thanks @justinhelmer

justinhelmer:
:-)

davidmarkclements:
v5.10.10 released :D

browser.js
justinhelmer: 👍

justinhelmer:
Thanks @davidmarkclements !

PR #597, opened 2019-02-05
Fix bug Global serializer not invoked for strings when `base: null`

yellowbrickc:


lib/tools.js
```diff
@@ -68,67 +68,64 @@ function asString (str) {
...
-  // to catch both null and undefined
-  const hasObj = obj !== undefined && obj !== null
-  const objError = hasObj && obj instanceof Error
-  msg = !msg && objError === true ? obj.message : msg || undefined
+  var objectWithMessage = Object.assign({}, obj)
+  const objError = obj instanceof Error
+  // Object.assign makes a shallow copy
+  if (objError) objectWithMessage.stack = obj.stack
+  const messageKey = this[messageKeySym]
+  objectWithMessage[`${messageKey}`] = !msg && objError === true ? obj.message : msg || undefined
```
davidmarkclements:
```suggestion
  objectWithMessage[messageKey] = !msg && objError === true ? obj.message : msg || undefined
```

lib/tools.js
```diff
@@ -68,67 +68,64 @@ function asString (str) {
 }
 
 function asJson (obj, msg, num, time) {
-  // to catch both null and undefined
-  const hasObj = obj !== undefined && obj !== null
-  const objError = hasObj && obj instanceof Error
-  msg = !msg && objError === true ? obj.message : msg || undefined
+  var objectWithMessage = Object.assign({}, obj)
+  const objError = obj instanceof Error
+  // Object.assign makes a shallow copy
```
davidmarkclements: `stack` isn't not copied because of shallow copy - it's at the top level 
it's not copied because it's non-enumerable (which is why we have to explicitly output the stack in the logs)

lib/tools.js
```diff
@@ -68,67 +68,64 @@ function asString (str) {
 }
 
 function asJson (obj, msg, num, time) {
-  // to catch both null and undefined
-  const hasObj = obj !== undefined && obj !== null
-  const objError = hasObj && obj instanceof Error
-  msg = !msg && objError === true ? obj.message : msg || undefined
+  var objectWithMessage = Object.assign({}, obj)
```
davidmarkclements: this is likely going to cause bottlenecks, particular for objects with a lot of top level properties

test/basic.test.js
```diff
@@ -248,7 +248,12 @@ test('set the base', async ({ is, same }) => {
 test('set the base to null', async ({ is, same }) => {
   const stream = sink()
   const instance = pino({
-    base: null
+    base: null,
```
davidmarkclements: can you please add this as a separate test case

lib/tools.js
yellowbrickc: I think, I can do w/o copying the object. This means the metadata will contain the property msg/messageKey
Is this acceptable?

yellowbrickc:
I run the both benchmarks a few times and the result was always different, even the measurements for bunyan and co. In the last run some numbers for pino are higher, some are lower.
I'm attaching the 2 text files here. 
[bench_master.txt](https://github.com/pinojs/pino/files/2836042/bench_master.txt)
[bench_issue_479.txt](https://github.com/pinojs/pino/files/2836044/bench_issue_479.txt)

mcollina:
Numbers should be stable, maybe you had something else running on your computer? Docker, Chrome, Slack could all be responsible :/.

davidmarkclements:
.. ok I'm just thinking this through.. can you please correct verify this assumption:

we're making a change to the metadata API (by adding a `msg` property) to solve the global serializer not being invoked when base is null?

@yellowbrickc can you also add comparison benchmarks – as @mcollina says try making sure you have minimum background processes. Also, choose benchmarks that would hit this path and feel free to comment out the benchmarks for other loggers (which also might help with clearer numbers). Even with background processes disabled, they'll always vary slightly, but you're looking for within ~1-2%

yellowbrickc:
I run the file /internal/just-pino.bench.js twice per branch and the numbers are reliable now. They look good I think but it is your call of course
[bench_issue.txt](https://github.com/pinojs/pino/files/2837786/bench_issue.txt)
[bench_master.txt](https://github.com/pinojs/pino/files/2837789/bench_master.txt)

yellowbrickc:
> we're making a change to the metadata API (by adding a `msg` property) to solve the global serializer not being invoked when base is null?

The only change regarding the metadata (the tests prove this) is that `this.lastObj` gets the property `msg`. The global serializer is applied to this object (before it became the `lastObject` ) This way it is ensured that the same logic is applied to the `msg` as to all the other properties to be logged.

But you are right, the connection to `base=null` is not visible. I debugged now again and found the culprit: https://github.com/pinojs/pino/blob/master/pino.js#L88
The serializer is called in tools.js in the method asChindings https://github.com/pinojs/pino/blob/master/lib/tools.js#L148 

I think there is no connection between the `base = null` and the missing call to the serializer, the reproduction through console.log has led to false conclusion. Fact is, the prop msg was logged differently from the rest and this is the bug.

yellowbrickc:
Hi @davidmarkclements @mcollina 
what should we do with this PR, merge or throw away? 
Maybe this bug is not really a bug because nobody seems missing a fix for it...

PR #609, opened 2019-03-16
lock tap dep

mcollina:
FIxes Node 6 support.


jsumners:
What is the Node 6 issue?

mcollina:
We have tests failing on Node 6 because quodaa is conflicting with latest tap.

mcollina:
You are right!

PR #612, opened 2019-03-19
Add pino-browser support for hapi-pino

lidel:
This PR updates `browser.js` to expose/stub symbols and serializer methods expected by `hapi-pino`.

Without this, TypeError happens [here](https://github.com/pinojs/hapi-pino/blob/d50a86766932b0d7b5b168c963f4cc66d4be2aa7/index.js#L6) and [here](https://github.com/pinojs/hapi-pino/blob/d50a86766932b0d7b5b168c963f4cc66d4be2aa7/index.js#L23-L24). 

This is useful in extremely niche use case when one wants to run code with embedded Hapi (HTTP server) in browser context enriched with [chrome.sockets APIs](https://developer.chrome.com/apps/sockets_tcpServer) (such as Chrome Apps or blessed extensions in Chromium-based browsers).

In my case, to run js-ipfs (with embedded hapi) in WebExtension context: https://github.com/ipfs-shipyard/ipfs-companion/issues/664

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -6,6 +6,8 @@ module.exports = pino
 
 var _console = global.console || {}
 var stdSerializers = {
+  wrapRequestSerializer: passthrough,
+  wrapResponseSerializer: passthrough,
   req: mock,
   res: mock,
   err: asErrValue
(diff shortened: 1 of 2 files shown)


mcollina:
Can you add unit test?

lidel:
@mcollina  Added, coverage is 100% again.

PR #614, opened 2019-03-21
transports conversation starter [ RFC ] [ DO NOT MERGE ]

davidmarkclements:
This isn't a PR that should be seriously considered for merging at this point, but with Node 12 near release we have the opportunity to explore `worker_threads` as a means to keep the event loop free while also reducing complexity at the operations level.

This is the beginning of that exploration. It's not a suggestion at an API as yet, purely a conversation starter.

* given that we have an ecosystem of transports it would be useful to have some kind of adapter that can consume these transports
* worker_threads have a `process.stdout` and `process.stdin` settings however `process.stdin` may not have expected properties, like `fd` (hence the need patch, otherwise this line https://github.com/pinojs/pino-pretty/blob/master/bin.js#L43 causes a crash). 
* In any case, it turns out using a worker thread and it's process IO abstraction is more efficient than piping to a separate process 

Taking the benchmark at https://github.com/pinojs/pino-http/blob/master/benchmarks/pino-server.js and running with the following command:

```sh
node benchmarks/pino-server.js | pino-pretty > /dev/null
```

Load testing with autocannon defaults gives approx. 27k req/s on my machine.

If we adapt `pino-server` to the following, using this branch:

```js
var http = require('http')
var server = http.createServer(handle)
var pino = require('pino')
var logger = require('../')({
  logger: pino(pino.transport(require.resolve('pino-pretty/bin')))
})

function handle (req, res) {
  logger(req, res)
  res.end('hello world')
}

server.listen(3000)
```

We can then use the following to get the same output:

```sh
node benchmarks/pino-server.js > /dev/null
```

On my machine, I'm seeing an average of 29k req/s.

While this looks like a trivial no-brainer right now, for moving beyond this point I'm seeing a potentially steep cliff edge.

* If the user doesn't want to log to stdout and wants to set their own destination we need to work out how to inject this into the worker thread. This may lead to using a `SharedArrayBuffer` as an efficient way to pass state. Once we start doing that we need to start using `Atomics` for state synchronization. This could lead to a lot of code, effort and bugs. We could set the `stdout` option to `true` but the overhead of piping from the workers stdout to the dest might not be worth it (also, it means not being able to use SonicBoom I think). 
* The worker thread will keep the process open. If you unref it, the main thread will close the worker thread before it can finish logging. In order to avoid this situation we'll also have to communicate state between main thread and worker thread to determine the idleness of the worker thread (there is no API for this that I can see). 
* There needs to be some thought around API, for instance, what about chaining transports? Remember we can only log to process.stdout from the worker thread. If we set `stdout` to true we can get an `stdout` stream from the worker thread, but then we may be starting to head towards choking the event loop if we have a bunch of piping between worker threads in the main process. 
* A nice to have would be forging some kind of cohesion with a browser API for remote browser logging that fits in with all this.

### DIFF
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -2,10 +2,15 @@
 
 const { Worker } = require('worker_threads')
 
-function transport (name, args = []) {
-  const filename = require.resolve(name)
-  const worker = new Worker(filename, {
-    execArgv: args,
+function transport (filePath, args = []) {
+  args.unshift(filePath)
+  const code = `
+    process.stdin.fd = 1
+    process.argv.push(...${JSON.stringify(args)})
+    require('${filePath}')
+  `
+  const worker = new Worker(code, {
+    eval: true,
     stdin: true
   })
   return worker.stdin
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -1,0 +1,14 @@
+'use strict'
+
+const { Worker } = require('worker_threads')
+
+function transport (name, args = []) {
+  const filename = require.resolve(name)
+  const worker = new Worker(filename, {
+    execArgv: args,
+    stdin: true
+  })
+  return worker.stdin
+}
+
+module.exports = transport
--- a/pino.js
+++ b/pino.js
@@ -4,6 +4,7 @@ const stdSerializers = require('pino-std-serializers')
 const redaction = require('./lib/redaction')
 const time = require('./lib/time')
 const proto = require('./lib/proto')
+const transport = require('./lib/transport')
 const symbols = require('./lib/symbols')
 const { assertDefaultLevelFound, mappings, genLsCache } = require('./lib/levels')
 const {
@@ -121,7 +122,7 @@ function pino (...args) {
 
 pino.extreme = (dest = process.stdout.fd) => buildSafeSonicBoom(dest, 4096, false)
 pino.destination = (dest = process.stdout.fd) => buildSafeSonicBoom(dest, 0, true)
-
+pino.transport = transport
 pino.final = final
 pino.levels = mappings()
 pino.stdSerializers = serializers


davidmarkclements:
CI is failing because obviously: worker_threads are only supported on 11+
CI seems to be failing on 11 as well, but doesn't fail on macOs at least... so... try on mac for now.

Another point: this new API obviously wouldn't be backwards compatible, it should probably throw on older versions (rather than shim to an in-process solution). *Unless* there's a viable worker_threads polyfill (but then.. native add on complexity)

lib/transport.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+const { Worker } = require('worker_threads')
+
+function transport (filePath, args = []) {
+  args.unshift(filePath)
+  const code = `
+    process.stdin.fd = 1
+    process.argv.push(...${JSON.stringify(args)})
```
mcollina: I don’t know where we set with this in core, but it still under debate if a thread doing this will not alter the main thread.

lib/transport.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
+
+const { Worker } = require('worker_threads')
+
+function transport (filePath, args = []) {
+  args.unshift(filePath)
+  const code = `
```
jsumners: Do we really need to use inline code generation?

jsumners:
Atomics?!? What is this madness?

lib/transport.js
mcollina: I have some code to do this that is not doing live code gen. I’ll stick it into a module asap.

mcollina:
Welcome to threads James!

davidmarkclements:
> How would somebody use this?

As a way to consume current transports instead of piping on the command line
> 
> I think this could be done as a module.

yes it could, but this PR is about instigating a new narrative around transports.
What I'd like to get to in Pino 6 is a new recommendation around transports
(at least for Node 12) that plays nicely with serverless, application servers, managed clusters etc.
So if we put something like this in core it would be for user experience/guidance reasons.
The other reason would be if it was developed into allowing other destinations etc, which may need more core alterations/involvement.

This code is just a catalyst, tip of the iceberg to show how we could consume current transports - the idea is to inform a new *type* of transport (Pino Transports v2) as well.

davidmarkclements:
> Atomics?!? What is this madness?

yeah I know, if there's a way to avoid them I'd like to - it depends on whether the performance of `postMessage` and/or process io is roughly equivalent. But we can only know by benchmarking.. which means we have to mess with them to find out.. 😆

lib/transport.js
davidmarkclements: so, if you're not using the `eval` option you can pass `execArgv` and it overrides.
Using `eval` clears all the args (because it's like repl mode), and setting it doesn't affect the parent.

lib/transport.js
davidmarkclements: If you look at the prior commit, it wasn't doing this.
The only reason I did this was patch `process.stdin.fd` so pino-pretty didn't crash.
I did it this way so it's easy to see what's going on, but can we simply load another file (or the same file and check `isMainThread`) - it's no biggie.

jsumners:
I'm not opposed to bringing the transport concept into core. But I think we should just set a hard line and leave the old ones to the way they currently work instead of trying to shoehorn them into a new API. I'd rather see something like:

```js
const logger = pino({
  transports: [require('t1'), require('t2'), ..., require('tN')]
}) 

// some time later
logger.addTransport(require('foo'))
```

davidmarkclements:
> I'm not opposed to bringing the transport concept into core. But I think we should just set a hard line and leave the old ones to the way they currently work instead of trying to shoehorn them into a new API.

* the issue with that is we'd cause a fracture in the ecosystem. Pino + Node 10 and down would have to use the current transports, Pino 6 + Node 12 on could use the new one. 
* since worker threads essentially work performantly with process.stdin/stdout the transports currently use then *why not* shoehorn them, that could still be the best approach for transports. For transport creators, this is only one transport to maintain that can be used as preferred or needed by the node version. 
* we could meet halfway though: the API you suggest could be a light wrapped around each of the transports. But then, we push the effort into the transport modules, and they'd have to use the worker threads to be efficient, but there would nothing stopping them from not doing that, so the constraints may incentivize slower ecosystem transports. So why not reduce the overhead and potential confusion and create an adapter in core for existing transports?

davidmarkclements:
just documenting an update on thinking

* the purpose for using worker threads is to allow transports to be configured in-process to alleviate the operations overhead of piping between processes
* using worker thread transports is obviously going to require async writes, this creates a similar set of trade offs and constraints as extreme mode
* worker threads execution can stop at any point, I'm unclear on whether the worker exit event alone is enough in order to flush the logs on exit – we need to investigate the death cycle of a worker thread in relation to the main thread
* worker threads won't work in some (all??) serverless environments (definitely not aws lambda) so we need some kind of fallback for these environments.
* we want to try to retain backwards compatibility with <= v5 transports, even if it's degraded in some way

It's counterintuitive without absorbing the above points, but this leads me to the following conclusion:

**A Pino v6 transport is a function with synchronously processes a log buffer.**

**A Pino v6 transport function _may_ have asynchronous fire and forget operations within within the synchronous function, but with clearly stated caveats.**

This is the common intersection of all constraints:

* Synchronous execution in a separate thread obviously won't block the main event loop, which is the primary concern keeping logging fast
* In the event that we have to flush (and transform) logs on the final tick (e.g. in failure scenarios such as unhandledException) everything has to happen in that tick, synchronously this can be done in the main thread in the same/similar way to how we handle extreme mode. This will work for writes but not necessarily for tcp sockets (unless there's a library (genuinely) supports sync socket writes). If that's the case we should update pino-socket to provide two modes - fire&forget async for normal operations and sync writes for flushing scenarios
* In a serverless environment, transforms will be applied in the main thread. This is fine since the constraints of the event loop in the context of IO become largely irrelevant in an an environment the spins up a instance per request (or in the event of "lambda warming" it's also largely irrelevant since the process won't be under pressure while being kept alive)
* As a side benefit, this new approach to transports requires exit handling as standard. This means we do away with the concept of extreme mode by making it the default mode.
* For backwards compatibility I'm wondering if we can use wrapper as outlined here to consume older transports, and connect them in a pipeline with v6 transport functions wrapped @mcollina https://github.com/mcollina/syncthrough stream wrapper - which wraps sync functions into a streams compatible interface (it also happens to be a lot faster than streams, with implicit back pressure)
* In also means core internals won't have to drastically change, pino still ultimately sync writes to a buffer per log message, but in v6 transports case it writes to a SharedArray buffer

davidmarkclements:
@mcollina @jsumners wdyt?

mcollina:
In order to send the data down to the worker thread, we will need to serialize and deserialize the data. We have two options:

1. make the log line a JSON in the main thread
2. use the V8 serialization API that is used to marshal data between the thread

Given that the option 2 is faster than generating JSON, however we would have to assemble a full new JS object to be sent down to the Thread. This might be expensive. Currently we are avoiding this by using several piece of strings that are pre-concatenated.

Generating the JSON line in the main thread is much closer to our current operation model, and it would not require a massive rewrite to implement the worker thread transports.

We definitely need to benchmark this and make a call.

---

> In the event that we have to flush (and transform) logs on the final tick (e.g. in failure scenarios such as unhandledException) everything has to happen in that tick, synchronously. This can be done in the main thread in the same/similar way to how we handle extreme mode. This will work for fs writes but not necessarily for tcp sockets (unless there's a library that supports sync socket writes). If that's the case we should update pino-socket to provide two modes - fire&forget async for normal operations and sync writes for flushing scenarios

I do not think this is possible. As soon as a data chunk is moved off to the worker thread, we do not have access to it from the main thread anymore (SharedArrayBuffers might be a thing, but they might also be too expensive, again we need to benchmark.

Essentially this means that we need to keep a `sync` mode available whenever this is needed. In order to keep this viable and performant for file and stdout, I fear we might have to keep JSONification in the main thread, basically defying the purpose of doing this.

davidmarkclements:
> In order to send the data down to the worker thread, we will need to serialize and deserialize the data. 

Not if we use a shared array buffer...

mcollina:
Pino is a _simple_ library. All that queuing seems very complex, with a lot of pieces that could go very wrong.

davidmarkclements:
> Pino is a _simple_ library. All that queuing seems very complex, with a lot of pieces that could go very wrong.

Well.. we took a complicated issue and made it simple to use. Once we go down a threads path, we have to absorb complexity in order to make an easy to use API plus support high performance. There's demand for in-process transports, but we also want to keep things fast. 

simple in-process transport API – low logging overhead – simple implementation... pick two.

Raynos:
It should be noted that sending a buffer to `fs.write()` is probably just as fast as transferring a buffer to a worker thread.

For the main use case of logging to a file descriptor the worker thread shouldn't improve performance.

The problem with this example :

```
# 27k QPS
node benchmarks/pino-server.js | pino-pretty > /dev/null
# 29 QPS
# This pino-server.js puts `pino-pretty/bin` in a worker thread
node benchmarks/pino-server.js > /dev/null
```

^ your logging to stdout here, that's going to be a bottleneck.

What you really want to do is this

```
node benchmarks/pino-server.js > ./benchmarks/log.json &
tail -f ./benchmarks/log.json | pino-pretty > /dev/null
```

Alternatively configure pino to write to `fs.createWriteStream()` or configure pino to write to a file directly using `pino.destination()` or `pino.extreme()`

It's worthwhile to benchmark whether adding worker threads can speed up the "write to a file" use case before adding worker threads to `pino`

jsumners:
I really don't have capacity to weigh in here. I'd have to do too much research to even begin.

jsumners:
Let's punt this to v7. Agreed @davidmarkclements ?

mcollina:
I've remove this from Pino v6. I think we should ship asap and this is a bit too premature.

euglv:
Can you give a rough estimate when this will be implemented?

jsumners:
> Can you give a rough estimate when this will be implemented?

Please read the thread. This question is clearly answered.

mcollina:
> Can you give a rough estimation when this will be implemented?

Would you like to work on it or sponsor its development?

euglv:
@mcollina, I do not have enough time to work on it now. 
But I investigated this further and can confirm that the latest version of `pino` works with [Workers](https://nodejs.org/api/worker_threads.html#worker_threads_worker_threads) out of the box. 

Here is an example of my `logger.js`, that uses `pino-http-send` transport:
```
const {  Worker } = require('worker_threads');
const loggerWorker = new Worker('require("pino-http-send")', {
  stdin: true,
  eval: true,
  argv: ['--url', 'http://localhost:8080']
});

const logger = require('pino')({},{
  write(...args) {
    loggerWorker.stdin.write(...args);
  }
});

module.exports = logger;
```
As I understand [worker.ref()/worker.unref()](https://nodejs.org/api/worker_threads.html#worker_threads_worker_unref) documentation, by default node.js should not stop worker thread when main thread ends. But I believe in rare cases node can crash aborting any code in worker thread, and logging through separate process may be more reliable. 
Here is an example of how to get rid of piping (to use with `pm2`) and use standard transports in separate process:
```
const { exec } = require('child_process');
const loggerProcess = exec('pino-http-send --url http://localhost:8081');

const logger = require('pino')({},{
  write(...args) {
    loggerProcess.stdin.write(...args);
  }
});

module.exports = logger;
```
I did not test the performance of these methods. But I think they should be more performant than Winston.

davidmarkclements:
@euglv yeah it's going to take a lot more than that do safely. The worker thread can indeed be interrupted which means final logs can be lost, you have to block the main thread with an atomic. However doing that all the time is obviously going to cause massive performance bottlenecks. Just treating a worker thread like a spawned process isn't going to get us where we need to go, we need to define a new transport format and then a wrap to wrap old transports while also having an efficient locking mechanism. If you do find time or sponsorship let us know because this is actually a huge task.

valonhaliti:
My idea is for **pino** to log into a file, and a logger worker thread to tail into that file and to save logs in a db somewhere. Is this idea any good?

euglv:
> @euglv yeah it's going to take a lot more than that do safely. The worker thread can indeed be interrupted which means final logs can be lost, you have to block the main thread with an atomic. However doing that all the time is obviously going to cause massive performance bottlenecks. Just treating a worker thread like a spawned process isn't going to get us where we need to go, we need to define a new transport format and then a wrap to wrap old transports while also having an efficient locking mechanism. If you do find time or sponsorship let us know because this is actually a huge task.

I do not think it is possible to create the wrapper around old transports with [atomics](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Atomics)
Old transports are not designed to trigger any event when they finished transporting logs. Unsure where potential wrapper should call `Atomics.notify()` to unblock the main thread.

euglv:
I was not right in my previous message. 
This example do the following:
Main thread creates a worker, worker spawns a new process with transport. 
Shared Int32Array is used for atomics. It have a counter of how many messages were written from worker to transport.
When message is logged with the finalLogger - finalLogger sends message to worker and locks the main thread until worker sends message to transport. 
I did quick tests with uncaught exceptions and `process.exit()` and the following code sends all messages through `pino-http-send` transport:
```
const shared = new SharedArrayBuffer(4);
const sharedArray = new Int32Array(shared);

const {  Worker } = require('worker_threads');

const loggerWorker = new Worker(`
  const MAX_INT32 = 2147483647;
  const workerThreads = require('worker_threads');
  const sharedArray = new Int32Array(workerThreads.workerData.shared);
  const { spawn } = require('spawn-default-shell');
  const loggerProcess = spawn(workerThreads.workerData.cmd);
  workerThreads.parentPort.on('message', (message) => {
    loggerProcess.stdin.write(message, () => {
      let count = Atomics.load(sharedArray, 0);
  
      if (count == MAX_INT32)
        Atomics.store(sharedArray, 0, -MAX_INT32);
      else
        Atomics.add(sharedArray, 0, 1);
  
      Atomics.notify(sharedArray, 0);
    });
  });
`, {
  eval: true,
  workerData: { shared,
  // Configure wrapper
    cmd: 'pino-http-send --url http://localhost:3030/'
  }
});

function waitFor(count){
  let val = Atomics.load(sharedArray, 0);
  while (val != count ){
    Atomics.wait(sharedArray, 0, val);
    val = Atomics.load(sharedArray, 0);
  }
}

const MAX_INT32 = 2147483647;

// How many massages sent by main thread
let msgCount = 0;

const logger = require('pino')({},{
  write(...args) {
    console.log(args[0]);
    msgCount++;
    if (msgCount > MAX_INT32)
      msgCount = -MAX_INT32;
    loggerWorker.postMessage(args[0]);
  }
});

const finalLogger = require('pino')({},{
  write(...args) {
    console.log(args[0]);
    msgCount++;
    if (msgCount > MAX_INT32)
      msgCount = -MAX_INT32;
    loggerWorker.postMessage(args[0]);
    // Wait for all messages to be sent
    waitFor(msgCount);
  }
});
```

davidmarkclements:
hey @euglv you're now closer to a POC I've written (but never published) however you still have the issue with 

* use postMessage for each message (expensive)
*  unless I'm reading it wrong the atomics are blocking for every message (throttles main thread)

if you like I can publish my POC (as a separate repo) and you can PR that repo with additional tests, api docs, and then PR core to integrate it with pino@next, and write the docs etc?

euglv:
Hey @davidmarkclements
In the code above atomics are blocking only `finalLogger`.
`worker.postMessage` makes sense only if we skip JSON stringification in the main thread. But this may be incompatible with custom `Serializers` and objects with methods.  
If we have a large JSON object, for example `const data =  process.report.getReport()`
Then this code:
```
let s = JSON.stringify(data) + '\n';
worker.stdin.write(s);
```
blocks the main thread almost the same way as this code:
```
worker.postMessage(data)
```
But the later code blocks the main thread a little bit less time than the first (1436 ms vs 1732 ms when sending 10000 large JSON objects).
So if transport needs to parse JSON objects - then `worker.postMessage(jsonData)` - will be more efficient. But if transport works with logs as text - then 
```
let s = JSON.stringify(data) + '\n';
worker.stdin.write(s);
```
will be more efficient.
(Tested with node v14.5.0 on Windows)

euglv:
But when logging in Worker thread using this approach:
```
let s = JSON.stringify(process.report.getReport()) + '\n';
worker.stdin.write(s);
```
Does not gives any performance improvement over standard OS piping (`node logger | pino-http-send ...`)

But wrapper over old transports that log using worker and transport in another process - works in two time slower than the standard pino version.

mcollina:
I would recommend to check out how https://github.com/piscinajs/piscina transfer data back and forth. This has been optimized quite a bit.

cc @addaleax

davidmarkclements:
@euglv excellent, well what do you think to taking on my POC and writing tests/docs etc?

euglv:
> I would recommend to check out how https://github.com/piscinajs/piscina transfer data back and forth. This has been optimized quite a bit.
I have looked to this library and think that it is not designed to do asynchronous jobs. And it is slower than `worker.stdin.write(str)`. It wraps everything in `Promises`, which  is not required. Anyway, writing string to shared array (`textEncoder.encodeInto()`) should block the main thread longer than using `worker.stdin.write(str)` because writes to `stdin` are asynchronous.

euglv:
@davidmarkclements 
> @euglv excellent, well what do you think to taking on my POC and writing tests/docs etc?

I have discovered that it is impossible to use `worker.stdin` destination and `Atomics.wait()` for the final logger. `worker.stdin.write()` is asynchronous function, but it starts somewhere on next ticks. 
And this code:
```
worker.stdin.write(str);
Atomics.wait(...);
```
blocks the main thread before `str` is written to the worker thread `stdin`.

But:
```
worker.postMessage(str);
Atomics.wait(...);
```
works as expected. 

I think `worker.stdin` should be used for the regular logging and `postMessage` for the final.

Also `postMessage` should be used to log messages from other workers to the logger worker. I do not know how to share logger worker's `stdin` with the other workers. 

For the above reasons the worker transport should definitely read messages from two sources - `stdin` stream and `on('message')`

davidmarkclements:
POC for transport engine started here: https://github.com/pinojs/pinole

* intended API: see `default-processor.js` - we can do an auto generating thing for transports, but the idea is you specify a file path to a transport processor which is essentially a sink. it can be an async function, normal function or callback-accepting function (currently async/normal supported)
* performance: see `index.js` - the main thread writes into a shared buffer for both meta data and data, the worker thread polls the meta data section of the shared buffer to know when to pass writes into the processor 
* safety: see the flushSync method of `index.js` - we use Atomics.wait (main thread) and Atomics.notify (worker thread) to pause the main thread (which is perfect for final tick), then perform the write in the worker, then update the meta data status so the Atomics.wait is ended
* concept: see the test.js - it writes two messages, and then simulates a filled queue, when you Ctrl+c the flushSync is called on SIGINT which writes out the last pieces via the worker thread (using the blocking mechanism described in prior bullet point)

there's a long way to go and I'm sure there's bugs right now, but this is the essential idea

Once Node 10 is out of LTS (bugger worker threads in 10) we can release this with Pino 7

mcollina:
Can we ship this as a separate module targeting Node 12+ _right now_? Or would this need some changes to pino itself?

davidmarkclements:
no its going require changes to Pino core APIs, we *could* support the transport API in Node 12 and throw if it's attempted to be used on Node 10 but that could be problematic from an expectations-management perspective

davidmarkclements:
closing for https://github.com/pinojs/pino/pull/937

Issue #613, opened 2019-03-20
Top level redacting with censor function


runk:
Currently, if I want to use wildcard redaction with a custom censor function it omits the object from the output.

Broken test for reference:

```js
test('redact – supports top level wildcard', async ({ is }) => {
  const stream = sink()
  const instance = pino({
    redact: {
      paths: ['*'],
      censor: () => '[Redacted]'
      // replacing it with string `censor: '[Redacted]'` solves the problem
    }
  }, stream)
  instance.info({
    req: {
      id: 7915,
      method: 'GET',
      url: '/',
      headers: {
        host: 'localhost:3000',
        connection: 'keep-alive',
        cookie: 'SESSID=298zf09hf012fh2; csrftoken=u32t4o3tb3gg43; _gat=1;'
      },
      remoteAddress: '::ffff:127.0.0.1',
      remotePort: 58022
    }
  })
  const { req } = await once(stream, 'data')
  is(req, '[Redacted]')
})
```

This is the place where it seems to be breaking: https://github.com/pinojs/pino/blob/master/lib/redaction.js#L75 - obviously fn cannot be serialized, which gives `undefined` as result, which leads to situation when object is omitted from output at all.

mcollina:
Would you like to send a PR here or in fast-redact to address this problem?

davidmarkclements:
I’d say make the PR here because pino manages the top level integration
with fast-redact

runk:
It's seems to be an issue of pino itself, so has to be fixed in here perhaps. Let me see what I can do.

PR #617, opened 2019-03-24
Fix wildcard redacting with a custom censor function

runk:
Fixes https://github.com/pinojs/pino/issues/613


runk:
@mcollina what's the process of getting this pr into master?

mcollina:
I was just waiting others to review, but I think we can release this now.

PR #626, opened 2019-04-01
Replace obsolete dependency with an up-to-date one

kibertoad:
Had also to fix Windows support for running tests, otherwise it was impossible to commit, as tests were failing. I can remove dependency replacement and restore it in a later PR after Windows support one is merged and commits are possible, but it's really tiny.


package.json
```diff
@@ -62,11 +62,12 @@
     "benchmark": "^2.1.4",
     "bole": "^3.0.2",
     "bunyan": "^1.8.12",
+    "cross-env": "^5.2.0",
     "docsify-cli": "^4.2.1",
     "execa": "^1.0.0",
     "fastbench": "^1.0.1",
     "flush-write-stream": "^2.0.0",
-    "fresh-require": "^1.0.3",
```
kibertoad: This dependency generates security warning noise during audit, and is entirely unnecessary - there is an up-to-date alternative from well-respected sindresorhus.

test/pretty.test.js
```diff
@@ -8,236 +8,240 @@ const writer = require('flush-write-stream')
...
-  require.cache[require.resolve('pino-pretty')].exports = prettyFactory
-})
-
-test('can send pretty print to custom stream', async ({ is }) => {
-  const dest = new Writable({
-    objectMode: true,
-    write (formatted, enc) {
-      is(/^INFO.*foo\n$/.test(formatted), true)
-    }
+const isWin = process.platform === 'win32'
```
kibertoad: no changes to tests, only "if" block added.

jsumners: Couldn’t they just be avoided by doing a `process.exit(0)` before the tests if the platform is Windows?

kibertoad: Does tap execute all tests in separate processes? Otherwise there's a risk to abort test process prematurely. I thought about doing `return`, but you can't return from non-function.

jsumners: Yes, each test file is run in its own process.

kibertoad: Change implemented, thanks!

test/helper.js
```diff
@@ -7,6 +7,12 @@ const pid = process.pid
 const hostname = os.hostname()
 const v = 1
 
+const isWin = process.platform === 'win32'
+
+function getPathToNull () {
+  return isWin ? '\\\\.\\NUL' : '/dev/null'
+}
```
mcollina: Is that file actually /dev/null? Or we should remove it before exit?

test/pretty.test.js
```diff
@@ -8,6 +8,12 @@ const writer = require('flush-write-stream')
 const { once } = require('./helper')
 const pino = require('../')
 
+const isWin = process.platform === 'win32'
+// skip tests on Windows as colour codes are different and tests fail
+if (isWin) {
```
mcollina: I think you should emit a tap comment here to say why this file is being skipped.

I think the API is tap.comment(str)

test/helper.js
kibertoad: It is, indeed, null: https://stackoverflow.com/questions/11053977/how-can-i-write-to-the-nul-device-under-windows-from-node-js

test/pretty.test.js
kibertoad: Addressed!

kibertoad:
@mcollina What is left before this can be merged?

PR #635, opened 2019-04-04
Remove the msg property after logging

### LINKED ISSUE #633: Pino logger mutate (pollute) my input objects with logging information (immutability)
I'm using pino as logger. Everything was fine util i discovered that pino mutate my object injecting "logging" properties.

Here here a snippet

```

let useful_object = {test: true}
logger.info(useful_object, "my logging information)

console.log(useful_object)

//{"level":30,"time":1554358981834,"pid":27753,"hostname":"mother","test":true,"msg":"my logging information","v":1}
//{ test: true, msg: 'my logging information' }
```

(check that useful_object now has the msg property)

There is a way to configure pino to use immutable objects instead append properties to them? Or i have to clone my objects?

```js
// create an object copy before logging
logger.info({...useful_object}, "my logging information")
```

*That object pollution is worst when i integrate pino in express.js with `express-pino-logger` where a lot of properties are attached to it*

mcollina:
This problem was added in https://github.com/pinojs/pino/commit/e3866516b19667f6ba717eaf2a9fa21a1be24ea0.

Fixes #633.

cc @wdalmut

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -71,7 +71,11 @@ function asJson (obj, msg, num, time) {
   const objError = obj instanceof Error
   const messageKey = this[messageKeySym]
   if (obj === undefined || obj === null) obj = {}
+  let oldMessage
   if (msg) {
+    if (obj.hasOwnProperty(messageKey)) {
+      oldMessage = obj[messageKey]
+    }
     obj[messageKey] = msg
   } else if (objError === true) {
     obj[messageKey] = obj.message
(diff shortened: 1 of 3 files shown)


lib/tools.js
```diff
@@ -131,6 +135,15 @@ function asJson (obj, msg, num, time) {
     }
   }
 
+  if (oldMessage !== undefined) {
+    obj[messageKey] = oldMessage
+  } else {
+    // This delete is not going to cause problems,
+    // because the messageKey is the last property
+    // added. Turbofan <3
```
davidmarkclements: I love that we get to use this hahahaha

PR #637, opened 2019-04-05
Fix 636

### LINKED ISSUE #636: Missing message when using with pretty print
The latest version of pino (5.12.1), this pull request to be exact https://github.com/pinojs/pino/pull/635, causes message to be undefined when passing to pretty-print if logging using only Object like the following:

```
const logger = pino({
  prettyPrint: true,
})
logger.info({ msg: 'test!' })
```

Expected Output:
```
[1554452604021] INFO  (3605 on re7eal-VirtualBox): test!
```

Output:
```
[1554452604021] INFO  (3605 on re7eal-VirtualBox): 
```

mcollina:
Fixes #636.

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -100,38 +100,36 @@ function write (_obj, msg, num) {
   const t = this[timeSym]()
   const messageKey = this[messageKeySym]
   const objError = _obj instanceof Error
+  var obj
 
-  var obj = _obj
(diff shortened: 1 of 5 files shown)


davidmarkclements:
v5.12.2 released

should we deprecate v5.12.1 - that's a pretty nasty bug

Issue #649, opened 2019-05-10
"global is not defined" when used with webpack in the browser


jfahrenkrug:
Hi,

I'm using pino in the browser, in an Angular 7 app that's built with `ng build` (which uses webpack under the hood). It builds fine, but when I run the application and want to log something, I get this error:

```
Uncaught ReferenceError: global is not defined
    at Object.../node_modules/pino/browser.js (vendor.js:197635)
```
This is the line it points to: `var _console = global.console || {}`

I could work around this by using a hack like `window.global = window` or something. But I'm sure there must be a cleaner way. Am I missing something? Is there a setting I should be using?

Thanks!

Johannes

mcollina:
We are not really Angular users.
The browser implementation worked well for some time, but it's definitely possible to improve it. Would you like to send a PR to fix this?

davidmarkclements:
from a technically pure perspective we should change it to `globalThis` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

but ie/edge don't support it (of course.) 

if we remove `global` it will break node based tests. Also if we *just* change it to `window` that limits possibilities of web worker logging (`self`). The globalthis polyfill on npm looks like it will add a lot of payload weight. 

So I think we should @mathiasbynens outrageous polyfill from https://mathiasbynens.be/notes/globalthis:

```js
(function() {
	if (typeof globalThis === 'object') return;
	Object.defineProperty(Object.prototype, '__magic__', {
		get: function() {
			return this;
		},
		configurable: true
	});
	__magic__.globalThis = __magic__;
	delete Object.prototype.__magic__;
}());
```

Although I would do it like this:
```
function polyfillGlobalThis () {
  if (typeof globalThis !== 'undefined') return globalThis
  Object.defineProperty(Object.prototype, 'globalThis', {
    get: function () {
      delete Object.prototype.globalThis
      this.globalThis = this  
    },  
    configurable: true
  })
  return globalThis
}
```

davidmarkclements:
just to make it clear: `polyfillGlobalThis` will both return *and set* globalThis in the global scope
The fact we return `globalThis` is what triggers the getter which is what triggers the setting of .. the global.. `globalThis`. It's absolutely wild.

mcollina:
+1 for the approach

jfahrenkrug:
@mcollina @davidmarkclements Thank you guys for working on this so quickly! Awesome!

[...]

PR #650, opened 2019-05-11
Pino browser fixes

### LINKED ISSUE #647: Provide transpiled browser version
Hello,
I'm trying to use pino in a regular next.js project and it fails due to missing polyfills for IE11. Babel does not transpile node_modules by default and the integration is a bit hacky. I recommend to provide a build for the recent most used browsers http://gs.statcounter.com/.

davidmarkclements:
fixes #647 and #649 

NOTE: MERGE and RELEASE https://github.com/pinojs/hapi-pino/pull/79  *FIRST*

In #612 we added symbols to browser pino. 

This is breaking backward compatibility, and rather than add additional bytes for a polyfill to the payload overhead that was already added by including all symbols I think we should remove the symbols from the browser as they were only included for a "very niche use case" in #612.

We can instead update hapi-pino to use Symbol.for since the serializers symbol is a global symbol.

This also includes the smallest possible polyfill for `globalThis` for a standard-compliant cross platform global, see https://github.com/pinojs/pino/issues/649#issuecomment-491525080

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -162,7 +162,6 @@ pino.levels = {
 }
 
 pino.stdSerializers = stdSerializers
-pino.symbols = require('./lib/symbols')
 
 function set (opts, logger, level, fallback) {
   var proto = Object.getPrototypeOf(logger)
--- a/browser.js
+++ b/browser.js
@@ -4,7 +4,7 @@ var format = require('quick-format-unescaped')
 
 module.exports = pino
 
-var _console = global.console || {}
+var _console = polyfillGlobalThis().console || {}
 var stdSerializers = {
   mapHttpRequest: mock,
   mapHttpResponse: mock,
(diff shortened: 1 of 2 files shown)


jfahrenkrug:
@davidmarkclements  Not that I'm a reviewer or anything, but what do you think about adding a comment to `polyfillGlobalThis()` that links to https://mathiasbynens.be/notes/globalthis? I think if anyone comes across that code down the line they'd be grateful to have a thorough explanation of the zaniness that is going on ;-)

browser.js
```diff
@@ -162,7 +162,6 @@ pino.levels = {
 }
 
 pino.stdSerializers = stdSerializers
-pino.symbols = require('./lib/symbols')
```
mcollina: I have a fear that this makes this change semver-major.
I would prefer it not to be.

davidmarkclements: yep. But I think this is a case of semver purity vs practicality. polyfilling symbols is extra overhead and not fully possible (e.g. typeof symbol). There is only one use case, and it's technically server side - so I think an argument can be made in this case for considering the breaking of browser compat a bug, and removing this to be the solution to that bug. Since  this change along with the hapi-pino change will literally affect no one, as there are no other cases for pino.symbols in the browser.

PR #653, opened 2019-05-16
legacy mobile safari globalThis fallback

### LINKED ISSUE #652: Mobile Safari - issues with globalThis Polyfill
We updated to the latest release today (as of today 5.12.4). We immediately started receiving `"undefined is not an object (evaluating 'this.globalThis=this')"` for anyone using the site on **Mobile Safari**.
We have reverted to a previous release. 

We are using webpack and do not have any other Polyfill for `globalThis`.

Is there anyone who would be able to look at this?

davidmarkclements:
closes #652

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -4,7 +4,7 @@ var format = require('quick-format-unescaped')
 
 module.exports = pino
 
-var _console = polyfillGlobalThis().console || {}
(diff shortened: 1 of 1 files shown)


craigtaub:
Thanks for getting a fix out quickly.

PR #659, opened 2019-05-25
readme: the log object is on the request object

XVincentX:
According to the documentation — the `log` object is on the `reply` object; apparently it's on the `request` one instead.


davidmarkclements:
doc fixes are always welcome - thank you! 🙏

PR #666, opened 2019-06-17
Add pretty write stream creator

ovhemert:
Makes it easier to create a pretty-print stream.

This will prevent ugly client code to get a separate pretty print stream using pino-multi-stream.

See PR pinojs/pino-multi-stream#20 and issue pinojs/pino#534

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -516,6 +516,11 @@ test('auto pino.destination with a string as second argument', async ({ same }) 
   })
 })
 
+test('creates pretty write stream', async ({ is }) => {
+  const prettyStream = pino.createPrettyWriteStream()
+  is(typeof prettyStream.write, 'function')
+})
+
 test('does not override opts with a string as second argument', async ({ same }) => {
   const tmp = join(
     os.tmpdir(),
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -1,4 +1,5 @@
 'use strict'
+
 const os = require('os')
 const { join } = require('path')
 const { readFileSync } = require('fs')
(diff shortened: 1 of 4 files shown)


test/basic.test.js
```diff
@@ -516,6 +516,11 @@ test('auto pino.destination with a string as second argument', async ({ same })
   })
 })
 
+test('creates pretty write stream', async ({ is }) => {
+  const prettyStream = pino.createPrettyWriteStream()
+  is(typeof prettyStream.write, 'function')
+})
+
```
mcollina: Can you add a couple more tests?

1. with a custom prettifier
2. that calls pino-pretty

There should also be a test file specific for pretty-print, so maybe they can live there.

ovhemert: Yes,... will do so. Currently working on the docs already.

mcollina:
We need to have this feature in core because [`prettifierMetaWrapper()`](https://github.com/pinojs/pino/blob/bcbc397464180e811d2d5dbc5e92989c6c6da2b1/lib/tools.js#L165-L240) is in core, and *it must be in core*, because it's deeply tied to Node core internals.

I'm happy if this get change in v6. However this is an issue with pino-multi-stream right now https://github.com/pinojs/pino-multi-stream/pull/20.

IMHO we could use a better name for it.

davidmarkclements:
I see.. in that case I think we should actually expose the API in `pino-multi-stream` - and then support that API via in an interface in core (as we already do for all things pino-multi-stream)

jsumners:
How about dropping support for `pino-multi-stream` in v6? It's been the second sentence of the readme since it was released:

> The purpose of pino-multi-stream is to provide a stop-gap method for migrating from the Bunyan logger.

ovhemert:
@mcollina : moved functionality to pinojs/pino-multi-stream#26

If fine to implement there, let's drop this PR.

davidmarkclements:
Totally agree James - the need for it can be removed by web worker
transports

jsumners:
So plan of action is?:

1. Merge this PR
2. Add an issue to _remove_ this method on the `next` branch
3. Add a deprecation notice in `pino-multi-stream`

mcollina:
I’m -1 to add any deprecation notice until the new mechanism is in

Issue #669, opened 2019-06-21
timestampKey option to customize  timestamp field name


ilya-mim:
Is it possible to add option for customizing timestamp field name?
Reason is that now `destination.lastTime` value calculated wrong if custom `timestamp` function is supplied. It extracts a section of a string begin from 8th character. 

https://github.com/pinojs/pino/blob/master/lib/proto.js#L133

```
timestamp: () => `,"myTime":${Date.now()}`,

destination[Symbol.for('pino.metadata')] = true;
destination.lastTime // <- wrong value
```

mcollina:
Fixed in https://github.com/pinojs/pino/pull/675.

PR #675, opened 2019-07-04
more careful timestamp retrieval

ilya-mim:
Issue [#669](https://github.com/pinojs/pino/issues/669)
It allows to retrieval lastTime value more careful when custom time label is used.

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -12,6 +12,7 @@ const {
   messageKeySym,
   writeSym,
   timeSym,
+  timeSliceIndexSym,
   streamSym,
   serializersSym,
   useOnlyCustomLevelsSym,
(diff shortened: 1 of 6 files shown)


lib/proto.js
```diff
@@ -130,7 +130,8 @@ function write (_obj, msg, num) {
     // it is not needed anymore
     stream.lastMsg = msg
     stream.lastObj = obj
-    stream.lastTime = t.slice(8)
+    const timeMatch = t.match(/\d+/)
```
jsumners: We shouldn’t have a regular expression in this hot path. It would be better to use `.slice` with a correct count.

test/pretty.test.js
```diff
@@ -178,7 +178,7 @@ test('with custom timestamp', async ({ is }) => {
     cb()
```
jsumners: I’m not sure about overloading this test to test two things. I think there should be separate tests to cover both items.

lib/proto.js
ilya-mim: Added `timeSliceIndexSym` with correct count.

PR #695, opened 2019-08-14
feat: add timestampKey option

### LINKED ISSUE #694: Add timestampKey option to customize timestamp field name
A corollary to `messageKey` for time (e.g. `timestampKey`) would be useful. I'm aware that this can be achieved using the [timestamp option](http://getpino.io/#/docs/api?id=opt-timestamp), however, this requires passing in a function, whereas passing in a string (per `messageKey`) is a simpler API for those who are otherwise happy with the default time format (epoch).

Related to #669, #657.

tlvince:
Closes #694

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -172,6 +172,13 @@ for passing in as a value for this option.
 
 **Caution**: attempting to format time in-process will significantly impact logging performance.
 
+<a id=opt-timestampkey></a>
+#### `timestampKey` (String)
+
+Default: `'time'`
+
+The string key for the 'time' in the JSON object.
+
 <a id=opt-messagekey></a>
 #### `messageKey` (String)
 
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -15,6 +15,7 @@ const {
   messageKeySym,
   writeSym,
   timeSym,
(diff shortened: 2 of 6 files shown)


tlvince:
Isn't that the same as the current [timestamp](http://getpino.io/#/docs/api?id=timestamp-boolean-function) option?

Out of interest, what's slow about this implementation?

davidmarkclements:
Every addition of extra state to the context (`this`) has a passive affect on performance, consider that child loggers have to inherit this state etc. etc. Also creating a custom function keeps the unixTime and epochTime faster, because there's less concatenation. 

In any event, speed isn't the main issue here, the main issue is it actually changes the timestamp api - it requires additional documentation of the key being passed in. Which is actually useless for those creating their own functions, because they can specify any key they want as it is. 

So essentially, making timestampKey a convenience option for `timestamp: () => ',"myKey":' + Date.now()'`  makes most sense. 

Also, you'll need to either throw if `timestamp` is supplied as well as `timestampKey` (and update docs to reflect that)

tlvince:
Fair enough. `timestamp` is enough for my use case, so I'll close this.

PR #707, opened 2019-09-11
fix(serializers): properly transfer Symbol based serializers

MickVanDuijn:
When adding serializers through `.child`, the Symbol based serializers were not transferred. For example: https://github.com/pinojs/pino/blob/master/docs/api.md#serializerssymbolforpino-function

Thanks to @ZwaarContrast for helping with debugging!

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -72,11 +72,17 @@ function child (bindings) {
   const instance = Object.create(this)
   if (bindings.hasOwnProperty('serializers') === true) {
     instance[serializersSym] = Object.create(null)
-    for (var k of [...Object.getOwnPropertySymbols(serializers), ...Object.keys(serializers)]) {
+    for (var k in serializers) {
       instance[serializersSym][k] = serializers[k]
     }
-    for (var bk of [...Object.getOwnPropertySymbols(bindings.serializers), ...Object.keys(bindings.serializers)]) {
+    for (var ks of Object.getOwnPropertySymbols(serializers)) {
+      instance[serializersSym][ks] = serializers[ks]
+    }
+    for (var bk in bindings.serializers) {
       instance[serializersSym][bk] = bindings.serializers[bk]
+    }
+    for (var bks of Object.getOwnPropertySymbols(bindings.serializers)) {
+      instance[serializersSym][bks] = bindings.serializers[bks]
     }
   } else instance[serializersSym] = serializers
   if (bindings.hasOwnProperty('customLevels') === true) {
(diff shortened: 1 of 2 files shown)


lib/proto.js
```diff
@@ -72,10 +72,10 @@ function child (bindings) {
   const instance = Object.create(this)
   if (bindings.hasOwnProperty('serializers') === true) {
     instance[serializersSym] = Object.create(null)
-    for (var k in serializers) {
+    for (var k of [...Object.getOwnPropertySymbols(serializers), ...Object.keys(serializers)]) {
       instance[serializersSym][k] = serializers[k]
     }
-    for (var bk in bindings.serializers) {
+    for (var bk of [...Object.getOwnPropertySymbols(bindings.serializers), ...Object.keys(bindings.serializers)]) {
```
mcollina: Can you create only 1 instance of these arrays? Given that we will be iterating an array now, I prefer we use a standard `for (;;)` loop instead, as it's faster than for-of.

davidmarkclements: this will definitely be slower, even with a for loop. `for in` is still fastest way to loop through key names

MickVanDuijn: split the loops into two parts: a `for-in` on the keys, plus a `for(;;)` on the symbols

MickVanDuijn: timings before this PR:
```benchBunyanCreation*10000: 574.174ms
benchBoleCreation*10000: 233.600ms
benchPinoCreation*10000: 241.948ms
benchPinoExtremeCreation*10000: 131.857ms
benchPinoNodeStreamCreation*10000: 250.361ms
benchBunyanCreation*10000: 544.266ms
benchBoleCreation*10000: 225.199ms
benchPinoCreation*10000: 244.614ms
benchPinoExtremeCreation*10000: 127.672ms
benchPinoNodeStreamCreation*10000: 228.082ms```

timings after this PR:
```benchBunyanCreation*10000: 597.158ms
benchBoleCreation*10000: 241.598ms
benchPinoCreation*10000: 239.223ms
benchPinoExtremeCreation*10000: 134.678ms
benchPinoNodeStreamCreation*10000: 237.142ms
benchBunyanCreation*10000: 510.368ms
benchBoleCreation*10000: 200.756ms
benchPinoCreation*10000: 216.045ms
benchPinoExtremeCreation*10000: 119.050ms
benchPinoNodeStreamCreation*10000: 209.216ms```

they seem to heavily fluctuate between runs, but there is no major difference.
**Note:** the child benchmarks don't include `serializers`, so this loop is never hit in those benchmarks

PR #708, opened 2019-09-11
Add MySQL/MariaDB transport

theproductiveprogrammer:
Hi - I've added a transport for MySQL/MariaDB which can extract fields from anywhere in the log stream and map them to database columns. It can also save the entire stream as a JSON field in MySQL. I found this useful for my latest project using the wonderful pino logger.
You can check the release here: https://www.npmjs.com/package/pino-mysql

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -234,7 +234,6 @@ $ node app.js | pino-mysql -c db-configuration.json
 
 `pino-mysql` can extract and save log fields into corresponding database field
 and/or save the entire log stream as a [JSON Data Type][JSONDT]. 
-simple to setup and use.
 
 For full documentation and command line switches read the [readme][pino-mysql].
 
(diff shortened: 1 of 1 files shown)


davidmarkclements:
once @jsumners feedback is addressed and resolved we're good to merge 👍

theproductiveprogrammer:
> Wouldn't you at least want to suggest this schema?
> 
> ```sql
> create table logs (
>   id int(10) unsigned not null auto_increment primary key,
>   created timestamp default current_timestamp,
>   log json,
> 
>   index created (created)
> )
> ```
> 
> And _maybe_ an index on JSON? Maybe also show some generated columns to index specific values in JSON.

I've provided a couple of minimal DB schema's in the README and it also contains some examples of extracting fields from the log.

jsumners:
There is no blockage from me on this PR. I'll just point out that you can save cycles in your transport by defining a column like:

```sql
req_url varchar generated always as ( json_unquote(json_extract(log, '$.req.url')) ) stored
```

Basically, let the database engine take care of extracting the field if it exists.

theproductiveprogrammer:
@jsumners I didn't know MySQL could do that. Thanks! That's very cool. I've added your suggestion to the README as well.

PR #711, opened 2019-09-16
Log literal null values (fixes #706)

### LINKED ISSUE #706: No way to distinguish between null and undefined
Using `pino.info(null)` vs `pino.info(undefined)` results in the same JSON output. This is really unfortunate for debugging logs that might print objects or similar data where a value can be undefined or null. Distinguishing between these values can save lots of time when debugging.

**My suggestion:**
I suggest adding a type property that holds a stringified version of the type present, e.g. `object`, `string`, `undefined` and (even though `typeof` would print `object`) `null`.

I can open a PR for this if needed.

Thanks!

NikxDa:


NikxDa:
@davidmarkclements #706

davidmarkclements:
I think this is also a breaking change... for instance, given `pino.info(foo)`, where `foo` is `null` this will generate JSON with `msg: null` whereas before there would be no `msg` key at all. 

So a transport (or whatever post-processing) that parses the JSON as (`var o = JSON.parse(logLine)`) if it were to do a property presence check `msg in o` or `o.hasOwnProperty('msg')` or even if it were to do `o.msg === undefined` all of these approaches would break with this change.

The problem is, even if we release as a major, transports (or any post processing libs) are peer deps so there's no semver controls in place for transports.. and the question also is, is anyone actually doing that... we can't know because there may be proprietary logging setups.. in which case it probably should be major to communicate the breaking change

NikxDa:
> Would you mind adding a test with `pino.info(null, 'hello world')`? I think `'msg`' should be `'hello world'`, right?

Should it? My understanding was that the `n` parameter in `(o, ...n)` is used as formatting fillers for a format string in `o`. `null` can't be formatted, anyway. What is the expected behavior in this case? I see you are using `quick-format-unescaped` to do the formatting, but a quick look into it didn't give me much information on formatting `null`. @mcollina

davidmarkclements:
http://getpino.io/#/docs/api?id=interpolationvalues-any

>All arguments supplied after message are serialized and interpolated according to any supplied printf-style placeholders (%s, %d, %o|%O|%j) or else concatenated together with the message string to form the final output msg value for the JSON log line.

So given `pino.info({}, 'a', ' message')` the `msg` key should be `a message`.

In the case of `null` – `pino.info(null, 'a', ' message') – we don't want to lose that extra message information.

For instance `pino.info(obj, 'this is obj at point X')` . if `obj` may be null we don't want to lose the message. 

So it does seem like the most reasonable behaviour is to consider `null` to be the same as `{}` when additional args are supplied

davidmarkclements:
I think this would work?

```js
function genLog (z) {
  return function LOG (o, ...n) {
    if (typeof o === 'object') {
      if (o === null) {
        o = {}
      } else if (o.method && o.headers && o.socket) {
        o = mapHttpRequest(o)
      } else if (typeof o.setHeader === 'function') {
        o = mapHttpResponse(o)
      }
      this[writeSym](o, format(null, n, this[formatOptsSym]), z)
    } else this[writeSym](null, format(o, n, this[formatOptsSym]), z)
  }
}
```

jsumners:
@NikxDa are you able to resolve the conflicts here and address @davidmarkclements's feedback?

NikxDa:
Hey @jsumners, thanks for the push. Seems like this got lost on my side somewhere along the way. I'll look into where I got with this tonight!

NikxDa:
@davidmarkclements @jsumners I've added another test for the case that you outlined. The solution David proposed didn't quite work out because an empty object will still not print the literal `null` value, but remove the `msg` key completely. I ended up solving this by omitting the null value when more parameters are passed. Can you please have a look and check whether any edge cases come to mind?I'm not too familiar with the rest of the codebase...

jsumners:
@mcollina ?

jsumners:
Let's call it semver-major.

@NikxDa would you might rebasing onto the `next` branch?

NikxDa:
@jsumners Rebased onto `next` branch and changed the PR base. Seems like the coverage test didn't like the rebase & following force-push too much, but all code tests pass. Lmk if you know a way to fix that failing test :)

jsumners:
Run this to see where the coverage has decreased:

https://github.com/pinojs/pino/blob/d0f953b17f02d4625a069d2edc1f672eb7c028b7/package.json#L24

NikxDa:
@jsumners Coverage hasn't decreased, and `cov-ui` is giving 100% on all items. That's why I mentioned that I think this has something to do with the rebase... Any ideas?

jsumners:
Closed by #777

PR #712, opened 2019-09-17
add Nest Framework documentation

iamolegga:


jsumners:
Can this example be written in vanilla JavaScript?

davidmarkclements:
this is tricky because it's a clash of approaches - Nest is (obviously) pro-transpilation of the server (thus supporting typescript or non-ratified JS proposals like decorators) whereas Pino has a philosophy of supporting the all approaches, which means avoiding transpilation. 

One nit at least, all our examples don't use semi-colons so it should at least subscribe to that for consistency.

davidmarkclements:
another option is to just remove the example and say "see readme"

jsumners:
I literally know nothing about the framework in question. If the example has to be then way then 🤷‍♂️

iamolegga:
Hi, thanks for reviewing this, and thanks for the library!

> Can this example be written in vanilla JavaScript?

As @davidmarkclements said before, It is more idiomatic way to write example in TS, because most of Nest projects are written in TS, and TS-example should be more friendly to Nest-developers.

> One nit at least, all our examples don't use semi-colons so it should at least subscribe to that for consistency.

fixed in last commit

> another option is to just remove the example and say "see readme" - especially because other examples focus on installing into a framework as middleware/plugins and that isn't included here (but is in the readme)

Sorry, I do not understand this sentence, I see, that more _complex_ `hapi` example is good to be here. So I thought Nest example is good too, because it's even smaller.

PR #716, opened 2019-09-30
Better describe `pino-pretty` dependency

briancoit:
`pino-pretty` is currently specified in `devDependencies` only, although it's required at run time . If I understand correctly, the aim is to not increase dependency footprint with `pino-pretty` when it tends to be used for development purposes only.

If my understanding is correct, then I think it makes sense that we specify `pino-pretty` as an optional peer dependency, allowing `pino-pretty` to be optionally provided - this may help with resolving `pino-pretty` when using pnpm (strict non-flat node_modules).

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,5 +91,13 @@
     "pino-std-serializers": "^2.3.0",
     "quick-format-unescaped": "^3.0.2",
     "sonic-boom": "^0.7.5"
+  },
+  "peerDependencies": {
+    "pino-pretty": "^2.4.0"
+  },
+  "peerDependenciesMeta": {
+    "pino-pretty": {
+      "optional": true
+    }
   }
 }


jsumners:
The assertion that it is required at runtime is not technically true. It is required if you deem it so. 

This PR is a duplicate that has already been decided — https://github.com/pinojs/pino/pull/699

briancoit:
> The assertion that it is required at runtime is not technically true. It is required if you deem it so.

It is required if I specify `pretty: true`. Purely from a packaging perspective, that's a dependency which is described nowhere - it is in fact an optional dependency. The only rationale I can see on #699 is that NPM will display a warning - which was acknowledged as an NPM issue/shortcoming . I'm not sure a detail specific to one package manager, should influence how a package describes itself - it addresses the issue in the wrong place IMO. 

Without specifying the peer dependency, we are relying on a NPM/Yarn detail (node_modules flattening) to allow the module to be resolved. As it stands, with pnpm, there is no great way to resolve this outside of flattening modules (as per Yarn/NPM) or injecting the dependency as a child of `pino` via `pnpmfile.js` hooks. We could of course also just disable `pretty` and save ourselves the workarounds - although not ideal.

An alternative solution (while not breaking the dependency tree) might be for the `pretty` config property to allow an instance of the `pino-pretty` module to be passed explicitly, falling back to a local `require('pino-pretty')` when `pretty == true`.

Happy to move discussion to #699 - but of 1000+ dependencies in our monorepo - pino is currently the only package which will fail - I don't think it's something that should be overlooked.

mcollina:
> Without specifying the peer dependency, we are relying on a NPM/Yarn detail (node_modules flattening) to allow the module to be resolved. 

This is incorrect. The Node module resolution algorithm is clearly described in https://nodejs.org/api/modules.html#modules_all_together: under that, resolving a module from the parent dependencies is *not* a NPM detail. It is `pnpm` that deviates from this standard / prohibit this behavior.

> An alternative solution (while not breaking the dependency tree) might be for the pretty config property to allow an instance of the pino-pretty module to be passed explicitly, falling back to a local require('pino-pretty') when pretty == true.

That works for me, would you like to send a PR?

briancoit:
Understood - although I'm not aware of any specific spec around how package managers should place dependencies though.

Happy to raise a PR for that change - but would emphasize that there's still a gap in the package definition in any case. Although we'd allow `pino-pretty` to be provided, there's still no means of constraining the version of `pino-pretty` used.

On the surface, this is a minor issue, but if `pino-pretty` ever see's breaking changes, any issues may not be apparent to the user, and made more difficult to debug particularly in complex monorepo setups where other `pino-pretty` versions in the repo may satisfy a version by chance, until published.

PR #721, opened 2019-10-03
Bind pino instance to prettifier

### LINKED ISSUE #720: Access to pino instance from prettifier function
Hi. Are there any method that will allow to get pino instance from formatter? Probably we can bind it to `this` somewhere when prettifier is constructed?

It could be utilized, for example, for getting levels (especially if there are custom levels).

SkeLLLa:
This PR passes pino instance to prettifier function.

It can be used to access configured levels, for example, here https://github.com/pinojs/pino-pretty/blob/master/lib/colors.js#L32 levels are hardcoded and if there are `customLevels` setup, then it will not work correctly.

Closes #720.

PS: I'm not a fan of `tap`, so tests might look not right.

### DIFF
--- a/docs/pretty.md
+++ b/docs/pretty.md
@@ -23,13 +23,12 @@ The API requires modules provide a factory function which returns a prettifier
 function. This prettifier function must accept either a string of NDJSON or
 a Pino log object. A psuedo-example of such a prettifier is:
 
-Pino instance is passed as `this` into prettifier factory function. But it 
-not initialized yet, so if you need it you may pass it via closure to `prettifier`
-function.
+The uninitialized Pino instance is passed as `this` into prettifier factory function,
+so it can be accessed via closure by the returned prettifier function.
 
 ```js
 module.exports = function myPrettifier (options) {
-  // `this` is bind to pino instance (but not yet initialized).
+  // `this` is bound to the pino instance
   // Deal with whatever options are supplied.
   return function prettifier (inputData) {
     let logObject
--- a/docs/pretty.md
+++ b/docs/pretty.md
@@ -23,8 +23,13 @@ The API requires modules provide a factory function which returns a prettifier
 function. This prettifier function must accept either a string of NDJSON or
 a Pino log object. A psuedo-example of such a prettifier is:
 
+Pino instance is passed as `this` into prettifier factory function. But it 
+not initialized yet, so if you need it you may pass it via closure to `prettifier`
+function.
+
 ```js
 module.exports = function myPrettifier (options) {
+  // `this` is bind to pino instance (but not yet initialized).
   // Deal with whatever options are supplied.
   return function prettifier (inputData) {
     let logObject
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -153,7 +153,8 @@ function asChindings (instance, bindings) {
 
 function getPrettyStream (opts, prettifier, dest) {
   if (prettifier && typeof prettifier === 'function') {
-    return prettifierMetaWrapper(prettifier.call(this, opts), dest)
+    prettifier = prettifier.bind(this)
+    return prettifierMetaWrapper(prettifier(opts), dest)
   }
   try {
     var prettyFactory = require('pino-pretty')
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -153,7 +153,7 @@ function asChindings (instance, bindings) {
 
 function getPrettyStream (opts, prettifier, dest) {
   if (prettifier && typeof prettifier === 'function') {
-    return prettifierMetaWrapper(prettifier(opts), dest)
+    return prettifierMetaWrapper(prettifier.call(this, opts), dest)
   }
   try {
     var prettyFactory = require('pino-pretty')
@@ -294,7 +294,7 @@ function createArgsNormalizer (defaultOptions) {
     }
     if (prettyPrint) {
       const prettyOpts = Object.assign({ messageKey }, prettyPrint)
-      stream = getPrettyStream(prettyOpts, prettifier, stream)
+      stream = getPrettyStream.call(this, prettyOpts, prettifier, stream)
     }
     return { opts, stream }
   }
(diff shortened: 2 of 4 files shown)


docs/pretty.md
```diff
@@ -23,8 +23,13 @@ The API requires modules provide a factory function which returns a prettifier
 function. This prettifier function must accept either a string of NDJSON or
 a Pino log object. A psuedo-example of such a prettifier is:
 
+Pino instance is passed as `this` into prettifier factory function. But it
```
davidmarkclements:
```suggestion
The uninitialized Pino instance is passed as `this` into prettifier factory function, so it can be accessed via closure by the returned prettifier function.
```

docs/pretty.md
```diff
@@ -23,8 +23,13 @@ The API requires modules provide a factory function which returns a prettifier
 function. This prettifier function must accept either a string of NDJSON or
 a Pino log object. A psuedo-example of such a prettifier is:
 
+Pino instance is passed as `this` into prettifier factory function. But it 
+not initialized yet, so if you need it you may pass it via closure to `prettifier`
```
davidmarkclements:
```suggestion

```

docs/pretty.md
```diff
@@ -23,8 +23,13 @@ The API requires modules provide a factory function which returns a prettifier
...
+function.
```
davidmarkclements:
```suggestion
```

docs/pretty.md
```diff
@@ -23,8 +23,13 @@ The API requires modules provide a factory function which returns a prettifier
...
+
 ```js
 module.exports = function myPrettifier (options) {
+  // `this` is bind to pino instance (but not yet initialized).
```
davidmarkclements:
```suggestion
  // `this` is bound to the pino instance
```

lib/tools.js
```diff
@@ -153,7 +153,7 @@ function asChindings (instance, bindings) {
 
 function getPrettyStream (opts, prettifier, dest) {
   if (prettifier && typeof prettifier === 'function') {
-    return prettifierMetaWrapper(prettifier(opts), dest)
+    return prettifierMetaWrapper(prettifier.call(this, opts), dest)
```
davidmarkclements: we should `bind` the prettifier at init time instead of using `call` each time

lib/tools.js
```diff
@@ -294,7 +294,7 @@ function createArgsNormalizer (defaultOptions) {
     }
     if (prettyPrint) {
       const prettyOpts = Object.assign({ messageKey }, prettyPrint)
-      stream = getPrettyStream(prettyOpts, prettifier, stream)
+      stream = getPrettyStream.call(this, prettyOpts, prettifier, stream)
```
davidmarkclements: `createArgsNormalizer` and `getPrettyStream` are only used internally, we can change the function signature without breaking the API.  Therefore pass in the `instance` as another parameter to `createArgsNormalizer` and instead of using call here pass the `instance` directly into the `getPrettyStream` function

pino.js
```diff
@@ -58,7 +58,8 @@ const normalize = createArgsNormalizer(defaultOptions)
 const serializers = Object.assign(Object.create(null), stdSerializers)
 
 function pino (...args) {
-  const { opts, stream } = normalize(...args)
+  const instance = {}
+  const { opts, stream } = normalize.apply(instance, args)
```
davidmarkclements: pass the instance into `createArgsNormalizer` instead of using `apply` here

pino.js
```diff
@@ -96,7 +97,7 @@ function pino (...args) {
   assertDefaultLevelFound(level, customLevels, useOnlyCustomLevels)
   const levels = mappings(customLevels, useOnlyCustomLevels)
 
-  const instance = {
+  Object.assign(instance, {
```
davidmarkclements: can we avoid doing this? It will slow down init time

SkeLLLa: This is called only once and the difference between `Object.assign` and `{}` will be nanoseconds. So Comparing to other code and node itself there will be less than statistical uncertainty.

mcollina: This is just run once.. this might be relevant only for serverless if at all.

davidmarkclements:
That’s assuming that everyone uses child loggers instead of new instances.
I’ve seen the latter in the wild

mcollina:
> That’s assuming that everyone uses child loggers instead of new instances.
I’ve seen the latter in the wild

I've seen that happen only at module scope/maybe 10 or 20 times within a project. In other terms, this does not sit in a hot path for me.

davidmarkclements:
Fair enough - do we know why we’re doing this though? I don’t fully see the
reason

pino.js
mcollina: Actually this has a side effect, as all the definitions of the hidden class would not be done in the same place.
Can you revert?

(Good spot @davidmarkclements).

davidmarkclements:
Oh yeah that too Hahahah

pino.js
SkeLLLa: How could I pass instance to  `createArgsNormalizer` if it's declared in upper scope and I see no way how to pass the instance that's declared in `pino` function into function declared in upper scope.

pino.js
SkeLLLa: If instance is needed to be passed to normalize function then it should be defined before calling normalize. And normalize returns `opts` that are used for constructing the instance. So I don't know what's the way of avoiding this.

mcollina: I think this might not be needed at all.

In https://github.com/pinojs/pino/blob/cf6cf489d6f0b3f5ded6e27ece10febef3525598/lib/tools.js#L174 you have the last logger object used. You might just call https://github.com/pinojs/pino/blob/cf6cf489d6f0b3f5ded6e27ece10febef3525598/lib/tools.js#L237 with `pretty.call(this.lastLogger, ...)` instead.

jsumners:
@davidmarkclements have your concerns been addressed?

davidmarkclements:
@jsumners yeah lgtm

pino.js
jsumners: @SkeLLLa how about resolving this so we can merge?

PR #724, opened 2019-10-08
Added package pino-express

benrei:
Added package pino-express to ecosystem.md#community section


davidmarkclements:
hey @benrei does this offer anything over `express-pino-logger`?

benrei:
Looks like the only difference is that you can customize log output.

davidmarkclements:
We recommend output customisation is performed with transports

benrei:
I see, sounds like a better solution for most cases.

benrei:
I had another look at `express-pino-logger`. Looks like it doesn't have the feature to check for duration time of a request. I think the `pino-express` could be a nice option for smaller projects with a bit more flexibility, than the minimal and fast `express-pino-logger`

davidmarkclements:
hey @benrei if you update the text to state it's an alternative to express-pino-logger and then explain the trade-offs I'd be happy to accept the PR

jsumners:
Closing this as it seems abandoned. Please reopen if you find the time to address the feedback and take this over the finish line.

PR #729, opened 2019-10-12
docs: clarify that pino can be accessed on both request and reply

ralphtheninja:
The text in the documentation says `reply.log` and the code uses `request.log`. Clarify that both can be used.


davidmarkclements:
this should be on fastify :D

davidmarkclements:
oh sorry my bad

PR #730, opened 2019-10-12
always sync flush fatal log messages

davidmarkclements:
it occurred to me today that if you ever write a fatal log, it's because the server is crashing or because you need to crash the server.

e.g.

```js
if (thisIsAnImpossibleSituation) {
  pino.fatal(`there's no where to go, we have to crash`)
  process.exit(1)
}
```

In this scenario, we *always* want to sync flush, and there's no legit scenarios where fatal would be used other than in a death spiral.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -53,7 +53,7 @@ Default: `'info'`
 
 One of `'fatal'`, `'error'`, `'warn'`, `'info`', `'debug'`, `'trace'` or `'silent'`.
 
-Additional levels can be added to the instance via the `customLevels` option. 
+Additional levels can be added to the instance via the `customLevels` option.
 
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -53,7 +53,7 @@ Default: `'info'`
 
 One of `'fatal'`, `'error'`, `'warn'`, `'info`', `'debug'`, `'trace'` or `'silent'`.
 
-Additional levels can be added to the instance via the `customLevels` option.
+Additional levels can be added to the instance via the `customLevels` option.
```
davidmarkclements:
```suggestion
Additional levels can be added to the instance via the `customLevels` option.
```

PR #732, opened 2019-10-23
fix(redaction): fix corner cases with paths

ethanresnick:
In particular, this supports single-item paths where the first item is bracketed (e.g. `{ redact: ["['some-key']"] }`) and paths where one of the segments is the empty string [in brackets] (e.g., `{ redact: ["a[''].x"] }`).

Conceptually, the change is quite simple: it uses the namespace rather, than the string, even for single-segment paths, since the namespace has the brackets + quotes removed. It also changes the namespace `.replace` regex to allow 0 character (i.e., `.*` instead of `.+`) between the quotes.

### DIFF
--- a/lib/redaction.js
+++ b/lib/redaction.js
@@ -19,28 +19,30 @@ function redaction (opts, serialize) {
     rx.lastIndex = 0
     const first = rx.exec(str)
     const next = rx.exec(str)
-    // top level key:
-    if (next === null) {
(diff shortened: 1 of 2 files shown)


ethanresnick:
@mcollina Next steps to getting this merged/published? I'd love to delete my fork :)

mcollina:
@davidmarkclements can you take a look?

PR #733, opened 2019-10-24
test(pretty): add failing test for serializers in children

marcghorayeb:
Hello,

I was recently trying to fix an issue with weird serialization in Pino + Fastify when I actually uncovered that it's an issue with the prettifying stream when developing. Not sure as to what was the intent here but the `prettifierMetaWrapper` in `tools.js` re-serializes every attributes (line 227), even though some of them were serialized by the child. Not sure how to fix this, removing those lines fix my use case but breaks the current tests as well. Let me know how I can help :)

Thanks for the great library!

### DIFF
--- a/test/fixtures/pretty/child-serializers.js
+++ b/test/fixtures/pretty/child-serializers.js
@@ -1,0 +1,18 @@
+global.process = { __proto__: process, pid: 123456 }
+Date.now = function () { return 1459875739796 }
+require('os').hostname = function () { return 'abcdefghijklmnopqr' }
+var pino = require(require.resolve('./../../../'))
+var log = pino({
+  prettyPrint: true,
+  serializers: {
+    foo (obj) {
+      if (obj.an !== 'object') {
+        throw new Error('kaboom')
+      }
+
+      return 'bar'
+    }
+  }
+})
+var child = log.child({ foo: { an: 'object' } })
+child.info('h')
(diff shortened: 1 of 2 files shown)


mcollina:
Good spot! Would you like to work on a fix as well? We’d need to re-apply the serializers there.

marcghorayeb:
> Good spot! Would you like to work on a fix as well? We’d need to re-apply the serializers there.

I can try - I'll try to whip something up by the end of this week.

marcghorayeb:
> Good spot! Would you like to work on a fix as well? We’d need to re-apply the serializers there.

My understanding of that method is a bit limited, specially in regards to the rest of the codebase. My initial thought was that the prettifier stream shouldn't be handling/serializing/redacting anything, am I wrong in this assumption? From looking at this method, I think I am.

jsumners:
I do not understand the test. The second log, `child.log('h')`, causes an uncaught exception. Are you saying that the serializer shouldn't even be invoked for this log invocation?

marcghorayeb:
> I do not understand the test. The second log, `child.log('h')`, causes an uncaught exception. Are you saying that the serializer shouldn't even be invoked for this log invocation?

So it's been some time since I've looked at the code, sorry. From memory, the custom serializers in children get called twice, once when you're instantiating the child, and once when you're logging something, which gives issues in various cases. The test should actually show that. Let me know if that makes more sense.

mcollina:
This seems expected and how serializers should work. Maybe we should document this as a known limitation.

mcollina:
This seems expected and how serializers should work. Maybe we should document this as a known limitation.

mcollina:
Any update on this one?

jsumners:
I don't know what to do with it.

PR #740, opened 2019-11-11
catch potential write errors on fatal flush sync

### LINKED ISSUE #739: "EAGAIN: resource temporarily unavailable, write" on fatal
Before https://github.com/pinojs/pino/pull/730, the use of `fatal` was safe.
After that version, my application crash with `EAGAIN: resource temporarily unavailable, write`.

The linked lines live in sonicboom repo:

https://github.com/mcollina/sonic-boom/blob/e1b31fb93c5a8c306cdc16021e81b4995056a5aa/index.js#L245-L248

This behaviour was considered a patch but the behaviour changes.

Can I may send a PR with a patch (maybe a configuration flag)?

allevo:
fix #739 

This PR allows the developer to choose if flush sync on fatal.
The default is `false`

Bench:
## this branch
```
% npm run bench-basic

> pino@5.13.6 bench-basic /Users/tommasoallevi/Downloads/repo/pino
> node benchmarks/utils/runbench basic

Running BASIC benchmark

benchBunyan*10000: 482.314ms
benchWinston*10000: 364.687ms
benchBole*10000: 348.318ms
benchDebug*10000: 225.755ms
benchLogLevel*10000: 216.953ms
benchPino*10000: 460.753ms
benchPinoExtreme*10000: 185.467ms
benchPinoNodeStream*10000: 175.491ms
benchBunyan*10000: 408.892ms
benchWinston*10000: 283.859ms
benchBole*10000: 181.260ms
benchDebug*10000: 225.092ms
benchLogLevel*10000: 203.503ms
benchPino*10000: 395.198ms
benchPinoExtreme*10000: 178.076ms
benchPinoNodeStream*10000: 148.326ms

==========
BASIC benchmark averages
Bunyan average: 445.603ms
Winston average: 324.273ms
Bole average: 264.789ms
Debug average: 225.423ms
LogLevel average: 210.228ms
Pino average: 427.976ms
PinoExtreme average: 181.772ms
PinoNodeStream average: 161.909ms
==========
System: Darwin/darwin x64 19.0.0 ~ Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz (cores/threads: 8)
```

## master
```
% npm run bench-basic

> pino@5.13.6 bench-basic /Users/tommasoallevi/Downloads/repo/pino
> node benchmarks/utils/runbench basic

Running BASIC benchmark

benchBunyan*10000: 465.974ms
benchWinston*10000: 371.409ms
benchBole*10000: 363.848ms
benchDebug*10000: 235.101ms
benchLogLevel*10000: 216.776ms
benchPino*10000: 389.143ms
benchPinoExtreme*10000: 184.884ms
benchPinoNodeStream*10000: 165.704ms
benchBunyan*10000: 392.238ms
benchWinston*10000: 286.630ms
benchBole*10000: 172.467ms
benchDebug*10000: 223.721ms
benchLogLevel*10000: 204.255ms
benchPino*10000: 389.809ms
benchPinoExtreme*10000: 190.509ms
benchPinoNodeStream*10000: 147.684ms

==========
BASIC benchmark averages
Bunyan average: 429.106ms
Winston average: 329.019ms
Bole average: 268.158ms
Debug average: 229.411ms
LogLevel average: 210.516ms
Pino average: 389.476ms
PinoExtreme average: 187.696ms
PinoNodeStream average: 156.694ms
==========
System: Darwin/darwin x64 19.0.0 ~ Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz (cores/threads: 8)
```

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -27,7 +27,7 @@ const levelMethods = {
       try {
         stream.flushSync()
       } catch (e) {
-        // do nothing
+        // https://github.com/pinojs/pino/pull/740#discussion_r346788313
       }
     }
   },
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -6,7 +6,8 @@ const {
   useLevelLabelsSym,
   changeLevelNameSym,
   useOnlyCustomLevelsSym,
-  streamSym
+  streamSym,
+  forceFlushOnFatalSym
 } = require('./symbols')
 const { noop, genLog } = require('./tools')
 
@@ -23,7 +24,7 @@ const levelMethods = {
   fatal (...args) {
     const stream = this[streamSym]
     logFatal.call(this, ...args)
-    if (typeof stream.flushSync === 'function') {
+    if (this[forceFlushOnFatalSym] && typeof stream.flushSync === 'function') {
       stream.flushSync()
     }
   },
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -6,8 +6,7 @@ const {
   useLevelLabelsSym,
   changeLevelNameSym,
   useOnlyCustomLevelsSym,
-  streamSym,
-  forceFlushOnFatalSym
+  streamSym
 } = require('./symbols')
 const { noop, genLog } = require('./tools')
 
@@ -24,8 +23,12 @@ const levelMethods = {
   fatal (...args) {
     const stream = this[streamSym]
     logFatal.call(this, ...args)
-    if (this[forceFlushOnFatalSym] && typeof stream.flushSync === 'function') {
-      stream.flushSync()
+    if (typeof stream.flushSync === 'function') {
+      try {
+        stream.flushSync()
+      } catch (e) {
+        // do nothing
+      }
     }
   },
   error: genLog(levels.error),
--- a/test/levels.test.js
+++ b/test/levels.test.js
@@ -429,5 +429,4 @@ test('fatal method should call async when sync-flushing fails', ({ equal, fail, 
 
   const instance = pino(stream)
   doesNotThrow(() => instance.fatal(messages[0]))
-  once(stream, 'data')
 })
--- a/test/levels.test.js
+++ b/test/levels.test.js
@@ -429,7 +429,6 @@ test('fatal method should not sync-flushes the destination if sync flushing is d
   instance.fatal('this is fatal')
   await once(stream, 'data')
   doesNotThrow(() => {
-    stream.flushSync = undefined
     instance.fatal('this is fatal')
   })
 })
--- a/test/levels.test.js
+++ b/test/levels.test.js
@@ -409,7 +409,7 @@ test('fatal method sync-flushes the destination if sync flushing is available', 
   stream.flushSync = () => {
     pass('destination flushed')
   }
-  const instance = pino({ forceFlushOnFatal: true }, stream)
+  const instance = pino(stream)
   instance.fatal('this is fatal')
   await once(stream, 'data')
   doesNotThrow(() => {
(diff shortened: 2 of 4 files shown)


davidmarkclements:
Instead of the flag can we use a try/catch of a sync write and then fallback to the non-sync write in the catch

allevo:
Thanks for the fast feedback.
Does `try/catch` drop the performance down?

davidmarkclements:
Not any more, but it doesn’t matter because fatal should only be used when
a process has to die, so it’s never in a hot path

davidmarkclements:
I’m with @mcollina right now, we’re discussing it. If the sync write fails the async write will fail too (but silently). So if we want silent failure of the write it should be wrapped in the try catch but with an empty catch. Or rethrow with a more descriptive error

allevo:
`If the sync write fails the async write will fail too` are you sure? Is this still valid for a network socket?

mcollina:
What's your use case of flushing data on `fatal`? If it's something like the following (which is the correct meaning of a fatal log):

```js
logger.fatal('something went really wrong, crashing')
process.exit(1)
```

In this case, if we remove the `flushSync()`, your fatal log message _might_ not be logged if another write is in progress.
Digging deep specifically in #739, the socket has its buffer full (that's the meaning of `EAGAIN`) and as a result your fatal log message will be lost when you do `process.exit(1)` anyway.

Is a bad crash better or worse than a swallowed log message?

allevo:
It's clear the reason for `flushSync`. I prefer to lose a message instead of crash the app.

NB:
```
logger.fatal('something went really wrong, crashing')
await closeAllFD()
process.exit(1)
```
this lines work differently allowing the fatal log to crash the app without a fd cleaning up.

I've implemented the try catch without the retry.

mcollina:
Can you update the title of this PR?

allevo:
Done

davidmarkclements:
I'm not entirely sure we should swallow the error though, wouldn't we want to let the user know that their final log message wasn't written? Instead shouldn't we rethrow an error explaining that?

lib/levels.js
```diff
@@ -24,7 +24,11 @@ const levelMethods = {
     const stream = this[streamSym]
     logFatal.call(this, ...args)
     if (typeof stream.flushSync === 'function') {
-      stream.flushSync()
+      try {
+        stream.flushSync()
+      } catch (e) {
+        // do nothing
```
davidmarkclements: I actually think we should do something. @mcollina @jsumners what's your thoughts

jsumners: If you can't write a log message that the user will _maybe_ see, what do you intend to do to alert them to that message not being written? From the limited understanding I have of the issue, at this point everything is fucked at this point so 🤷‍♂

davidmarkclements: right but it's worth knowing about? this just swallows it

if we throw an error there will be an understanding that something is wrong

or we can even just `console.warn` if throwing seems too drastic

jsumners: Remove the `try/catch` and it will throw. Wasn't the point of this to remove the throw?

davidmarkclements: yep. but the message thrown has clearly caused confusion, we could throw an error with more of an explanation

jsumners: Sure, sounds good to me.

davidmarkclements: 🤷‍♂ @mcollina

mcollina: I think swallowing the error is ok here. Essentially we can not write a log line that says we are dying by a fatal error. Notifying a developer is usually done via log files.. which we cannot use in this stance. I think this is correct.

jsumners: Compromise: update the useless “do nothing” comment with a link to the discussion.

davidmarkclements:
```suggestion
        // https://github.com/pinojs/pino/pull/740#discussion_r346788313
```

test/levels.test.js
```diff
@@ -417,3 +417,17 @@ test('fatal method sync-flushes the destination if sync flushing is available',
...
+  const messages = [
+    'this is fatal 1'
+  ]
+  const stream = sink((result) => equal(result.msg, messages.shift()))
+  stream.flushSync = () => { throw new Error('Error') }
+  stream.flush = () => fail('flush should be called')
+
+  const instance = pino(stream)
+  doesNotThrow(() => instance.fatal(messages[0]))
+  once(stream, 'data')
```
mcollina: Can you add an assertion here?

allevo:
You are welcome. Opinion sharing is needed to build better software 😄

test/levels.test.js
allevo: Test what?
The return value?

mcollina:
yes!
Il 15 nov 2019, 20:47 +0100, Tommaso Allevi <notifications@github.com>, ha scritto:

test/levels.test.js
allevo: The retuned value should be tested Into synk callback. Or am I missing something?

mcollina: then once() should not be needed

allevo: Addressed

PR #741, opened 2019-11-14
feat(mixin): add mixin config option

watson:
@mcollina here's the "context" PR as we talked about at NodeConf. I ended up calling the feature "mixin" instead, as I think that suits the use-case better. But feel free to suggest other names 😃

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -96,6 +96,27 @@ logger.foo('hi')
 logger.info('hello') // Will throw an error saying info in not found in logger object
 ```
 
+#### `mixin` (Function):
+
+Default: `undefined`
+
+If provided, the `mixin` function is called each time one of the active
+logging methods is called. The function must return an object. The
+properties of the returned object will be added to the logged JSON.
+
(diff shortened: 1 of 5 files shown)


jsumners:
Wouldn't this function need to be synchronous? That should be called out in the docs.

I'm not enamored with the name "mixin" but I don't have a better suggestion either.

watson:
@mcollina

Fixed 👍 

@jsumners:

> Wouldn't this function need to be synchronous? That should be called out in the docs.

Yes, it should be synchronous. My intent was that this was obvious from the docs, as it says that the function "must return an object" and that "the properties of the returned object will be added to the logged JSON". So while I didn't use the word synchronous directly, I thought that the usage could easily be derived. But of you like me to specifically state that it should be called synchronous, I'll do that 👍

jsumners:
I think it is better to be explicit because people do weird things.

watson:
@jsumners I've updated the docs 👍

PR #742, opened 2019-11-14
ci(travis): add Node.js 12+13 to Travis CI test matrix

watson:


mcollina:
@watson this is failing on Node v13. Would you mind adding a patch for that? It's just going to be a test fix, as some node API has changed.

davidmarkclements:
hah! @mcollina this is because we're using `same(obj1, obj2)` - if we used `is` for specific values there'd be no breakage on this one 😆

watson:
@mcollina fixed

PR #743, opened 2019-11-17
CI : Add GitHub action

zekth:
Duration of the matrix build is around 2minutes.
Example : https://github.com/zekth/pino/commit/1bd66552bd64c78585c83596ad11829be3eae497/checks?check_suite_id=314800692

So i've simply moved everything to Github action CI. I see there is a secure env var in the `travis.yml` but the CI is not failing without it; is it not used anymore in the CI process? Otherwise we can use `Secrets` vars in the project setting : https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets

Also do i delete the `travis.yml` ?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,11 +17,11 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Install
         run: npm i
-      - name: run
-        if: !startsWith(matrix.os, 'windows')
+      - name: run with coverage
+        if: (!startsWith(matrix.os, 'windows'))
         run: npm run ci
       # Windows just pass the test without checking the Coverage
       # Due to skipped tests
-      - name: run
+      - name: run without coverage
         if: startsWith(matrix.os, 'windows')
         run: npm run test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,4 +18,10 @@ jobs:
       - name: Install
         run: npm i
       - name: run
+        if: !startsWith(matrix.os, 'windows')
         run: npm run ci
+      # Windows just pass the test without checking the Coverage
+      # Due to skipped tests
+      - name: run
+        if: startsWith(matrix.os, 'windows')
+        run: npm run test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-16.04]
+        os: [macOS-10.14, windows-latest, ubuntu-latest]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
...
+on: [push, pull_request]
+jobs:
+  build:
+    name: ${{ matrix.node-version }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-16.04]
+        node-version: [6, 8, 10, 11]
```
mcollina: Would you mind adding 12 and 13?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,21 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  build:
+    name: ${{ matrix.node-version }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-16.04]
```
mcollina: Would you mind using ubuntu-latest, and add WIndows and Mac OS X as well?

zekth:
I've added node 12 / 13 and win and mac os. Let's see the result of the run : https://github.com/zekth/pino/commit/f82147bcd4f6b2a40805c123d77ff9a53eb188cc/checks?check_suite_id=314805922

mcollina:
> For win CI the threshold may need to be adjusted

Please proceed.

> `# Subtest: http request support via serializer without request connection`

This is due to a change in Node core. We have to change this to have `is` assertions for specific values. Would you mind to take care of that change?

mcollina:
> This is due to a change in Node core. We have to change this to have is assertions for specific values. Would you mind to take care of that change?

I implied only for that specific test :). SOrry for the confusion.

zekth:
Everything is green now :)
https://github.com/zekth/pino/commit/4726199ca7ed564d4ab63573c0ef108a160dcb3a/checks?check_suite_id=315896709

zekth:
> > This is due to a change in Node core. We have to change this to have is assertions for specific values. Would you mind to take care of that change?
> 
> I implied only for that specific test :). SOrry for the confusion.

Yeah moved to `isLike`

zekth:
We can set ubuntu as referall for coverall don't you think?

jsumners:
Yes.

zekth:
I'll do it in a further PR. Do we keep travis for the transition period?

jsumners:
Sounds like a reasonable thing to do.

zekth:
Seems like the tests for windows are not stable : https://github.com/pinojs/pino/runs/307985680

I'll try to find out why

PR #744, opened 2019-11-18
CI : Github action improvement

zekth:
As discussed in https://github.com/pinojs/pino/pull/743#pullrequestreview-318291555 Adding the coverage job.

So it's on the same yml file because it's triggered by the same events: 
```yaml
on: [push, pull_request]
```
But it can be splitted. Also @jsumners in your example you do the matrix job but is it revelant to have it here? Because we want to test once the coverage and not the compatibility on each node version like in the `build` job. Do we keep it on only one node version like LTS ?

Note: Also increased the `sleep` in the `pino.destination` test because IO in the windows image seems to be slower and make the tests fail; with `300` it passes regularly.

Last run for ref : https://github.com/zekth/pino/commit/3aa8b0ff1a6110705638e3abeb786de44378da85/checks?check_suite_id=316635812

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,30 +20,19 @@ jobs:
       - name: Tests
         run: npm run test
   coverage:
-    name: coverage Node ${{ matrix.node-version }}
-    runs-on: ${{ matrix.os }}
-    strategy:
-      fail-fast: false
-      matrix:
-        os: [ubuntu-latest]
-        node-version: [6, 8, 10, 11, 12, 13]
+    name: coverage
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js 12
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: 12
       - name: Install
         run: npm i
       - name: run with coverage
         run: npm run cov-ci
-      - name: Coveralls Parallel
+      - name: Coveralls
         uses: coverallsapp/github-action@master
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -9,6 +9,24 @@ jobs:
...
+      - name: Tests
+        run: npm run test
+  coverage:
+    name: coverage Node ${{ matrix.node-version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        node-version: [6, 8, 10, 11, 12, 13]
```
jsumners: There's no need to run coverage this many times.

zekth:
> The parallel configuration is necessary here because of the matrix of jobs. See https://github.com/coverallsapp/github-action/#complete-parallel-job-example
> 
> My suggestion to have a `coverage.yml` would result in, basically:
> 
> ```yaml
> name: Coverage
> runs-on: ubuntu-latest
> steps:
>   - uses: actions/checkout@v1
>   - uses: actions/setup-node@v1
>   - run: npm i
>   - run: npm run cov-ci
>   - uses: coverallsapp/github-action@master
>      with:
>         github-token: ${{ secrets.GITHUB_TOKEN }}
> ```

Yeah totally agree, that's why i've asked in the PR. So which version of node you suggest? 12 (LTS) ?

zekth:
CI seems still a bit flaky on Windows : https://github.com/pinojs/pino/runs/313765404

Increasing the sleep time? Seems like IO is randomly slow in this env.

mcollina:
would you mind to send a PR?
Il 26 nov 2019, 13:45 +0000, Vincent LE GOFF <notifications@github.com>, ha scritto:

PR #748, opened 2019-12-01
CI : Move from sleep to watch exists

zekth:
As mentionned here https://github.com/pinojs/pino/pull/744#issuecomment-558634866

Windows CI seems a bit flaky on the IO. I've extended the sleep time for `pino.destination` tests.

Hope this time it will be stable on all the runs.

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -479,7 +479,7 @@ test('pino.destination', async ({ same }) => {
   )
   const instance = pino(pino.destination(tmp))
   instance.info('hello')
-  await sleep(300)
+  await sleep(500)
   const result = JSON.parse(readFileSync(tmp).toString())
   delete result.time
   same(result, {
@@ -498,7 +498,7 @@ test('auto pino.destination with a string', async ({ same }) => {
   )
   const instance = pino(tmp)
   instance.info('hello')
-  await sleep(300)
+  await sleep(500)
   const result = JSON.parse(readFileSync(tmp).toString())
   delete result.time
   same(result, {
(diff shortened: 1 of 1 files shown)


zekth:
> lgtm.
> 
> However we should really change these tests to not have a sleep.

We can move to `fs.watchfile` and hook on the creation of the file? This can be added into an util func with a timeout argument. https://nodejs.org/docs/latest-v6.x/api/fs.html#fs_fs_watchfile_filename_options_listener

mcollina:
it would be really cool!
Il 1 dic 2019, 12:01 +0100, Vincent LE GOFF <notifications@github.com>, ha scritto:

zekth:
> it would be really cool! Il 1 dic 2019, 12:01 +0100, Vincent LE GOFF <notifications@github.com>, ha scritto:
> […](#)
> > lgtm. > However we should really change these tests to not have a sleep. We can move to fs.watchfile and hook on the creation of the file? This can be added into an util func with a timeout argument. https://nodejs.org/docs/latest-v6.x/api/fs.html#fs_fs_watchfile_filename_options_listener — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe.

Ok so don't merge this one. I'll tackle it!

zekth:
So i've extended the timeout for the wait but it resolves when it finds the file, so on most of platforms it would be really quicker. I've moved to `exists `because `Watch `needs the file to already exists.

I'm ok with this current approach. What about you?

test/basic.test.js
```diff
@@ -1,14 +1,29 @@
...
+      resolve()
+    }
+  }, 100)
+  const timer = setTimeout(() => {
+    clearTimeout(timer)
+    clearInterval(interval)
+    reject(new Error(`${filename} was not created with a Timeout:${timeout}`))
+  },
+  timeout
+  )
```
mcollina: Would you mind removing the timer and just add a counter in the interval?
we can also move the timeout to a constant inside this function.

zekth: I've pushed a change. I thinks it fit your review, or may i've misunderstood 🙃

PR #754, opened 2019-12-12
feat(bindings): allow setting of bindings

erikwlarsen:
I'm in the process of replacing a node-bunyan logger, and pino fits perfectly. Only issue is that we have a use case for dynamically setting bindings. Let me know if this or something like it (or something completely different that gets the job done) would be a good fit for pino. Thanks!

### DIFF
--- a/test/basic.test.js
+++ b/test/basic.test.js
@@ -69,6 +69,20 @@ test('set bindings on child instance', async ({ same }) => {
   same(child.bindings(), { name: 'basicTest', foo: 'bar' })
 })
 
+test('child should have bindings set by parent', async ({ same }) => {
+  const instance = pino({ name: 'basicTest', level: 'info' })
+  instance.setBindings({ foo: 'bar' })
+  const child = instance.child({})
+  same(child.bindings(), { name: 'basicTest', foo: 'bar' })
+})
+
+test('child should not share bindings of parent set after child creation', async ({ same }) => {
+  const instance = pino({ name: 'basicTest', level: 'info' })
+  const child = instance.child({})
+  instance.setBindings({ foo: 'bar' })
+  same(child.bindings(), { name: 'basicTest' })
+})
+
 function levelTest (name, level) {
   test(`${name} logs as ${level}`, async ({ is }) => {
     const stream = sink()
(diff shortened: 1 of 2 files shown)


test/basic.test.js
```diff
@@ -51,6 +51,24 @@ test('bindings contain the name and the child bindings', async ({ same }) => {
...
 
+test('set bindings on instance', async ({ same }) => {
+  const instance = pino({ name: 'basicTest', level: 'info' })
+  instance.setBindings({ foo: 'bar' })
+  same(instance.bindings(), { name: 'basicTest', foo: 'bar' })
+})
+
+test('newly set bindings overwrite old bindings', async ({ same }) => {
+  const instance = pino({ name: 'basicTest', level: 'info', base: { foo: 'bar' } })
+  instance.setBindings({ foo: 'baz' })
```
jsumners: What happens if you create a new child after invoking this function?

erikwlarsen: Good question -- the new child should have the same bindings as its parent at the point of the child's creation, since these are ultimately instance properties. Later changes to the parent's bindings should not affect the child. I'll add a couple more tests to illustrate.

test/basic.test.js
```diff
@@ -51,6 +51,38 @@ test('bindings contain the name and the child bindings', async ({ same }) => {
...
+  instance.setBindings({ foo: 'bar' })
+  const child = instance.child({})
+  same(child.bindings(), { name: 'basicTest', foo: 'bar' })
+})
+
+test('child should not share bindings of parent set after child creation', async ({ same }) => {
+  const instance = pino({ name: 'basicTest', level: 'info' })
+  const child = instance.child({})
+  instance.setBindings({ foo: 'bar' })
+  same(child.bindings(), { name: 'basicTest' })
```
mcollina: Can you add a test that verifies the instance bindings?

erikwlarsen: yep, done!

erikwlarsen:
Great, thanks for reviewing. What's the process for merging/releasing?

jsumners:
Patience.

PR #756, opened 2019-12-28
fix(redaction): remove undefined entries for removed number and boolean values

tuckbick:
This fixes an issue where removing a number or boolean value at a top-level path results in an `undefined` value instead of fully removing the entry.

`hapi-pino` issue for context: https://github.com/pinojs/hapi-pino/issues/94

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -105,8 +105,7 @@ function asJson (obj, num, time) {
         // this case explicity falls through to the next one
         case 'boolean':
           if (stringifier) value = stringifier(value)
-          data += ',"' + key + '":' + value
-          continue
+          break
         case 'string':
           value = (stringifier || asString)(value)
           break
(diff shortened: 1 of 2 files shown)


lib/tools.js
```diff
@@ -105,8 +105,7 @@ function asJson (obj, num, time) {
         // this case explicity falls through to the next one
```
jsumners: This comment is no longer relevant.

tuckbick: How so? It looks like it does explicitly fall through to the next case. I also don't see how it is relevant to this PR.

jsumners: I thought this comment was associated with the `'boolean'` case.

tuckbick: Ahh, gotcha. Thanks giving a look!

PR #759, opened 2020-01-07
Support new nestedKey config to avoid property name collisions with pino standard properties when logging objects

matthewadams:
Some people log objects a lot.  Those objects, in general, may have properties that conflict with pino's standard logging fields (`level`, `v`, `pid`, etc).  Also, when searching logs for logged objects, it's nice to always have them in a consistent location (that is, property of the root log record object).

This PR adds a new config option ~~`objectKey`~~ `nestedKey` that causes log records to include a key whose name is the value of the ~~`objectKey`~~ `nestedKey` config, and whose value is the logged object (if it is an object and not a string).  It allows folks using things like Kibana, StackDriver, MongoDB, etc, to consistently know the path in their log records to search for, and guarantees that no logged object's properties will conflict with pino's standard properties.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pino",
-  "version": "5.15.0",
+  "version": "5.16.0",
```
mcollina: We bump this on release, can you please revert the version change?

docs/api.md
```diff
@@ -201,6 +201,32 @@ Default: `'msg'`
 
 The string key for the 'message' in the JSON object.
 
+<a id=opt-objectkey></a>
+#### `objectKey` (String)
+
+Default: `null`
+
+If there's a chance that objects that you're logging have properties that conflict with those from pino itself (`level`, `timestamp`, `v`, `pid`, etc)
```
davidmarkclements: to keep consistency with the docs can you use the third person objective impersonal rather than second person perspective (e.g. don't use you). See other parts of the docs for an example of description perspective

docs/api.md
```diff
@@ -201,6 +201,32 @@ Default: `'msg'`
 
 The string key for the 'message' in the JSON object.
 
+<a id=opt-objectkey></a>
+#### `objectKey` (String)
```
davidmarkclements: `objectKey` is somewhat vague, can we change the name to `nestedKey`

package.json
matthewadams: Already did.

docs/api.md
matthewadams: renamed

docs/api.md
matthewadams: done

mcollina:
@davidmarkclements are you good with this now?

davidmarkclements:
> @davidmarkclements are you good with this now?

yes

davidmarkclements:
released https://github.com/pinojs/pino/releases/tag/v5.16.0 thank you @matthewadams

PR #767, opened 2020-02-03
experimental implementation of setting log level on root logger

### LINKED ISSUE #766: change log level globally or on root logger
This is somewhat related to #681 

I'd like to bring this discussion up again.

Let's say I have a bunch of loggers:

```js
const rootLogger = pino({level: 'info'});

const childLogger = rootLogger.child({name:'thechild'});
```

Currently one change the log level at runtime doing
`parentLogger.level = 'debug'`

Unfortunately this change in level doesn't propagate to the child logger and in order to really change the log level on the root logger and all children one has to somehow keep track of the children or use a proxy hack as described in #681.

I'm wondering if we can change pino to make level changes propagate to all children of a root logger or define the log level globally?

I think the current behaviour of the log level change not propagating to children is confusing and not useful.

I have a good enough understanding of the code base that I could create a PR that would change this so that level changes on the root propagate to children.

Also for reference, this actually **works** the way I described in winston@3:
```js
const winston = require("winston");

const rootLogger = winston.createLogger({
  level: "info",
  format: winston.format.json(),
  transports: [
    new winston.transports.Console({
      format: winston.format.json(),
    }),
  ],
});

const childLogger = rootLogger.child({ name: "child" });

rootLogger.level = "debug";

childLogger.debug("this is a debug message that will be logged");
```

cc @zdila

geekflyer:
do not merge

would technically fix #766

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -8,16 +8,9 @@ const {
   useOnlyCustomLevelsSym,
   streamSym
 } = require('./symbols')
-const { noop, genLog } = require('./tools')
+const levels = require('./standard-levels');
+const { genLog } = require('./tools')
 
-const levels = {
-  trace: 10,
-  debug: 20,
-  info: 30,
-  warn: 40,
-  error: 50,
-  fatal: 60
-}
 const logFatal = genLog(levels.fatal)
 const levelMethods = {
   fatal (...args) {
@@ -89,10 +82,6 @@ function setLevel (level) {
   const useOnlyCustomLevelsVal = this[useOnlyCustomLevelsSym]
 
   for (var key in values) {
-    if (levelVal > values[key]) {
-      this[key] = noop
-      continue
-    }
     this[key] = isStandardLevel(key, useOnlyCustomLevelsVal) ? levelMethods[key] : genLog(values[key])
   }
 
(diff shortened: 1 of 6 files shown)


lib/levels.js
```diff
@@ -89,10 +82,6 @@ function setLevel (level) {
   const useOnlyCustomLevelsVal = this[useOnlyCustomLevelsSym]
 
   for (var key in values) {
-    if (levelVal > values[key]) {
-      this[key] = noop
-      continue
-    }
```
mcollina: This change is going to slow us down significantly in case a log level is disabled. Changing this is an hard no for me.

geekflyer: are you sure this slowing anything down _significantly_? 
Effectively this will simply replace `() => {}` with 
```
    if (levelValOfMethod < levels[this[rootLevelSym]]) {

      // don't write the log if this method's log severity is lower than what the logger is configured to write
      return;
    }
```

where `levelValOfMethod < levels[this[rootLevelSym]]` is the "expensive" part if you will. This will walk up the prototype chain to get property and then do a numerical comparison. I understand that this adds some extra computation but I doubt this is slowing things down _significantly_ . More than happy to benchmark this though.

mcollina: using a no-op and specifically _the same no-op_ is a deliberate decision to provide a good hint to Turbofan so that it can optimize disabled log statements away.

mcollina:
Closing for no activity, feel free to reopen.

PR #772, opened 2020-02-13
Deprecate the `changeLevelName` option and alias it to `levelKey`

wraithy:
There was some discussion in [pino-pretty](https://github.com/pinojs/pino-pretty/pull/92) about this.

I'm not sure what sort of approach is normally used when deprecating something, but I thought I'd open a PR just to get the ball rolling.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -254,14 +254,18 @@ Enables printing of level labels instead of level values in the printed logs.
 Warning: this option may not be supported by downstream transports.
 
 <a id="changeLevelName"></a>
-#### `changeLevelName` (String)
+#### `changeLevelName` (String) - DEPRECATED
+Use `levelKey` instead. This will be removed in v7.
+
+<a id="levelKey"></a>
+#### `levelKey` (String)
 
 Default: `'level'`
 
 Changes the property `level` to any string value you pass in:
 ```js
 const logger = pino({
-  changeLevelName: 'priority'
+  levelKey: 'priority'
 })
 logger.info('hello world')
 // {"priority":30,"time":1531257112193,"msg":"hello world","pid":55956,"hostname":"x","v":1}
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -287,6 +287,11 @@ function createArgsNormalizer (defaultOptions) {
     if ('onTerminated' in opts) {
       throw Error('The onTerminated option has been removed, use pino.final instead')
     }
+    if ('changeLevelName' in opts) {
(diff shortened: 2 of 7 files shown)


docs/api.md
```diff
@@ -254,14 +254,18 @@ Enables printing of level labels instead of level values in the printed logs.
 Warning: this option may not be supported by downstream transports.
 
 <a id="changeLevelName"></a>
-#### `changeLevelName` (String)
+#### `changeLevelName` (String) - DEPRECATED
+Use `levelKey` instead. This will be removed in v6.
```
wraithy: I added this as a placeholder. Comments/edits welcome.

lib/tools.js
```diff
@@ -287,6 +287,11 @@ function createArgsNormalizer (defaultOptions) {
     if ('onTerminated' in opts) {
       throw Error('The onTerminated option has been removed, use pino.final instead')
     }
+    if ('changeLevelName' in opts) {
+      console.warn('The changeLevelName option is deprecated and will be removed in v6. Use levelKey instead.')
+      opts.levelKey = opts.changeLevelName
+      delete opts.changeLevelName
+    }
```
wraithy: Input welcome here.

test/levels.test.js
```diff
@@ -257,12 +258,16 @@ test('resets levels from labels to numbers', async ({ is }) => {
   instance.info('hello world')
 })
 
+test('aliases changeLevelName to levelKey', async ({ is }) => {
+  is(pino({ changeLevelName: 'priority' })[levelKeySym], 'priority')
```
wraithy: If there is a preference to avoid looking at "private" properties, there are other ways this can be tested.

docs/api.md
jsumners: Deprecation would last _through_ v6. So it would be removed in v7.

test/levels.test.js
davidmarkclements: yeah if possibly avoid private state in tests

davidmarkclements:
@jsumners deprecation warnings from other modules won't use the logger, however instead of `console.warn` it should be `process.emitWarning` 

As an off-thread aside, maybe a module (or even a pino option) for wrapping emitWarning into logger.warn would be good.

wraithy:
So in version 7, this aliasing code will need to be removed. Do you guys have a way of tracking that in Github? Should I leave a TODO in the code?

jsumners:
I created a project -- https://github.com/pinojs/pino/projects/2

Issue #764, opened 2020-01-29
Proposal: add format support


delvedor:
Hello everyone!
At Elastic, we have a custom format for JSON logs named [ECS](https://www.elastic.co/guide/en/ecs/current/index.html), it helps us run some automation behind the scenes and improve the user experience.
I'm working on a set of libraries that make the most commonly used logging frameworks to log directly into an ECS compliant format, but currently, it is not possible doing it for Pino, since the json structure is fixed, and the serializers are not powerful enough for solving the issue.

What do you think about introducing the `format` concept?
It would be an object that allows you to hook into Pino and specify how you want the data to be serialized.
Following an example, so it's more clear what I would like to implement.
```js
const Pino = require('pino')({
  format: {
      // called everytime createChild is executed
      onChild (level, pid) {
        return `'{"level":"${level}","pid":"${pid}"}`
      },
      // what the user logs via log.info(‘messge’, { data })
      onLog (mergingObject, message, ...interpolationValues) {
        // return serialized data
      },
      onTime (time) {
        return `'{"@timestamp":"${new Date(time).toISOString()}"`
      }
  }
})
```

With this feature, we can also start logging format different than JSON:
```js
const Pino = require('pino')({
  format: {
      onBase (level, pid) {
        return `'level: ${level} - pid ${pid}`
      },
      onLog (mergingObject, message, ...interpolationValues) {
        // return serialized data
      },
      onTime (time) {
        return `@timestamp - ${new Date(time).toISOString()}`
      }
  }
})
```

What do you think? If you would like to move forward with this feature, I can work on it.

Some additional points:
- This can simplify the meta protocol defined in http://getpino.io/#/docs/api?id=destinationsymbolfor39pinometadata39 as it should be designed to potentially side-step the internal rendering, simplifying the implementation of https://github.com/pinojs/pino-multi-stream for example.
- This will simplify the implementation of the “prettifierMetaWrapper” https://github.com/pinojs/pino/blob/6068257bbc8e3ebb73fe8f941e0e56a3eadb75d1/lib/tools.js#L168
- This will likely be a semver-major change

jsumners:
I suspect this will incur undesirable performance impacts. But it also feels related to https://github.com/pinojs/pino/pull/614

However, we are not opposed to any improvements that do not negatively impact performance.

nuragic:
Ciao Tomas @delvedor,

[...]

PR #775, opened 2020-02-18
Add formatters for core log components

delvedor:
Hello everyone!
This pr is the continuation of the proposal in https://github.com/pinojs/pino/issues/764.
Once I wrote the first PoC, I've changed quite a bit the approach. Following a brief explanation of what ported me to write this proposal.

As you know, one of the reasons why Pino is very fast™ is because there is a lot of caching going on, which makes impossible to have a single endpoint where a user can specify a format of the logs. Probably the best solution to this problem is https://github.com/pinojs/pino/pull/614.
The most powerful tool that Pino is offering for addressing the "format" problem, are the serializers. The only issue is that the serializers are for everything that will be passed in the pino methods, but not for the base log structure, such as the log level and all the bindings.
This proposal adds a serializer also for the level handling and the bindings, and it works as follows:
```js
const {
  bindingsSerializerSym,
  levelSerializerSym
} = require('./lib/symbols')

const pino = require('./')({
  serializers: {
    [bindingsSerializerSym] (bindings) {
      return {
        process: { pid: bindings.pid },
        hostname: bindings.hostname
      }
    },
    [levelSerializerSym] (label, number) {
      return {
        log: { level: label }
      }
    }
  }
})

pino.info('hello world')

const child = pino.child({
  foo: 'bar',
  nested: { object: true },
  serializers: {
    [bindingsSerializerSym] (bindings) {
      return { hello: 'world', ...bindings }
    }
  }
})

child.info('test')
```
The produced logs are:
```
{"log":{"level":"info"},"time":1582036600545,"process":{"pid":38765},"hostname":"Skadi.local","msg":"hello world"}
{"log":{"level":"info"},"time":1582036600546,"process":{"pid":38765},"hostname":"Skadi.local","process":{"pid":38765},"hostname":"Skadi.local","hello":"world","foo":"bar","nested":{"object":true},"msg":"test"}
```

What I like about this approach is that it's not "yet another option", but it's a feature and Pino itself is the first customer, also, the developer experience is better than configuring by hand the pieces of the JSON that needs to be assembled, like is happening for `time`.

The pull request is not yet ready to be merged, as I wanted to have an early review of the approach and see if you like it.
If so, what I think should still be done is:
- add a serializer also for timestamps
- add a serializer also for the message
- remove the multitude of options for customizing levels/timestamps/message, as the serializer would be the only way for changing these values
- provide more prepackaged serializers in `pino-std-serializers`

Following, some benchmarks:
```
=== basic.bench.js ===
# next branch
benchPino*10000: 194.908ms
benchPinoExtreme*10000: 72.674ms

# this pr
benchPino*10000: 188.696ms
benchPinoExtreme*10000: 71.806ms

=== child.bench.js ===
# next branch
benchPinoChild*10000: 217.076ms
benchPinoExtremeChild*10000: 80.304ms

# this pr
benchPinoChild*10000: 212.370ms
benchPinoExtremeChild*10000: 88.715ms
```
If this work will get accepted I want to do a follow up where users can configure the order of the keys of the JSON, it shouldn't be a problem to support it and I don't see it introducing too much overhead.


lib/levels.js
```diff
@@ -49,13 +49,14 @@ const initialLsCache = Object.keys(nums).reduce((o, k) => {
...
-  instance[lsCacheSym] = Object.keys(instance.levels.labels).reduce((o, k) => {
-    o[k] = instance[useLevelLabelsSym]
-      ? `{"${levelName}":"${instance.levels.labels[k]}"`
-      : flatstr(`{"${levelName}":` + Number(k))
-    return o
-  }, Object.assign({}, instance[lsCacheSym]))
+  const serializer = instance[serializersSym][levelSerializerSym]
+  const { labels } = instance.levels
+  const cache = {}
+  Object.keys(labels).forEach(number => {
```
mcollina: I would just drop the `forEach` and just use a for loop

lib/proto.js
```diff
@@ -66,9 +67,8 @@ Object.setPrototypeOf(prototype, EventEmitter.prototype)
 module.exports = prototype
 
 function child (bindings) {
-  const { level } = this
   const serializers = this[serializersSym]
-  const chindings = asChindings(this, bindings)
+  serializers[bindingsSerializerSym] = bindings => bindings
```
mcollina: Can we use a default for this in the prototype chain so we don't have to generate a new closure every time?

davidmarkclements:
great job - I think this is a best-effort solution given all constraints.

We need benchmarks for these new serializers (on vs off) 

Rather than loading from lib symbols, we should use Symbol.for symbols, as we do https://github.com/pinojs/pino/blob/master/lib/symbols.js#L33-L35

So on the same naming scheme as we have currently it would be:

```js
const pino = require('pino')({
  serializers: {
    [Symbol.for('pino.bindings')] (bindings) {
      return {
        process: { pid: bindings.pid },
        hostname: bindings.hostname
      }
    },
    [Symbol.for('pino.level')] (label, number) {
      return {
        log: { level: label }
      }
    }
  }
})
```

lib/levels.js
```diff
@@ -49,13 +49,14 @@ const initialLsCache = Object.keys(nums).reduce((o, k) => {
...
+    const level = serializer(labels[number], Number(number))
+    cache[number] = JSON.stringify(level).slice(0, -1)
```
davidmarkclements: there's a much faster escaping function here https://github.com/pinojs/pino/blob/master/lib/tools.js#L48

use that instead of JSON.stringify

lib/symbols.js
```diff
@@ -33,6 +33,8 @@ const wildcardFirstSym = Symbol('pino.wildcardFirst')
 const serializersSym = Symbol.for('pino.serializers')
 const wildcardGsym = Symbol.for('pino.*')
 const needsMetadataGsym = Symbol.for('pino.metadata')
+const bindingsSerializerSym = Symbol.for('pino.serializers.bindings')
+const levelSerializerSym = Symbol.for('pino.serializers.level')
```
davidmarkclements: So the problem with this naming scheme is it's incongruous. 

`Symbol.for('pino.serializers')` is used to get the serializers for a given logger instance (http://getpino.io/#/docs/api?id=loggersymbolfor39pinoserializers39), when we configure serializer via opts we don't use a symbol.

`Symbol.for('pino.*')` is a global serializer, but it it's NOT pino.serializers.* 

So either these should be called 

* `Symbol.for('pino.bindings')` 
* `Symbol.for('pino.level')` 

OR we need to deprecate and break: 

* `Symbol.for('pino.*')` -> `Symbol.for('pino.serializers.*')`
* `Symbol.for('pino.serializers')` -> `Symbol.for('pino.serializers-acesssor')`  (?)

One other alternative is to use do associate the dollar sign `$` with serializer setters/config symbols:

* `Symbol.for('pino.*')` -> `Symbol.for('pino.$.*')`
* `Symbol.for('pino.serializers')` stays the same 
* `Symbol.for('pino.$.bindings')`
* `Symbol.for('pino.$.level')`

For now I'd go with the easy option:

* `Symbol.for('pino.bindings')` 
* `Symbol.for('pino.level')` 

And then we'll batch assess Symbol namespaces in pino next

davidmarkclements: Another fundamental question is, should these be symbols. `pino.*` was deliberately provided as a symbol because we felt it could be a footgun and to show that it's distinct from normal serializers, in that it's one level above normal serializers.  These new serializers are even a level above that again. 

So we have three levels

* log base serializers (this PR)
* entire log object serializer  (`Symbol.for('pino.*'`)
* msg object key serializers (the original) 

I think we should have this API instead /cc @mcollina @jsumners 

```js
const pino = require('./')({
  superSerializers: {
    bindings (bindings) {
      return {
        process: { pid: bindings.pid },
        hostname: bindings.hostname
      }
    },
    level (label, number) {
      return {
        log: { level: label }
      }
    },
   '*' (obj) { // replaces Symbol.for('pino.*')
     obj.foo = 'whatever' 
     return obj
   }
  }
})
```

davidmarkclements: If the benchmarks for these super serializers are slow then it should instead be:

```js
const pino = require('./')({
  [Symbol.for('pino.superSerializers')]: {
    bindings (bindings) {
      return {
        process: { pid: bindings.pid },
        hostname: bindings.hostname
      }
    },
    level (label, number) {
      return {
        log: { level: label }
      }
    },
   '*' (obj) { // replaces Symbol.for('pino.*')
     obj.foo = 'whatever' 
     return obj
   }
  }
})
```

jsumners: I'm not really a fan of exposing a client API through usage of symbols unless it is necessary to keep a serialized object clean. I think they should be used internally to give strong indication that anyone using them _must_ expect breakage at some point. Thus, we wouldn't have to issue a new major if we decide to change symbol descriptors.

Which is to say, I prefer https://github.com/pinojs/pino/pull/775#discussion_r381267993

lib/levels.js
delvedor: I can't as the user here is retuning a js object, `asString` is built for escaping strings, isn't it?

lib/levels.js
```diff
@@ -52,10 +52,10 @@ function genLsCache (instance) {
   const serializer = instance[serializersSym][levelSerializerSym]
   const { labels } = instance.levels
   const cache = {}
-  Object.keys(labels).forEach(number => {
-    const level = serializer(labels[number], Number(number))
-    cache[number] = JSON.stringify(level).slice(0, -1)
-  })
+  for (const label in labels) {
```
jsumners: Should a `hasOwnProperty` check be used here?

delvedor:
> How is the bindings serializer called for the child logger? 

It is called when you create a child with the new bindings, the parent's bindings are already part of the instance, and they will not be passed to the serializer.
There is a small "issue" about this tho, I need to override the bindings serializer every time a child is created, otherwise, the new bindings will not be part of the log, because the parent's serializer is still being used.
If the user is not overriding it, I need to do it by hand with the `bindings => bindings`.
Now, when I override the bindings serializer for the child, also the parent gets updated (unless the user has provided their own implementation), it should not be a problem because the bindings serializer is called only once per instance synchronously on creation.

> Is it call for only the new parts added by the child logger?

Yes, see above.

> Is this enough to implement the elastic format?

Yes and no, I still need to work on the time and message serializer, and finally, a key order would be great.

> Can you please refactor the tests so all the cases for the old options are covered anyway?

Done.

lib/symbols.js
delvedor: For me either is fine, probably I would choose a different name, such as `globalSerializers`.
Still, there is the "issue" when creating a child, where I need to override the parent's bindings serializer.

delvedor: We could even call it `fomatters`.
```js
const pino = require('./')({
  formatters: {
    bindings (bindings) {
      return {
        process: { pid: bindings.pid },
        hostname: bindings.hostname
      }
    },
    level (label, number) {
      return {
        log: { level: label }
      }
    },
   log (obj) {
     obj.foo = 'whatever' 
     return obj
   }
  }
})
```

davidmarkclements: I'm happy with `formatters` :+1:

davidmarkclements: I'm not sure that `log` is the right name or not

lib/levels.js
davidmarkclements: don't we control the `labels` object? If the user can (legitimately, as per API) overwrite the labels object then yes. If they can't then it's not needed

lib/symbols.js
jsumners: https://github.com/pinojs/pino/blob/d237668d71e4e62224269eeef0b1b6dc4d668dbd/lib/levels.js#L119-L136

lib/levels.js
jsumners: https://github.com/pinojs/pino/blob/d237668d71e4e62224269eeef0b1b6dc4d668dbd/lib/levels.js#L119-L136

delvedor: I've added a stricter check in https://github.com/pinojs/pino/pull/775/commits/489a6cf95573c900651518e078143122aa456d05, but with or without it there are no test that will fail.
How can we test it?

delvedor:
I've updated the code and introduced the `formatters` top-level options, let me know what you think!
I'll add more tests soon and start working on message and time formatters as well.

lib/proto.js
```diff
@@ -91,13 +92,31 @@ function child (bindings) {
       instance[serializersSym][bks] = bindings.serializers[bks]
     }
   } else instance[serializersSym] = serializers
+  if (bindings.hasOwnProperty('formatters')) {
+    instance[formattersSym] = Object.create(null)
```
davidmarkclements: there's no need, in my mind, to create a null proto object.

lib/levels.js
davidmarkclements: by putting enumerable properties on the prototype of the labels object

if you can't do that via the API then the check isn't necessary

lib/proto.js
```diff
@@ -91,13 +92,31 @@ function child (bindings) {
...
+    for (k in formatters) {
```
davidmarkclements: since formatters is supplied by user, we can't rely on for in without hasOwnProperty. 

formatters only has three properties, we should optimize for speed by just looking for those properties:

```
instance[formattersSym] = {
  level: formatters.level,
  bindings: formatters.bindings 
}
```

this is both less code, and faster

pino.js
```diff
@@ -48,6 +50,14 @@ const defaultOptions = {
   serializers: Object.assign(Object.create(null), {
     err: defaultErrorSerializer
   }),
+  formatters: Object.assign(Object.create(null), {
```
davidmarkclements: I don't know when we started creating null objects for default options but it's completely unnecessary

mcollina: They are way faster.

davidmarkclements:
API looking good, still need equiv for `Symbol.for('pino.*')` 

Only thing now is to make sure it's fast, esp child logger init times

pino.js
davidmarkclements: wtf. I thought they were slower

and also .. wtf.

lib/levels.js
davidmarkclements: oh sorry I thought level was a number or string

if it's an object why not just access the property from the object?

lib/proto.js
jsumners: I despise them.

delvedor: I'm doing the same in the default options, I thought it was your preference because you are doing it in the serializers as well.
https://github.com/delvedor/pino/blob/dafb6940f7bb6dbc31afff247987369484616f17/pino.js#L50-L60

I'll remove it :)

lib/proto.js
delvedor: Good point!

test/levels.test.js
```diff
@@ -326,8 +348,11 @@ test('produces labels for custom levels', async ({ is }) => {
 test('setting changeLevelName does not affect labels when told to', async ({ is }) => {
   const instance = pino(
     {
-      useLevelLabels: true,
-      changeLevelName: 'priority'
```
mcollina: Can we keep supporting the `changeLevelName` in the default formatter with a deprecation warning?

lib/proto.js
```diff
@@ -91,13 +91,26 @@ function child (bindings) {
...
+      level: bindings.formatters.level || formatters.level,
+      bindings: bindings.formatters.bindings || resetChildingsFormatter,
+      log: bindings.formatters.log || formatters.log
+    }
+  } else {
+    instance[formattersSym] = {
+      level: formatters.level,
+      bindings: resetChildingsFormatter,
+      log: formatters.log
+    }
```
mcollina: Instead of creating two objects at two separate lines, could you wrap them in a single factory? It would help in them having the same shape.

test/levels.test.js
delvedor: I'll see what I can do, given that Pino is the first customer of the new formatter option, supporting the old way for changing the levels and labels is not straightforward, I fear.

lib/symbols.js
```diff
@@ -31,7 +31,7 @@ const wildcardFirstSym = Symbol('pino.wildcardFirst')
 // public symbols, no need to use the same pino
 // version for these
 const serializersSym = Symbol.for('pino.serializers')
-const wildcardGsym = Symbol.for('pino.*')
```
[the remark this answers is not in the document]
delvedor: Oks, How do you signal deprecations with Pino?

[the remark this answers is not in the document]
mcollina: I would use https://nodejs.org/api/util.html#util_util_deprecate_fn_msg_code.

See https://github.com/pinojs/pino/commit/bcaacba70c232986253c4f1888ca64a695f36749#diff-04e6fb630166e65e0231f8c63a731aa0L425.

delvedor:
Not sure why the test are failing... it works on my machine™!
I'll fix them tomorrow, please still take a look :)

lib/tools.js
```diff
@@ -351,6 +358,20 @@ function stringify (obj) {
   }
 }
 
+function Formatters (level, bindings, log) {
```
jsumners: Does this need to be a constructor function?

lib/tools.js
```diff
@@ -351,6 +358,20 @@ function stringify (obj) {
...
+  this.level = level
+  this.bindings = bindings
+  this.log = log
+}
+
+class DeprecationWarning extends Error {
```
jsumners: I'd prefer to avoid `class`. What is wrong with a regular error?

jsumners:
@delvedor pull in upstream changes from `next` and rebase. Your errors seem to be related to a PR merged yesterday.

lib/tools.js
davidmarkclements: agreed - unless there's some measurably significant reason not to instead return an object then let's just return an object

lib/tools.js
davidmarkclements: agreed this is unnecessary noise - because these passed to `process.emitWarning`, which just logs the error message. You don't even need an error for `process.emitWarning` using strings is fine

pino.js
```diff
@@ -76,9 +84,40 @@ function pino (...args) {
...
   } = opts
 
+  const allFormatters = new Formatters(
+    formatters.level,
+    formatters.bindings,
+    formatters.log
+  )
+
+  if (useLevelLabels && !changeLevelName) {
+    process.emitWarning(new DeprecationWarning('useLevelLabels is deprecated, use the formatters.level option instead', 'PINODEP001'))
```
davidmarkclements: just pass strings to emitWarning - theres no benefit to using error objects

jsumners:
This is the test that is failing -- https://github.com/delvedor/pino/blob/9b336b9a2e3b21c824c65d2e286c56faf0d790dd/test/levels.test.js#L431-L444

I cannot pull down your branch and step through it until tonight at the earliest.

jsumners:
The issue is that on `levels.test.js:443` the value of `expected.msg` is `null` but the `result` object does not have a `msg` property at all. Which is to say, the serializer is regressing the work done in #711.

lib/tools.js
```diff
@@ -89,8 +92,11 @@ function asJson (obj, num, time) {
 
   var value
   var notHasOwnProperty = obj.hasOwnProperty === undefined
-  if (serializers[wildcardGsym]) {
-    obj = serializers[wildcardGsym](obj)
+  if (formatters.log) {
+    obj = formatters.log(obj)
+  }
+  if (msg) {
```
jsumners: This is where the error is introduced. Since `msg === null` it is not getting logged as `"msg": null` (as #711 implements).

delvedor:
There are a couple of tests that I don't know how to cover:
```js
function genLsCache (instance) {
  const formatter = instance[formattersSym].level
  const { labels } = instance.levels
  const cache = {}
  for (const label in labels) {
    if (!labels.hasOwnProperty(label)) continue // this line is not covered
    ....
```
```js
function asChindings (instance, bindings) {
  if (!bindings) { // this line is not covered
    throw Error('missing bindings for child Pino')
  }
```

Any suggestion?

delvedor:
Given the scope of this pr, I'm not sure it made sense to backport https://github.com/pinojs/pino/pull/772 to `next`, now in my local branch, I have tons of conflicts for a feature that would have been deprecated in any case.

jsumners:
The deprecation will be for the duration of the v6 release. So it needed to be kept. 

I did the master to next merge because I expected difficulty syncing this PR with master.

jsumners:
The `hasOwnProperty` test... if you cannot get prototype chain properties into the test then we don’t need the own check.

davidmarkclements:
> The `hasOwnProperty` test... if you cannot get prototype chain properties into the test then we don’t need the own check.

this is hands down the best thing about writing tests, knowing for sure that you can remove code

delvedor:
Hello! I've finished the merge and fixed all the conflicts, please take a look, since I had to rewrite a significant part of the code.

lib/proto.js
```diff
@@ -91,13 +92,26 @@ function child (bindings) {
       instance[serializersSym][bks] = bindings.serializers[bks]
     }
   } else instance[serializersSym] = serializers
+  if (bindings.hasOwnProperty('formatters')) {
+    instance[formattersSym] = buildFormatters(
+      bindings.formatters.level || formatters.level,
+      bindings.formatters.bindings || resetChildingsFormatter,
+      bindings.formatters.log || formatters.log
```
mcollina: please extract the `bindings.formatters` object into a local `const`, it will shelve some ms.

jsumners:
ping: @delvedor

jsumners:
@delvedor please pull in the latest `next` branch. It adds a GH Action to run benchmarks. See #786

delvedor:
I've addressed all the comments, I'll work on the benchmarks soon.
Regarding line 138 of `lib/tools.js`,I'm not sure how can cover that line, nor how the coverage was present before. Maybe with this pr that path no longer exists?

delvedor:
I've added a benchmark for the formatters, please take a look if it is was you were expecting.

test/basic.test.js
```diff
@@ -359,8 +359,8 @@ test('set the base to null and use a serializer', async ({ is, same }) => {
   const stream = sink()
   const instance = pino({
     base: null,
-    serializers: {
-      [Symbol.for('pino.*')]: (input) => {
+    formatters: {
+      log (input) {
```
mcollina: This test says "use a serializer". I think we need to update the test description

mcollina:
Also, there seems to be a drop in coverage.

delvedor:
I'll work on the docs :)

This is the log of the benchmarks with and without the formatters:
```
benchPinoNoFormatters*10000: 227.418ms
benchPinoFormatters*10000: 260.588ms
benchPinoNoFormatters*10000: 246.946ms
benchPinoFormatters*10000: 252.969ms
```

delvedor:
I've updated the documentation, I'm fairly sure that there is still work to do, please give it a look and provide feedback on how you would like to phrase this.

docs/api.md
```diff
@@ -150,13 +195,6 @@ matching the exact key of a serializer will be serialized using the defined seri
 
 * See [pino.stdSerializers](#pino-stdserializers)
 
-##### `serializers[Symbol.for('pino.*')]` (Function)
-
-Default: `undefined`
-
-The `serializers` object may contain a key which is the global symbol: `Symbol.for('pino.*')`.
-This will act upon the complete log object rather than corresponding to a particular key.
-
```
mcollina: I think this is missing the `DEPRECATED` block

docs/api.md
```diff
@@ -138,6 +138,51 @@ If an object is supplied, three options can be specified:
 * See the [redaction ⇗](/docs/redaction.md) documentation.
 * See [fast-redact#caveat ⇗](http://github.com/davidmarkclements/fast-redact#caveat)
 
+<a id=opt-formatters></a>
+#### `formatters` (Object)
+
+An object containing functions for formatting the shape of the logs.
```
jsumners:
```suggestion
An object containing functions for formatting the shape of the log lines.
```

docs/api.md
```diff
@@ -138,6 +138,51 @@ If an object is supplied, three options can be specified:
...
+These functions should return an JSONifiable object and they
```
jsumners:
```suggestion
These functions should return a JSONifiable object and
```

docs/api.md
```diff
@@ -138,6 +138,51 @@ If an object is supplied, three options can be specified:
...
+should never throw. You can use this functions to shape the final JSON log in the way you need. For example you can use it for changing the level key name or enrich the default bindings.
```
jsumners:
```suggestion
should never throw. These functions allow for full customization of
the resulting log lines. For example, they can be used to change
the level key name or to enrich the default metadata.
```

docs/api.md
```diff
@@ -138,6 +138,51 @@ If an object is supplied, three options can be specified:
...
+
+##### `level`
+
+Changes the shape of the log level. The default shape is `{ level: number }`.
+The function takes two arguments, the label of the level (eg `'info'`) and the numeric value (eg: `30`).
```
jsumners:
```suggestion
The function takes two arguments, the label of the level (e.g. `'info'`)
and the numeric value (e.g. `30`).
```

docs/api.md
```diff
@@ -138,6 +138,51 @@ If an object is supplied, three options can be specified:
...
+  level (label, number) {
+    return { level: number }
+  }
+}
+```
+
+##### `bindings`
+
+Changes the shape of the bindings. The default shape is `{ pid, hostname }`.
+The function takes a single argument which is the bindings object. It will be called every time you create a child logger.
```
jsumners:
```suggestion
The function takes a single argument, the bindings object. It will
be called every time a child logger is created.
```

docs/api.md
```diff
@@ -138,6 +138,51 @@ If an object is supplied, three options can be specified:
...
+const formatters = {
+  bindings (bindings) {
+    return { pid: bindings.pid, hostname: bindings.hostname }
+  }
+}
+```
+
+##### `log`
+
+Changes the shape of the log object. This function will be called every time one of the log methods (such as `.info`) is called with all the arguments passed to the log method but the message. By default it does not change the shape of the log object.
```
jsumners:
```suggestion
Changes the shape of the log object. This function will be called every time
one of the log methods (such as `.info`) is called. All arguments passed to the
log method, except the message, will be pass to this function. By default it does
not change the shape of the log object.
```

delvedor:
Hey folks, I've created a https://github.com/elastic/ecs-logging-js/pull/11 for adding the support of [ecs](https://www.elastic.co/guide/en/ecs/current/index.html) to Pino by using this new feature, everything looks fine! 🎉

mcollina:
Can you amend the title and description of this PR to name "formatters"? Thanks

delvedor:
@mcollina I've updated the title, I'll keep the description as is, otherwise, all the discussion below will lose sense. What do you think?

mcollina:
ok for me, thanks!

jsumners:
@delvedor if you can just resolve the current conflict then we can get this merged and publish an RC. It's an easy resolution but I don't understand the GitHub UI for doing it.

mcollina:
> @delvedor if you can just resolve the current conflict then we can get this merged and publish an RC. It's an easy resolution but I don't understand the GitHub UI for doing it.

I've done it.

jsumners:
> > @delvedor if you can just resolve the current conflict then we can get this merged and publish an RC. It's an easy resolution but I don't understand the GitHub UI for doing it.
> 
> I've done it.

🎉

PR #780, opened 2020-02-28
Fixed stream metadata loss during prettification

nlapshin:
For streams with Symbol('metadata') true flag, pino metadata was lost during prettification.

RE: https://github.com/pinojs/pino-multi-stream/issues/31

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -173,12 +173,24 @@ function prettifierMetaWrapper (pretty, dest) {
     lastMsg: null,
     lastObj: null,
     lastLogger: null,
+    get destStreamContext () {
+      if (dest[needsMetadataGsym] === true) {
+        return Object.assign(dest, {
+          lastLevel: this.lastLevel,
+          lastMsg: this.lastMsg,
+          lastObj: this.lastObj,
+          lastTime: this.lastTime,
+          lastLogger: this.lastLogger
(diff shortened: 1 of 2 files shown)


lib/tools.js
```diff
@@ -173,12 +173,24 @@ function prettifierMetaWrapper (pretty, dest) {
...
+      }
+      return dest
+    },
     flushSync () {
       if (warned) {
         return
       }
       warned = true
-      dest.write(pretty(Object.assign({
+      dest.write.call(this.destStreamContext, pretty(Object.assign({
```
mcollina: Copying streams metadata might cause _significant issues_. I would just attach the properties to the stream itself, mutating it.

nlapshin: Fixed it

mcollina: I think you can do something like:

```js
function setMetadataProps (dest, that) {
      if (dest[needsMetadataGsym] === true) {
        dest.lastLevel = that.lastLevel
        dest.lastMsg = that.lastMsg
        dest.lastObj = that.lastObj
        dest.lastTime = that.lastTime
        dest.lastLogger = that.lastLogger
      }
}
```

and then call in both places.

nlapshin: Created setMetadataProps function

PR #781, opened 2020-02-29
Merge `master` into `next`

jsumners:


jsumners:
That merged a lot cleaner than I thought it would.

PR #782, opened 2020-03-03
add pino-pg transport

Xstoudi:


mcollina:
Would you mind to add a unit test for your module?

Xstoudi:
Sure, it's planned for 14th of March.
I'll leave a comment here when it's done.

Xstoudi:
Sorry for the wait, I pushed test and changed the way the transports work. Check it out and let met know if you thing more changes are needed.

Xstoudi:
I think this one is mergeable now that the issue is solved :)

PR #786, opened 2020-03-10
Add GitHub action to benchmark PRs

jsumners:
This adds a GitHub action that will run our benchmark suite against the incoming PR's code and the branch that PR targets. The results will be uploaded as artifacts on the action results. These artifacts are kept around by GH for a period of 40 days. We should also be able to review the action logs to see the result after that time.


jsumners:
image

PR #791, opened 2020-03-17
Avoid the term "extreme", there are just sync and async destinations.

### LINKED ISSUE #785: Call pino.destination as pino.sync and pino.extreme as pino.async
Operationally they provide two very different approaches and I think we would be best to not call `extreme()` as a drastic measure but rather `async()` because that reflect what it is without making judgements.

mcollina:
Fixes #785

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -575,7 +575,9 @@ sync: false })`.
 
 This is an asynchronous, fire and forget, operation.
 
-The use case is primarily for asynchronous logging, which may hold log lines while others are being written. The `logger.flush` method can be used to flush the logs
+The use case is primarily for asynchronous logging, which may buffer
+log lines while others are being written. The `logger.flush` method can be
+used to flush the logs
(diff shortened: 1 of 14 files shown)


docs/api.md
```diff
@@ -572,19 +570,18 @@ console.log(anotherChild.bindings())
...
 
-Flushes the content of the buffer when using a `pino.extreme` destination.
+Flushes the content of the buffer when using `pino.destination({
+sync: false })`.
 
 This is an asynchronous, fire and forget, operation.
 
-The use case is primarily for Extreme mode logging, which may hold up to
-4KiB of logs. The `logger.flush` method can be used to flush the logs
+The use case is primarily for asynchronous logging, which may hold log lines while others are being written. The `logger.flush` method can be used to flush the logs
```
jsumners:
```suggestion
The use case is primarily for asynchronous logging, which may buffer
log lines while others are being written. The `logger.flush` method can be
used to flush the logs
```

Issue #793, opened 2020-03-20
Serialization is adding empty object to string


jsumners:
```js
'use strict'

const pino = require('pino')
const log = pino({
  redact: {
    paths: ['password'],
    censor: '[REDACTED]'
  }
})

log.info('help', {})
```

The result will be:

```
{"level":30,"time":1584733267938,"pid":58479,"hostname":"mini2018","msg":"help {}"}
```

Notice that the `msg` key has the value `help {}`. It _should_ have the value `help`.

I have trace it down to this line of `quick-format-unescaped` -- https://github.com/davidmarkclements/quick-format-unescaped/blob/9e5147d77fc2cc1b7bc6ad391647e75b6792afec/index.js#L100

I'm not sure why it is happening though. It shouldn't be trying to inject any formatted values because there isn't a format token in `log.info('help', {})`.

jsumners:
This isn't solved until we get the new dependency integrated here.

PR #795, opened 2020-03-23
Remove implicit appending of objects to message strings

jsumners:
This fixes issue #793. 

We will need to add a major call to notice for this in the release notes for v6. Something like:

> **Implicit formatting removed**
> Previously, Pino emulated Bunyan's behavior when interpreting logs like:
> 1. `log.info('a message', { an: 'object'})`
> 1. `log.info('a', 'silly', 'message')`
>
> The old behavior would result in 1 yielding `"msg":"a message {\"an\":\"object\"}"` and 2 yielding
> `"msg":"a silly message"`. As of this release, 1 will yield `"msg":"a message"` and 2 will yield `"msg":"a"`.
>
> To get the same results in Pino v6 as in previous releases actual format identifiers should be provided, e.g.:
> 1. `log.info('a message %j', { an: 'object' })`
> 1. `log.info('a %s %s', 'silly', 'message')`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -448,9 +448,8 @@ then be interpolated accordingly.
 #### `...interpolationValues` (Any)
 
 All arguments supplied after `message` are serialized and interpolated according
-to any supplied printf-style placeholders (`%s`, `%d`, `%o`|`%O`|`%j`)
-or else concatenated together with the `message` string to form the final
-output `msg` value for the JSON log line.
+to any supplied printf-style placeholders (`%s`, `%d`, `%o`|`%O`|`%j`) to form
+the final output `msg` value for the JSON log line.
 
 ```js
 logger.info('hello', 'world')
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -30,6 +30,7 @@ function noop () {}
 function genLog (z) {
   return function LOG (o, ...n) {
     if (typeof o === 'object') {
+      var msg = o
       if (o !== null) {
         if (o.method && o.headers && o.socket) {
           o = mapHttpRequest(o)
@@ -38,8 +39,14 @@ function genLog (z) {
         }
       }
       if (this[nestedKeySym]) o = { [this[nestedKeySym]]: o }
-      const formatParams = (o === null && n.length === 0) ? [null] : n
-      this[writeSym](o, format(null, formatParams, this[formatOptsSym]), z)
+      var formatParams
+      if (msg === null && n.length === 0) {
+        formatParams = [null]
+      } else {
+        msg = n.shift()
+        formatParams = n
+      }
+      this[writeSym](o, format(msg, formatParams, this[formatOptsSym]), z)
     } else {
       this[writeSym](null, format(o, n, this[formatOptsSym]), z)
     }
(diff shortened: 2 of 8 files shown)


jsumners:
Looks like benchmarks improved 🤔

mcollina:
Is there any docs to change?

jsumners:
@mcollina good call. Docs updated.

doubliez:
Is there any way to restore previous behavior when upgrading to Pino v6? We have a big codebase and make quite extensive use of the fact that extra arguments would be automatically stringified and concatenated together, so it would be painful and error-prone to go through every log statement and add format identifiers.

jsumners:
The result of #803, if implemented and accepted, could solve your issue.

jsumners:
@doubliez I just released v6.1.0 which will allow you to do:

```js
const pinoOptions = {
  hooks: { logMethod }
}

function logMethod (args, method) {
  if (args.length === 2) {
    args[0] = `${args[0]} %j`
  }
  method.apply(this, args)
}
```

doubliez:
@jsumners Great, thanks, that should solve my issue.

tomaszlukasik:
Hi, you forgot to updated first to examples in here http://getpino.io/#/docs/api?id=interpolationvalues-any

jsumners:
Oops. @tomaszlukasik would you like to submit a fix?

tomaszlukasik:
> Oops. @tomaszlukasik would you like to submit a fix?

I was about to do that, but @mcollina was faster 😞  😉

guillenotfound:
So this is now supported by default: https://github.com/pinojs/pino/blob/8786e3acbb0f50eeed13d4d599b4f25b0fa43730/docs/api.md#logmethod

We no longer need to change order to use `logger.info(string, obj)`.

thw0rted:
I'm late to the party updating to Pino 6.x so I just discovered this.  I'm not sure where else to ask, but: would it make sense to add *some* channel to complain when arguments are passed to a log method that don't actually get used?

Let's say I have code that started as `console.info("the client did a thing", clientDetails)`, then I migrated to Pino.  It worked under 5.x using `log.info("the client did a thing", clientDetails)`.  Now, I decided to handle the changes in 6.0 by updating all those calls to the (arguably more useful) `log.info({msg: "the client did a thing", clientDetails})`.  Great, now I get the details-object as a named field in the log JSON.  But maybe I have a thousand such calls, and I missed a few when I updated.  Now, I'm quietly logging the message but totally discarding the details, and I won't notice until I check the logs at some future point.

Put another way: if I didn't want the logger to do *something* with that second argument, I wouldn't have passed it.  Could Pino maybe tell me when an argument is ignored?

mcollina:
That'd be expensive to do in all cases, but it's something you can easily monkey-patch for while developing/migrating.

jsumners:
Use https://github.com/pinojs/pino/blob/40fa7f6e7aed95a9e916787a047a9f56627e33c7/docs/api.md#logmethod

thw0rted:
I think I see what you mean.  So, I could use the `hooks` option as in the comment above to check when one of the log methods is called with more arguments than formatting placeholders, and, uh, do something (probably complain on the console, at least to start).

PR #799, opened 2020-03-25
Add a basic LTS doc

jsumners:
This PR adds a very simple document describing our LTS policy. In short, it says we only support LTS releases of core and that we will synchronize with core as our schedules allow.


jsumners:
Is it not?

mcollina:
We are dropping Node.js 6.x and 8.x in Pino v6.x, not in Pino v5.

jsumners:
Please make a suggestion to the doc.

jsumners:
@mcollina updated. PTAL

PR #802, opened 2020-03-30
Add "silent" property with "noop" value to the pino instance

baterson:
Hello this is a fix from issue raised in `pino-http` https://github.com/pinojs/pino-http/issues/79  
I'm not sure if this naive implementation is correct. Maybe we need to add `silent` to the levels list and update all of the methods which are related to the log levels. But if you ok with this implementation, I'll add tests to this PR.


jsumners:
I'm not sure about this. We have always documented `silent` as a level value distinctly _not_ a logger method -- https://github.com/pinojs/pino/blob/529906ed868ec362d66b228dbc10bd201b766202/docs/api.md#loggerlevel-string-gettersetter

Changing that would mean a new semver major change, IMHO. I also think it would break from the standardized interface that is implemented, i.e. the log4j logger interface.

I think the better fix is at -- 
https://github.com/pinojs/pino-http/blob/eacc870e6d9d8e24608d27bd7bcda528ae162d47/logger.js#L53-L69

baterson:
> Would you mind to add a bunch of unit tests?

What about the comment above? We can add check for "silent" level there https://github.com/pinojs/pino-http/blob/eacc870e6d9d8e24608d27bd7bcda528ae162d47/logger.js#L53-L69 
As @jsumners suggested.  
But if you all agree on this change, I'll going to add tests for this PR.

mcollina:
I'm not necessarily sure about this fix because it's lacking unit tests. Specifically we need to check how it will work with child loggers, and if it interferers with changing the log level to "silent" and back at runtime.

baterson:
@mcollina I added tests. Please tell me if I missed something.

jsumners:
At a minimum, the docs would need to be updated. And I'd really like to know @davidmarkclements's thoughts.

baterson:
@jsumners  I updated docs API section.

mcollina:
Ouch, I thought this could land.

mcollina:
@jsumners should I revert?

jsumners:
@mcollina I was really hoping @davidmarkclements would add his opinion about whether or not this is a semver major change. I am undecided. 

Regardless, if it lands as minor and is really major, it wouldn't be the worst major thing to ever land as a minor. It could just be a bit surprising to some people.

davidmarkclements:
Hey sorry for my .. err.. silence 🤭 on this 

To me this is a minor and I’m also fine with it 👍

jsumners:
giphy

Issue #804, opened 2020-03-31
Pretty print does not respect formatters.log


giannisp:
I have found two issues while using `prettyPrint: true` and `formatters.log`.
- First one is straight-forward, pretty print does not apply `formatters.log` function (like it does with serializers and redact functions).
- Second one seems more complicated. While using formatters.log and returning a modified log message (even by doing something like `{ ...log }`), pretty print doesn't print the actual log message, it only prints the log object.  

I cloned the repo locally, and could easily fix the first one, however I was stuck on the second one as I am not familiar with all repo mechanics.

Here is a test that can illustrate both issues:

**test/fixtures/pretty/formatters.js**
```js
global.process = { __proto__: process, pid: 123456 }
Date.now = function () { return 1459875739796 }
require('os').hostname = function () { return 'abcdefghijklmnopqr' }
var pino = require(require.resolve('./../../../'))
var log = pino({
  prettyPrint: true,
  formatters: {
    log (obj) {
      // return obj
      return { foo: 'formatted_' + obj.foo }
    }
  }
})
log.info({ foo: 'bar' }, 'h')
```

**test/pretty.test.js**
```js
test('applies formatters', async ({ is, isNot }) => {
  var actual = ''
  const child = execa(process.argv[0], [join(__dirname, 'fixtures', 'pretty', 'formatters.js')])

  child.stdout.pipe(writer((s, enc, cb) => {
    actual += s
    cb()
  }))
  await once(child, 'close')
  isNot(strip(actual).match(/\(123456 on abcdefghijklmnopqr\): h/), null)
  isNot(strip(actual).match(/foo: "formatted_bar"/), null)
})
```

mcollina:
I'd recommend you to send a PR with a a failing test, I'm not following.

I'm not sure that `prettyPrint` can understand the `formatters` at all. It would probably need some change as well. Note that prettyPrint lives in its own module https://github.com/pinojs/pino-pretty

@delvedor  you implemented the formatters, wdyt?

giannisp:
Yes, formatters are not being evaluated at `prettyPrint: true` I think.  
They should be handled on this function if I understand the codebase correctly:
https://github.com/pinojs/pino/blob/master/lib/tools.js#L242

On my initial post ^ I have attached a new test, if that helps you to try it out locally.
Plus check my second bullet point.

mcollina:
I would recommend sending a Pull Request.

PR #809, opened 2020-04-02
Support formatters.log on pretty print

giannisp:
This is in reference to issue #804.  

I have made an attempt to add support for `formatters.log` while using `prettyPrint: true`, similarly to how `serializers` and `redact` functions are supported.  
I have added a new test with two asserts.
The second assert, which is related to the format action is passing the test.
The first assert however fails, as the log message is not included on the output. This is not a side-effect of this PR, this issue exists even on the current `master`, ie without my submitted modification on tools. 

While returning the original `obj` on `formatters.log`, the log message is present on the pretty print output. Thus with my limited knowledge of the repo, I can only assume that it might be a scope/reference issue with the log object at some other place.


test/fixtures/pretty/formatters.js
```diff
@@ -0,0 +1,13 @@
+global.process = { __proto__: process, pid: 123456 }
+Date.now = function () { return 1459875739796 }
+require('os').hostname = function () { return 'abcdefghijklmnopqr' }
+var pino = require(require.resolve('./../../../'))
+var log = pino({
+  prettyPrint: true,
+  formatters: {
+    log (obj) {
```
mcollina, line 8: I _think_ there is a msg property here with the message. You’d have to pass it through. Note that prettyPrint does not know anything about the new format, it just look for a given property.

mcollina:
cc @delvedor

test/fixtures/pretty/formatters.js
giannisp: Thanks for your suggestion! The msg property wasn't available on that context, but I have pushed a simple change to make it available.  
Had to comment-out an assert  on formatters tests though, not sure about its significance.

giannisp:
I think it's not too elegant here to have `formatters.log` function responsible for returning/preserving the msg property. As per the original implementation it is supposed to shape the log object only (not including the msg property).

test/fixtures/pretty/formatters.js
giannisp: Scratch previous comment, I think I figured out the proper way for handling this.

mcollina:
This should include a revert of https://github.com/pinojs/pino/commit/4f18f264ba60bcb0536e01613fcff4d92240192e.

lib/tools.js
```diff
@@ -232,12 +232,21 @@ function prettifierMetaWrapper (pretty, dest) {
       }
 
       var lastObj = this.lastObj
+      var lastMsg = this.lastMsg
```
mcollina, line 235: Instead of reverting the deletion of `lastMsg`, would it be possible to get the message from `lastObj` instead?

jsumners: That was going to be my review question.

giannisp: Thanks for the review!
`lastObj` seems to be nullable.  
Do you propose something like:
```js
stream.lastObj = obj ? { ...obj, [messageKey]: msg } : { [messageKey]: msg } 
// stream.lastMsg = msg
```  

I am not sure about the pattern or goal let's say of having the `obj` and `msg` separate or merged overall on the codebase. As it is now it looks more clean to have them separate on this instance.

test/fixtures/pretty/formatters.js
delvedor: You don't have access to the message inside the `log` formatter:
https://github.com/pinojs/pino/blob/30d4b7d24425c7b5c133d241e1f684eb0552822d/lib/tools.js#L108-L113

If needed we can change this behavior.

lib/tools.js
delvedor: There is no technical reason why the message is not a part of the `log` formatted, it was an API design decision because the user could apply some heavy change to the log object and forget to add back the message key and value.

I tried to add back the message to the log object:
```diff
+  if (msg !== undefined) {
+    obj[messageKey] = msg
+  }
   if (formatters.log) {
     obj = formatters.log(obj)
   }
-  if (msg !== undefined) {
-    obj[messageKey] = msg
-  }
```

And if I run the following code:
```js
const pino = require('./')({
  prettyPrint: true,
  formatters: {
    log (object) {
      return { ...object, foo: 'bar' }
    }
  }
})

pino.info('Hello world')
```
The log is `[1586004564068] INFO  (73505 on Skadi.local): Hello world`, s the message now is there, but all the additional info is lost.

giannisp:
GitHub hiccup while running the last few CI actions, please trigger a re-run when you get the chance, I don't have enough access.

mcollina:
Restarted.

jsumners:
> lgtm, @jsumners what do you think?

I am unclear about whether or not `lastMsg` really needed to be reverted.

giannisp:
> > lgtm, @jsumners what do you think?
> 
> I am unclear about whether or not `lastMsg` really needed to be reverted.

I have tried encapsulating the msg in lastObj (instead of passing lastMsg), and then unboxing it before applying the `formatters.log`:
```
stream.lastObj = { ...obj, [messageKey]: msg }
```
New test works, however it breaks the pretty/obj-msg-prop test. This illustrates that a log object will have a potential conflict if it has an existing prop named as `messageKey`! So seems safer to keep using `lastMsg`, unless there's another way apart from modifying `obj`.

PR #814, opened 2020-04-07
express-pino-logger should be deprecated

deiga:
It is 2 years old and still uses pino-http 4.0.0 and it doesn't actually contain any code

### DIFF
--- a/docs/web.md
+++ b/docs/web.md
@@ -37,12 +37,14 @@ See the [fastify documentation](https://www.fastify.io/docs/latest/Logging/) for
 ## Pino with Express
 
 ```sh
-npm install express-pino-logger
+npm install pino-http
 ```
 
 ```js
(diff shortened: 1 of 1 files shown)


docs/web.md
```diff
@@ -37,12 +37,14 @@ See the [fastify documentation](https://www.fastify.io/docs/latest/Logging/) for
...
 
 ```sh
-npm install express-pino-logger
+npm install pino-http
 ```
 
 ```js
-const app = require('express')()
-const pino = require('express-pino-logger')()
+import express from 'express'
```
jsumners, line 44: Why has this example been changed to ESM syntax? This isn’t copy and pasteable without further instruction.

mcollina: Actually true! Good spot

PR #821, opened 2020-04-10
docs: Multi stream destination

robertsLando:


docs/api.md
```diff
@@ -413,6 +414,30 @@ console.log(
...
+
+```js
+const pino = require('pino')
+
+const { multistream } = require('pino-multi-stream')
+var streams = [
+  {stream: process.stdout},
+  {level: 'error', stream: process.stderr},
+  {level: 'fatal', stream: process.stderr}
+]
```
mcollina, line 432: Can you make the default example using `pino.destination()` instead?

docs/api.md
```diff
@@ -4,6 +4,7 @@
   * [options](#options)
   * [destination](#destination)
   * [destination\[Symbol.for('pino.metadata')\]](#metadata)
+  * [destination: pino-multi-stream](#multi-stream)
```
jsumners, line 7: I do not think this belongs in the API documentation because it doesn't really have anything to do with the core API. If anything, this is an example showing how to use an ecosystem module to provide a stream.

docs/api.md
robertsLando: @mcollina I'm quite new to Pino (started today) so feel free to edit my PR :)

docs/api.md
robertsLando: I have checked for other section but cannot find a better place. I agree with you that that's not the best place for it but I also think that most users coming from other loggers may find that really useful!

mcollina: I think this might be the best spot: https://github.com/pinojs/pino/blob/master/docs/help.md

robertsLando: I'm missing somethin, I thought that the first element of pino-multi-stream was for the `default` stream but that is only for the `info` level. Is there a way to set a standard stream for levels that doesn't have a defined stream?

What I would like to do is simply make all streams go to stdout and fatal and error to stderr

Issue #835, opened 2020-05-11
Clarify prettyPrint documentation


AleksandrSl:
I think that there are some old "code" in the docs about pretty print.

For example this
```
  function isPinoLog (log) {
    return log && (log.hasOwnProperty('v') && log.v === 1)
  }
```
function. From this https://github.com/pinojs/pino/pull/623 I understood that `v` was removed some time ago.

Also there is no infromation about why type of inputData could be different. Based on pino code it seems to me that object is always passed. If there are acutally another cases, maybe it will be helpfull to clarify when they are possible.

mcollina:
Would you like to send a PR?

AleksandrSl:
Yaeh, I can remove mentions of `v` in docs, but I'm not sure about prettifier `inputData` type clarification.

mcollina:
I think it should be changed by the field
used by pino-pretty

AleksandrSl:
Sorry, I don't fully understand what
> it should be changed by the field
> used by pino-pretty

means

AleksandrSl:
Maybe string is passed when prettifier is used with pipe (as stated in docs - `cat app.log | pino-pretty`), and if prettifier is registered in app it will always get objects? Then this can be mentioned in docs and we can write less boiler plate in custom prettifiers, if they are not supposed to be used with pipes.

mcollina:
I don't understand :/

AleksandrSl:
In pretty print docs
```
  return function prettifier (inputData) {
    let logObject
    if (typeof inputData === 'string') {
      logObject = someJsonParser(inputData)
    } else if (isObject(inputData)) {
      logObject = inputData
    }
    if (!logObject) return inputData
    // implement prettification
  }
```
 it's stated that prettifier function for some reason could get from pino two types of `inputData` - `string` and `Object`. While in the pino code, prettifier function is only called with `Object`. So I don't understand why string option exist. 
Maybe this example was taken from `pino-pretty`, which has an option to parse logs from stream (i.e. `cat app.log | pino-pretty`) and in that case prettifier function could recieve `string`?

I think that prettifier example could be simplified, because you could write prettifier that works only from code and doesn't support piping logs into it.

jsumners:
Please read the text in that section of the docs prior to that code sample. It very clearly explains the example.

AleksandrSl:
@jsumners 

[...]

PR #623, opened 2019-03-30
Remove log version

thomasthiebaud:
Fixes #620 

Removing `v1` breaks a test in `pretty.test.js`. To fix it I have to update `pino-pretty`. I wanted to do so, but it seems there is a circular dependency when running tests `pino` -> `pino-pretty` -> `pino`.
How do you usually handle update both packages at the same time?

How should `pino-pretty` handle it? Removing this check

```
function isPinoLog (log) {
  return log && (log.hasOwnProperty('v') && log.v === 1)
}
```

seems to be the easiest way, but then `pino-pretty` will print logs not coming from `pino` like that

```
USERLVL:           
    hello: "world"
```

PR #836, opened 2020-05-11
Remove obsolete logger field mentions from docs

AleksandrSl:
Version logger field was removed in https://github.com/pinojs/pino/pull/623 but some places were missed. part of #835


AleksandrSl:
Seems that windows failure is not connected with docs changes

Issue #825, opened 2020-04-13
Drop all references to pino.extreme


mcollina:
As titled, this as been included inside pino.destination().

AleksandrSl:
I started working on this. Should we preserve term `extreme mode`, or it could be renamed to something like `asynchronous mode`? 
There are also a bunch of `extreme` variables in tests and `pino.extreme` usage. Should they also be replaced?

mcollina:
Please replace! It would be super helpful.

AleksandrSl:
Should calls to `pino.extreme` be replaced using `minLength: 4096`, if so maybe there is some sense in providing function wrapper?

mcollina:
> Should calls to pino.extreme be replaced using minLength: 4096, if so maybe there is some sense in providing function wrapper?

just replace them. The current API is more flexible.

AleksandrSl:
Can be closed, I think

PR #838, opened 2020-05-11
Remove pino extreme mentions in docs and code

AleksandrSl:
#825


AleksandrSl:
Some mentions are left in `test/syncfalse.test.js`, cause they test warning. I think that they should be removed together with `pino.extreme`

PR #839, opened 2020-05-12
Disable code coverage on Windows

mcollina:
Currently we fail on Windows and node v14 for some problem (https://github.com/nodejs/node/issues/33166).
Therefore, I'm disabling coverage on Windows.


mcollina:
I've landed this as it really does not change much and it makes our CI green.

PR #841, opened 2020-05-18
Allow to suppress prettyPrint flush sync warning

AleksandrSl:
Described in https://github.com/pinojs/pino-pretty/issues/108 issue.
I understood that this issue belongs more to the pino, should I close in pino/pretty and move to pino repo?

### DIFF
--- a/docs/pretty.md
+++ b/docs/pretty.md
@@ -85,10 +85,10 @@ will be written to the destination stream.
   See the [`pino-pretty` documentation][pp] for more information on the options
   that can be passed via `prettyPrint`.
    
-  Pretty stream doesn't guarantee final log writes and corresponding warning is written to logs on 
-  first synchronous flushing.
-  To disable this warning, e.g. if you use pretty stream consciously and know that some logs can be lost,
-  you can pass `suppressFlushSyncWarning : true` to prettyPrint:
+The default prettifier write stream does not guarantee final log writes.
+Correspondingly, a warning is written to logs on first synchronous flushing.
+This warning may be suppressed by passing `suppressFlushSyncWarning : true` to
+`prettyPrint`:
   ```js
   const pino = require('pino')
   const log = pino({
--- a/docs/pretty.md
+++ b/docs/pretty.md
@@ -84,5 +84,18 @@ will be written to the destination stream.
   ```
   See the [`pino-pretty` documentation][pp] for more information on the options
   that can be passed via `prettyPrint`.
+   
+  Pretty stream doesn't guarantee final log writes and corresponding warning is written to logs on 
+  first synchronous flushing.
+  To disable this warning, e.g. if you use pretty stream consciously and know that some logs can be lost,
+  you can pass `suppressFlushSyncWarning : true` to prettyPrint:
+  ```js
+  const pino = require('pino')
+  const log = pino({
+    prettyPrint: {
+      suppressFlushSyncWarning: true
+    }
+  })
+  ```
 
   [pp]: https://github.com/pinojs/pino-pretty
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -128,7 +128,7 @@ function asJson (obj, msg, num, time) {
           if (Number.isFinite(value) === false) {
             value = null
           }
-        // this case explicity falls through to the next one
+        // this case explicitly falls through to the next one
         case 'boolean':
           if (stringifier) value = stringifier(value)
           break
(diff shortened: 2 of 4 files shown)


lib/tools.js
```diff
@@ -177,18 +177,18 @@ function asChindings (instance, bindings) {
...
     prettyFactory.asMetaWrapper = prettifierMetaWrapper
-    return prettifierMetaWrapper(prettyFactory(opts), dest)
+    return prettifierMetaWrapper(prettyFactory(opts), dest, opts.suppressFlushSyncWarning)
   } catch (e) {
     throw Error('Missing `pino-pretty` module: `pino-pretty` must be installed separately')
   }
 }
 
-function prettifierMetaWrapper (pretty, dest) {
+function prettifierMetaWrapper (pretty, dest, suppressFlushSyncWarning) {
```
AleksandrSl, line 191: Seems to onvasive for me, but it's the most straightforward way

lib/tools.js
```diff
@@ -177,18 +177,18 @@ function asChindings (instance, bindings) {
 function getPrettyStream (opts, prettifier, dest, instance) {
   if (prettifier && typeof prettifier === 'function') {
     prettifier = prettifier.bind(instance)
-    return prettifierMetaWrapper(prettifier(opts), dest)
+    return prettifierMetaWrapper(prettifier(opts), dest, opts.suppressFlushSyncWarning)
   }
   try {
     var prettyFactory = require('pino-pretty')
     prettyFactory.asMetaWrapper = prettifierMetaWrapper
```
AleksandrSl, line 184: I also wanted to ask why we don't `bind` instance here?

lib/tools.js
```diff
@@ -177,18 +177,18 @@ function asChindings (instance, bindings) {
 function getPrettyStream (opts, prettifier, dest, instance) {
   if (prettifier && typeof prettifier === 'function') {
```
AleksandrSl, line 178: Another option/ Maybe we can emit warning here once on stream creation?
Something like 
```
      process.emitWarning(
        'PrettyPrint does not support flushing',
      )
```

mcollina: How could you detect if the user intend to call flushSync?

lib/tools.js
mcollina: I don’t understand, how would you bind() it?

AleksandrSl: Same way as at 179 `prettyFactory = prettyFactory.bind(instance)`.
`prettyFactory` seems to be the same function as `prettifier`

lib/tools.js
AleksandrSl: Fair remark, but isn't it more explicit to warn user during "application" start, instead of when problem is stumbled upon. 
Moreover `warning` can be missing from logs, cause application can exit before something has been written?

AleksandrSl: Sorry for silly questions/remarks I don't have much experience in production js (moreover nodejs) code, and just want to acquire understanding

docs/pretty.md
```diff
@@ -84,5 +84,18 @@ will be written to the destination stream.
   ```
   See the [`pino-pretty` documentation][pp] for more information on the options
   that can be passed via `prettyPrint`.
+   
+  Pretty stream doesn't guarantee final log writes and corresponding warning is written to logs on 
+  first synchronous flushing.
+  To disable this warning, e.g. if you use pretty stream consciously and know that some logs can be lost,
+  you can pass `suppressFlushSyncWarning : true` to prettyPrint:
```
jsumners, line 91:
```suggestion
The default prettifier write stream does not guarantee final log writes.
Correspondingly, a warning is written to logs on first synchronous flushing.
This warning may be suppressed by passing `suppressFlushSyncWarning : true` to
`prettyPrint`:
```

AleksandrSl: Thanks!

lib/tools.js
mcollina: > Fair remark, but isn't it more explicit to warn user during "application" start, instead of when problem is stumbled upon.

How can you detect that? This is triggered by folks manually calling `pino.stream.flushSync()` or are you triggering it in some other way?

lib/tools.js
mcollina: I still don't understand why would you bind it?

AleksandrSl: Ah, why. Just to conform to prettier API, 

> The uninitialized Pino instance is passed as this into prettifier factory function, so it can be accessed via closure by the returned prettifier function.

lib/tools.js
AleksandrSl: I meant without detection, just for every `pretty stream` user, for them to know that it's impossible to use `flushSync`. 
I know that this is written in docs, but warning still exist for some reason and is shown only once.
But now I see that this may be too disturbing.

> are you triggering it in some other way?
It's triggered by `logger.fatal` in our case

Issue #847, opened 2020-05-26
Object passed during logger instantiation not redacted


P4sca1:
Custom object properties are not redacted. This is especially critical for the `pino-http` package. There the `req` object can't be redacted, so things like `req.headers.authorization` are always exposed to logs.

Example code:
```ts
const pino = require('pino')
const express = require('express')
const expressPino = require('express-pino-logger')

const logger = pino({
    redact: {
      paths: ['*.headers.authorization'],
    },
  })
  
const app = express()

app.use(
  expressPino({
    logger,
  })  
)

app.listen(8000)
```

Start the app and do a http request:
`curl -H "authorization: Bearer test" http://localhost:8000`.
You will see that req.headers.authorization is not redacted.

What happens is that the `req` object is never passed to log redaction, because it is not passed to the logger functions, but during child logger instantiation:
https://github.com/pinojs/pino-http/blob/d593aa6084f1fe70a506ad8f202fc7f339b87fb5/logger.js#L94

Would be great if pino would redact custom log properties.

mcollina:
Would you have time to send a PR? We are a bit strained in time for the coming weeks and I would love to get this solved sooner than that.

I _think_ the problem happens because we are not passing through redaction when calling `child()`.

baterson:
@mcollina I would like to take this issue

mcollina:
go for it!

P4sca1:
@baterson Thank you very much for fixing this!

khalilchoudhry:
`redact: {
    paths: ['req.body.password', '*.headers.authorization'],
    remove: true
  }`
  
  Not sure if this is related or I need to ask this on pino-http repo but setting this redact field like:
  `
  const pino = require('pino');
  const pinoHttp = require('pino-http');
  
  const logger = pino({
  name: 'abc'
});

  const expressPino = pinoHttp({
  logger,
  redact: {
    paths: ['req.body.password', '*.headers.authorization'],
    remove: true
  }
  })
  `
  
  doesn't redact the mentioned fields. pino-http accepts redact as an input parameter but still doesn't redact the fields.
  setting redact option to pino var works

PR #856, opened 2020-06-02
Issue 87 child binding redact bug

baterson:
Hello, it's a fix for the issue https://github.com/pinojs/pino/issues/847


baterson:
@mcollina Thanks! Btw what does `chindings` mean? Is it "child bindings"?

mcollina:
> Btw what does chindings mean? Is it "child bindings"?

Exactly that. @davidmarkclements loves to create new words.

PR #868, opened 2020-06-27
order levels, trace maps to DEBUG severity

naseemkullah:
cc @rahulballal

As per gcp docs severity DEBUG(100) is for debug or trace information.

Signed-off-by: Naseem <naseem@transit.app>


naseemkullah:
My pleasure @mcollina.

PR #875, opened 2020-07-14
Wrap prototype in factory and guard against loss of scope. 

### LINKED ISSUE #833: TypeError: Cannot read property 'labels' of undefined
```js
//test.js
const pino = require('pino')();

process.on('warning', (warning) => {
  pino.info(warning);
});
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})
process.on('foo', () => {});
process.on('foo', () => {})

```

  return levels.labels[levelVal]
               ^

TypeError: Cannot read property 'labels' of undefined

admataz:
The scope of the Pino class instance is lost when circular refs are routed through `decirc` - and `levels` is undefined in the getter. 

This PR wraps the Pino prototype in a factory - which helps with scope, and adds a protection in the `getLevels` function, with a  simple test to keep coverage up. 

Fixes #833

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -65,7 +65,9 @@ const prototype = {
 Object.setPrototypeOf(prototype, EventEmitter.prototype)
 
 // exporting and consuming the prototype object using factory pattern fixes scoping issues with getters when serializing
-module.exports = () => prototype
+module.exports = function () {
+  return Object.create(prototype)
+}
 
 const resetChildingsFormatter = bindings => bindings
 function child (bindings) {
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -64,7 +64,8 @@ const prototype = {
 
 Object.setPrototypeOf(prototype, EventEmitter.prototype)
 
-module.exports = prototype
+// exporting and consuming the prototype object using factory pattern fixes scoping issues with getters when serializing
+module.exports = () => prototype
 
 const resetChildingsFormatter = bindings => bindings
 function child (bindings) {
--- a/test/levels.test.js
+++ b/test/levels.test.js
@@ -553,6 +553,7 @@ test('changing level from info to silent and back to info in child logger', asyn
   check(is, result, expected.level, expected.msg)
 })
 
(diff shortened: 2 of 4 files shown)


test/levels.test.js
```diff
@@ -550,3 +552,9 @@ test('changing level from info to silent and back to info in child logger', asyn
   result = await once(stream, 'data')
   check(is, result, expected.level, expected.msg)
 })
+
+test('trying to get levels when `this` is no longer a Pino instance returns an empty string', async ({ is }) => {
```
admataz, line 556: this test is out of line with the others - and because of what it's testing (not Pino) needs to access the `getLevel` function directly on an anonymous object.

lib/proto.js
```diff
@@ -64,7 +64,8 @@ const prototype = {
 
 Object.setPrototypeOf(prototype, EventEmitter.prototype)
 
-module.exports = prototype
+// exporting and consuming the prototype object using factory pattern fixes scoping issues with getters when serializing
+module.exports = () => prototype
```
jsumners, line 68: I'm not clear how this is any different. It's still returning the single prototype object. Wouldn't we instead want to do?:

```js
module.exports = function () {
  return Object.create(prototype)
}
```

admataz: Thank you. Yes, that makes sense. I _think_ this still works for the reported bug because the scope of where the factory function is called - but a new instance from the factory is better.

test/levels.test.js
mcollina: why is this out of line with the others? Add a reference to the issue

PR #876, opened 2020-07-14
Avoid data loss with conflictig 'msg' values.

### LINKED ISSUE #874: Including `msg` in object and string leads to silent data loss
To reproduce:

```
const pino = require('pino')();
pino.info({"msg":"OBJ"},"STRING")
```

The result is that "STRING" is logged and "OBJ" is silently discarded.

This could happen when you use `pino.info(obj)` on an object that happens to have a `msg` property.

The recommended fix when `msg` appears in the object and the string is to throw an error. The current behavior is silent data loss.

There's nothing worse than going to check your logs for some critical detail about a past event only to find it was silently discarded by your logging library and you can't retroactively fix it. If you've got an object that contains a msg property that you want to log, instead of:

    log.info(obj)

You can do:

    log.info({obj}).

The error thrown for the duplicate `msg` case could explain that.

markstos:
If msg appears as a string and in the object, preserve it as "originalMsg"

Before, the `msg` copy in the the object was lost.

Fixes #874

`npm run bench-basic` reports results within about a millisecond before and after this change, which is within the margin of arability on my laptop.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -488,6 +488,8 @@ The `message` parameter takes precedence over the `mergedObject`.
 That is, if a `mergedObject` contains a `msg` property, and a `message` parameter
 is supplied in addition, the `msg` property in the output log will be the value of
 the `message` parameter not the value of the `msg` property on the `mergedObject`.
+In case this conflict occurs, the `msg` property from the object will be preserved
+in the key `originalMsg`.
 
 The `messageKey` option can be used at instantiation time to change the namespace
 from `msg` to another string as preferred.
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -109,6 +109,10 @@ function asJson (obj, msg, num, time) {
     obj = formatters.log(obj)
   }
   if (msg !== undefined) {
+    // If msg appears as a string and in the object, preserve it as "originalMsg"
+    if (obj[messageKey] !== undefined) {
+      obj.originalMsg = obj[messageKey]
+    }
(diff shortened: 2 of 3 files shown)


jsumners:
@mcollina do you think these are significant?

**branch**
```
BASIC benchmark averages
Pino average: 233.880ms
PinoAsync average: 123.876ms
PinoNodeStream average: 261.453ms
```

**master**
```BASIC benchmark averages
Pino average: 221.129ms
PinoAsync average: 109.718ms
PinoNodeStream average: 263.892ms
```

I'm really wondering if this is already possible with the hook we added in v6.

mcollina:
I'll run it on my dedicated server and see

markstos:
On my laptop I tested several runs with and without the change and got different results over different runs.

The boils down the cost of a single `if` statement. No additional work is done if there's no conflict. 

I do appreciate your attention to performance concerns, though.

mcollina:
> I'm really wondering if this is already possible with the hook we added in v6.

Maybe, but this looks like a sane thing to add that does not hurt anybody.

mcollina:
wdyt @jsumners?

markstos:
While this PR was reverted for others who might end up here looking to solve the same problem, I'm noting that the same goal of avoiding data loss when `msg` values conflict can be met by using the `nestedKey` config value. This puts all the logged object keys in their own namespace: 

https://github.com/pinojs/pino/blob/master/docs/api.md#nestedkey-string

PR #882, opened 2020-07-31
Revert "Avoid data loss with conflicting 'msg' values."

jsumners:
Reverts pinojs/pino#876

This causes issues in dependent applications. See https://github.com/fastify/fastify/issues/2453

To get the same effect, use the log method hook:

```js
'use strict'

const log = require('pino')({
  level: 'debug',
  hooks: {
    logMethod (inputArgs, method) {
      if (inputArgs.length === 2 && inputArgs[0].msg) {
       inputArgs[0].originalMessage = inputArgs[0].msg
      }
      return method.apply(this, inputArgs)
    }
  }
})

log.info('no original message')
log.info({ msg: 'mapped to originalMsg' }, 'a message')

// {"level":30,"time":1596195825857,"pid":48580,"hostname":"mini2018","msg":"no original message"}
// {"level":30,"time":1596195825858,"pid":48580,"hostname":"mini2018","msg":"a message","originalMessage":"mapped to originalMsg","originalMsg":"mapped to originalMsg"}
```

cc: @markstos

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -488,8 +488,6 @@ The `message` parameter takes precedence over the `mergedObject`.
 That is, if a `mergedObject` contains a `msg` property, and a `message` parameter
 is supplied in addition, the `msg` property in the output log will be the value of
 the `message` parameter not the value of the `msg` property on the `mergedObject`.
-In case this conflict occurs, the `msg` property from the object will be preserved
-in the key `originalMsg`.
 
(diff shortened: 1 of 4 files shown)


mcollina:
Could you add that info to the docs as well, maybe in help?

jsumners:
@mcollina updated.

PR #884, opened 2020-08-04
Make the nestedKey only take effect in the serialized object and fix …

### LINKED ISSUE #883: Properly handling the 'nestedKey' option
Let me start by congratulating you guys on this awesome project and by saying I have read #872, #771 and #792 and the docs available [here](https://github.com/pinojs/pino/blob/master/docs/api.md#nestedkey-string).

Personally what I understand from the docs (and what I would expect to happen) is that the serializers will still receive the original object, not the one after the 'nestedKey' transformation (as the base is static, the dynamic part is the passed object and the serializers register only on root level properties, making the serializers register feature pointless when used with 'nestedKey'). But this is my opinion and is subject to interpretation.

However, activating the 'nestedKey' option breaks the standard serialization of errors (because the detection fails) which is a bug (standard options should not break each other).

The problem seems to be that the passed object is transformed [here](https://github.com/pinojs/pino/blob/83d80f63888aa0b68479d460706644c58257131c/lib/tools.js#L47) and afterwards is checked for it's type [here](https://github.com/pinojs/pino/blob/83d80f63888aa0b68479d460706644c58257131c/lib/proto.js#L144) effectively transforming all Errors into plain objects so the detection is never true.

I think it would be a good idea to do the transformation after the check.

If you guys agree I would be happy to make a PR.

mihai1voicescu:
…error detection and serialization (closes #883)

proto.js
* Copy the message property in case the obj is an error and there is a msg specified
* Load the msg as a property of obj if metadata is needed

symbols.js
* Add nestedKeyStrSym to quickly cache the string

tools.js
* Manually build the stringify for performance reasons (not too pretty...)

error.test.js + serializers.test.js
* Add a test for the new nestedKey interaction

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -20,7 +20,8 @@ const {
   serializersSym,
   formattersSym,
   useOnlyCustomLevelsSym,
-  needsMetadataGsym
+  needsMetadataGsym,
+  messageKeySym
 } = require('./symbols')
 const {
   getLevel,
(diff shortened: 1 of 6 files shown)


mihai1voicescu:
I seem to be having a problem with the http tests on local though. They just time out (on master before my commit also).

PR #885, opened 2020-08-04
Make the nestedKey only take effect in the serialized object and fix …

### LINKED ISSUE #883: Properly handling the 'nestedKey' option

mihai1voicescu:
…error detection and serialization (closes #883)

proto.js
* Copy the message property in case the obj is an error and there is a msg specified
* Load the msg as a property of obj if metadata is needed

symbols.js
* Add nestedKeyStrSym to quickly cache the string

tools.js
* Manually build the stringify for performance reasons (not too pretty...)

error.test.js + serializers.test.js
* Add a test for the new nestedKey interaction


mihai1voicescu:
I finally managed to make the http test run on local... new antivirus :) 

I will fix the coverage ASAP.

lib/tools.js
```diff
@@ -139,11 +137,45 @@ function asJson (obj, msg, num, time) {
...
+    }
+  }
+
+  let msgStr = ''
+  if (msg !== undefined) {
+    value = serializers[messageKey] ? serializers[messageKey](msg) : msg
+    const stringifier = stringifiers[messageKey] || wildcardStringifier
+
+    switch (typeof value) {
+      case 'function':
```
mihai1voicescu, line 150: Should I cover all these cases here?

lib/tools.js
```diff
@@ -139,11 +137,45 @@ function asJson (obj, msg, num, time) {
...
+        if (stringifier) value = stringifier(value)
+        msgStr = ',"' + messageKey + '":' + value
+        break
+      case 'string':
+        value = (stringifier || asString)(value)
+        msgStr = ',"' + messageKey + '":' + value
+        break
+      default:
+        value = (stringifier || stringify)(value)
+        msgStr = ',"' + messageKey + '":' + value
```
mcollina, line 168: Was this code lifted/copied from somewhere else?

mihai1voicescu: It's the handling of the values from above https://github.com/pinojs/pino/blob/83d80f63888aa0b68479d460706644c58257131c/lib/tools.js#L118

I duplicated it to avoid creating a separate function (for performance reasons).

But maybe all that handling is not needed for only the 'msg' key.

mihai1voicescu: Regrding the benchmarks registered benchmarks on the cloud, they seem to vary by a statistically significant ammount even in the other loggers:
image

But it's very late and I am tired so I might be overlooking something 😄.

I will investigate and get back to you.

Also, do you guys have a better tool other then diff for comparing the performance (maybe something that registers the % gain and the standard deviation of the ran tests)?

And I see the clould machine is an Azure, I guess it's a VM without guaranteed resources so the performance might vary over time (didn't use Azure a lot so I might be mistaken).

jsumners: The benchmarks that are automatically run by the PR process are run in GitHub's Actions platform. At the bottom of each file you will find a summary of the each result that is the average of all the runs for each result type.

mihai1voicescu: I ran the `update-bench-doc` task with both versions on my mac. The results seem to be almost the same (some times one version or the other is faster but not by something significant).

Left side is the old version and the right is the new PR.
image

I think the problem on the clould is the fact that it runs on a VM w/o guaranteed resources. This means performance can vary widly according to the usage of other VMs sharing the same underlying resources.

Do you guys agree? Maybe someone can double check on a phisical machine just to be sure.

jsumners: The VM resources are supposed to be the same across all runs -- https://docs.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

Anyway, the benchmark action is merely there to provide a quick inspection. True benchmark comparisons need to be run on a dedicated system if the PR benchmarks give an indication that something may be amiss.

mihai1voicescu: So I think that this means the benchmarks are OK. I will resolve this conversation and ask for a re-review.

lib/proto.js
```diff
@@ -148,11 +149,15 @@ function write (_obj, msg, num) {
...
-    }
 
     if (objError) {
+      if (!msg) {
+        // replace the main msg with the message of the error
+        msg = _obj.message
+      } else {
+        // if we do have a msg, we should keep the message found on the error
+        obj.message = _obj.message
+      }
```
jsumners, line 160: Why? The current implementation doesn't do this, does it?

lib/proto.js
```diff
@@ -165,6 +170,9 @@ function write (_obj, msg, num) {
   const stream = this[streamSym]
   if (stream[needsMetadataGsym] === true) {
     stream.lastLevel = num
+    if (msg) {
+      obj[this[messageKeySym]] = msg
+    }
```
jsumners, line 175: Why is this here?

lib/tools.js
jsumners: If this is going to be left in, yes. But I'm really confused as to why it is needed.

jsumners (dismissed):
I think this PR is addressing multiple things at once. If that is the case, please reduce it down to one issue and open subsequent PRs for other issues.

lib/tools.js
mihai1voicescu: Well the `msg` key needs to be treated separatly so it can be placed on the root level and not the nested key one.

lib/proto.js
mihai1voicescu: If we only pass the obj, if the `nestedKey` option is activated then the obj will have no msg key. If the option is false, then the obj will have the key.

I did this to be as close as possible to the  `nestedKey=false` behavior.

lib/proto.js
mihai1voicescu: It's true it does not.

Because it does not it results in loosing the `message` string when a `msg` is present, which, in my oppinion is not OK.

But it's true, maybe this should be a separate issue. Should I make a separate one?

jsumners: Yes, please keep issues scoped to one item. It is very difficult to review PRs that address multiple things. However, see https://github.com/pinojs/pino/pull/882

lib/proto.js
mihai1voicescu: The `msg` property is no longer placed on the `obj` in the `asJson` function.

Thing is, the `lastMsg` is also placed here so this is not requeired, but the behaviour will be a little diffrent (although arguably more correct), which I tried to avoid as much as possible, while keeping things performant.

lib/proto.js
mihai1voicescu: I created #889. #882 and  #876 do not seem related to this issue.

lib/proto.js
mihai1voicescu: After reading #882 I realised the `msg`, as deduced by the algorithm, maybe should not be in the `obj` at all.

This is why you guys put in the `originalMsg`, because you basically wanted both messages. But this is why I used the `nestedKey` (to avoid all collisions).

So maybe the good fix is to not replace the `msg` property in the `obj` and use `nestedKey`.

The [new handling of the `msg` algorithm in the PR](https://github.com/mihai1voicescu/pino/blob/783bbfa20db90a8a617057242851b60c265ee361/lib/tools.js#L146) will place the `msg` key as the last thing in the JSON log w/o interfering with the `obj`.

mihai1voicescu: I will remove the if, and leave the obj without the msg. This is available in lastMsg

mihai1voicescu:
I addressed the issues present in the PR comments.
Note that I solved https://github.com/pinojs/pino/pull/885#discussion_r468515301 by removing the if which causes the `obj` to miss the `msg` property. This is available in `this.lastMsg`. If you are against this I can put it back in.

test/basic.test.js
```diff
@@ -681,3 +681,36 @@ test('correctly log NaN', async (t) => {
...
+  const { msg } = await once(stream, 'data')
+  t.is(msg, null)
+})
+
+test('correctly skip Infinity', async (t) => {
+  const stream = sink()
+  const instance = pino(stream)
+
+  const o = { num: NaN }
+  instance.info(o, 42)
```
markstos, line 712: I think the test name is wrong here-- it appears to simply be a copy/paste of the test name above. Perhaps the intent was "correctly skip Number" ?

test/metadata.test.js
```diff
@@ -40,7 +40,7 @@ test('child loggers works', async ({ ok, same, is }) => {
       is(child, this.lastLogger)
       is(30, this.lastLevel)
       is('a msg', this.lastMsg)
-      same(this.lastObj, { from: 'child', msg: 'a msg' })
+      same(this.lastObj, { from: 'child' })
```
mcollina, line 43: I have a feeling this might be a semver-major change.

mcollina (approved):
lgtm, benchmarks are ok

test/basic.test.js
```diff
@@ -681,3 +681,36 @@ test('correctly log NaN', async (t) => {
...
+  const instance = pino(stream)
+
+  const o = { num: NaN }
+  instance.info(o, Infinity)
+
+  const { msg } = await once(stream, 'data')
+  t.is(msg, null)
+})
+
+test('correctly skip Infinity', async (t) => {
```
mcollina, line 707:
```suggestion
test('correctly skip NaN', async (t) => {
```

test/basic.test.js
mcollina: I'm not sure this test is really doing anything actually. @mihai1voicescu can you clarify?

mihai1voicescu: I forgot to rename the test 😢. Should test that given a number, will print the number as string in the msg.
Basically testing the switch case.

test/metadata.test.js
mihai1voicescu: I agree, should I update the version in the package.json?

jsumners:
My last review question never got answered. Is this PR solving multiple issues at once?

mihai1voicescu:
@jsumners it is now addressing a single issue (the nested property).
After your comments I split the other stuff in a different issue.

mcollina:
@jsumners are you ok in landing this?

jsumners:
@mcollina sorry, I'm having a really hard time following this and just don't have enough time to devote to it right now. I leave it up to you and @davidmarkclements.

mcollina:
Can you please rebase it? I'm sorry we've let this stale.

mcollina (approved):
lgtm

mihai1voicescu:
@mcollina looks like one check entitled `CI / 12 macOS-latest (pull_request)` failed but I don't see any logs associated with it.

What should I do?

jsumners:
GitHub Actions is having issues at this time https://www.githubstatus.com/incidents/m16jzl31gnqt

PR #893, opened 2020-08-12
Clear cached parsed chindings when they're changed with .setBindings

### LINKED ISSUE #890: logger.setBindings() doesn't clear logger[Symbol(pino.parsedChindings)]
#754 added a very useful `logger.setBindings()` method. But this doesn't play well with prettifiers, which appear to make use of a cached version of the bindings:

```js
> let child=logger.child({foo: 123});

// pino.parsedChindings is undefined at start
> child.bindings()
{ foo: 123 }
> child[Pino.symbols.parsedChindingsSym]
undefined

// when the first log call is made, pino.parsedChindings gets populated
> child.info('hello')
[2020-08-11 23:12:27.349 +0000] INFO : hello
    foo: 123
> child[Pino.symbols.parsedChindingsSym]
{ pid: 16561, hostname: 'nutrimatic', foo: 123 }

// logger.setBindings doesn't update pino.parsedChindings
> child.setBindings({bar: 456})
> child.bindings()
{ foo: 123, bar: 456 }
> child[Pino.symbols.parsedChindingsSym]
{ pid: 16561, hostname: 'nutrimatic', foo: 123 }

// so subsequent logs don't include the new bindings
> child.info('hello')
[2020-08-11 23:13:24.553 +0000] INFO : hello
    foo: 123
> child[Pino.symbols.parsedChindingsSym]
{ pid: 16561, hostname: 'nutrimatic', foo: 123 }

// manually clearing pino.parsedChindings causes it to be regenerated on the next log call
> delete child[Pino.symbols.parsedChindingsSym]
> child.info('hello')
[2020-08-11 23:14:28.240 +0000] INFO : hello
    foo: 123
    bar: 456
> child[Pino.symbols.parsedChindingsSym]
{ pid: 16561, hostname: 'nutrimatic', foo: 123, bar: 456 }
```

Would it be "safe" (performance-wise) to have `logger.setBindings()` automatically clear `pino.parserChindings` so that the next log call causes it to update? I think this may be OK since said updating only happens when a prettifier is enabled, so it won't impact production situations where there's no prettifier. Alternatively, could we add an additional method that manually clears `pino.parserChindings`?

I can submit a PR for either of these approaches; I'm just not sure which is best.

okofish:
This fixes #890; see that issue for discussion.

### DIFF
--- a/test/fixtures/pretty/child-with-updated-chindings.js
+++ b/test/fixtures/pretty/child-with-updated-chindings.js
@@ -1,0 +1,8 @@
+global.process = { __proto__: process, pid: 123456 }
+Date.now = function () { return 1459875739796 }
+require('os').hostname = function () { return 'abcdefghijklmnopqr' }
+var pino = require(require.resolve('./../../../'))
+var log = pino({ prettyPrint: true }).child({ foo: 123 })
+log.info('before')
+log.setBindings({ foo: 456, bar: 789 })
+log.info('after')
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -11,6 +11,7 @@ const {
   setLevelSym,
   getLevelSym,
   chindingsSym,
+  parsedChindingsSym,
   mixinSym,
   asJsonSym,
   writeSym,
@@ -136,6 +137,7 @@ function bindings () {
 function setBindings (newBindings) {
   const chindings = asChindings(this, newBindings)
   this[chindingsSym] = chindings
+  delete this[parsedChindingsSym]
 }
 
 function write (_obj, msg, num) {
--- a/test/pretty.test.js
+++ b/test/pretty.test.js
@@ -112,6 +112,20 @@ test('parses and outputs chindings', async ({ is, isNot }) => {
   is(strip(actual).match(/a: 1/g).length, 3)
 })
 
+test('applies updated chindings', async ({ is, isNot }) => {
+  var actual = ''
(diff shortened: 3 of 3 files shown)


test/fixtures/pretty/child-with-updated-chindings.js
```diff
@@ -0,0 +1,8 @@
+global.process = { __proto__: process, pid: 123456 }
+Date.now = function () { return 1459875739796 }
+require('os').hostname = function () { return 'abcdefghijklmnopqr' }
+var pino = require(require.resolve('./../../../'))
+var log = pino({ prettyPrint: true }).child({ foo: 123 })
+log.info('before')
+log.setBindings({ foo: 456, bar: 789 })
```
jsumners, line 7: Are we really exposing this as a public API? Can we exercise the code path without using this function?

okofish: Which function are you referring to, `.setBindings()`? The point of this PR is to make that method work correctly when a prettifier is in use.

mcollina: setBindings is already exposed.

jsumners: I don't see it in our docs -- https://getpino.io/#/docs/api

mcollina: Maybe it should be documented, anyway this PR does not add it.

PR #896, opened 2020-08-16
Apply err serializer everywhere.

### LINKED ISSUE #895: Error serializer only works on sub err object
I'm unsure to understand why the `err` serializer option only works with an `err` key and not with all passed errors.

I would expect `logger.info(ReferenceError('test'))` to also use the optional serializer passed to pino just like `logger.info({ err: ReferenceError('test') })` does.

Hence I would expect this test to pass:
```js
const stream = sink()
const parent = pino({
  serializers: {
    err: (e) => ({
      t: e.constructor.name,
      m: e.message,
      s: e.stack
    })
  }
}, stream)

parent.info(ReferenceError('test'))
const o = await once(stream, 'data')
is(typeof o, 'object')
is(o.t, 'ReferenceError')
is(o.m, 'test')
is(typeof o.s, 'string')
```

But unfortunately it doesn't. Seem that it still uses the default error serializer rather than the custom one.

Altering the default serializer doesn't help either.

```js
const stream = sink()
pino.stdSerializers.err = (e) => ({
  t: e.constructor.name,
  m: e.message,
  s: e.stack
})
const parent = pino({}, stream)

parent.info(ReferenceError('test'))
const o = await once(stream, 'data')
is(typeof o, 'object')
is(o.t, 'ReferenceError')
is(o.m, 'test')
is(typeof o.s, 'string')
```

mcollina:
Apply the error serializer if the error object is top-level. It does not seem to cause any performance regressions from my checks.

Fixes #895

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -151,9 +151,9 @@ function write (_obj, msg, num) {
     const serializers = this[serializersSym]
     obj = Object.assign(mixin ? mixin() : {}, serializers.err(_obj))
     // backward compat, set message
-    if (obj.message && !msg) {
-      msg = obj.message
-      obj.message = undefined
+    obj.message = undefined
+    if (!msg) {
+      msg = _obj.message
     }
   } else {
     obj = Object.assign(mixin ? mixin() : {}, _obj)
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -143,23 +143,20 @@ function setBindings (newBindings) {
 function write (_obj, msg, num) {
   const t = this[timeSym]()
   const mixin = this[mixinSym]
-  const objError = _obj instanceof Error
   var obj
 
   if (_obj === undefined || _obj === null) {
     obj = mixin ? mixin() : {}
+  } else if (_obj instanceof Error) {
+    const serializers = this[serializersSym]
+    obj = Object.assign(mixin ? mixin() : {}, serializers.err(_obj))
+    // backward compat, set message
+    if (obj.message && !msg) {
+      msg = obj.message
+      obj.message = undefined
+    }
   } else {
     obj = Object.assign(mixin ? mixin() : {}, _obj)
-    if (!msg && objError) {
-      msg = _obj.message
-    }
-
-    if (objError) {
-      obj.stack = _obj.stack
-      if (!obj.type) {
-        obj.type = 'Error'
-      }
-    }
   }
 
   const s = this[asJsonSym](obj, msg, num, t)
--- a/docs/api.md
+++ b/docs/api.md
@@ -257,6 +257,10 @@ These functions should return an JSONifiable object and they
 should never throw. When logging an object, each top-level property
 matching the exact key of a serializer will be serialized using the defined serializer.
 
+The serializers are applied when a property in the logged object matches a property
+in the serialiers. The only exception is the `err` serializer as it is also applied in case
+the object is an instance of `Error`, e.g. `logger.info(new Error('kaboom'))`.
+
 * See [pino.stdSerializers](#pino-stdserializers)
 
 ##### `serializers[Symbol.for('pino.*')]` (Function) - DEPRECATED
--- a/docs/api.md
+++ b/docs/api.md
@@ -258,7 +258,7 @@ should never throw. When logging an object, each top-level property
 matching the exact key of a serializer will be serialized using the defined serializer.
 
 The serializers are applied when a property in the logged object matches a property
-in the serialiers. The only exception is the `err` serializer as it is also applied in case
+in the serializers. The only exception is the `err` serializer as it is also applied in case
 the object is an instance of `Error`, e.g. `logger.info(new Error('kaboom'))`.
 
 * See [pino.stdSerializers](#pino-stdserializers)
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -324,6 +324,7 @@ function createArgsNormalizer (defaultOptions) {
       opts = null
     }
     opts = Object.assign({}, defaultOptions, opts)
+    opts.serializers = Object.assign({}, defaultOptions.serializers, opts.serializers)
     if ('extreme' in opts) {
       throw Error('The extreme option has been removed, use pino.destination({ sync: false }) instead')
     }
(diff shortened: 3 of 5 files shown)


lib/proto.js
```diff
@@ -143,23 +143,20 @@ function setBindings (newBindings) {
...
   const t = this[timeSym]()
   const mixin = this[mixinSym]
-  const objError = _obj instanceof Error
   var obj
 
   if (_obj === undefined || _obj === null) {
     obj = mixin ? mixin() : {}
+  } else if (_obj instanceof Error) {
+    const serializers = this[serializersSym]
+    obj = Object.assign(mixin ? mixin() : {}, serializers.err(_obj))
```
jsumners, line 152: What happens if `serializers: {}` or `serializers: { error (e) {} }`?

YouriT:
Be careful that this PR is bringing a breaking change if people were to leverage current behaviour.

jsumners:
> Be careful that this PR is bringing a breaking change if people were to leverage current behaviour.

Agreed.

lib/proto.js
mcollina: fixed now. PTAL

mcollina:
> Be careful that this PR is bringing a breaking change if people were to leverage current behaviour.

TBH, I would consider that a bug in anybody codebase :(. 

---

@jsumners I think this is almost good to go. I do not see any other way in which we can fix the change without breaking anybody.

YouriT:
> > Be careful that this PR is bringing a breaking change if people were to leverage current behaviour.
> 
> TBH, I would consider that a bug in anybody codebase :(.
> 
> @jsumners I think this is almost good to go. I do not see any other way in which we can fix the change without breaking anybody.

Maybe it deserves a version bump with a breaking change notice in changelog.

mcollina:
I'm torn. I don't see why anybody would not want this change, however we have recently bumped major and I do not want this to sit there for a long cycle.

This _is_ be breaking if folks are relying on the difference between things. However this pops every now and then.. and I do not see why somebody would not want this. In fact, it is number 1 reason why I do not recommend `logger.info(err)`.

lib/proto.js
```diff
@@ -143,23 +143,20 @@ function setBindings (newBindings) {
...
   const mixin = this[mixinSym]
-  const objError = _obj instanceof Error
   var obj
 
   if (_obj === undefined || _obj === null) {
     obj = mixin ? mixin() : {}
-  } else {
-    obj = Object.assign(mixin ? mixin() : {}, _obj)
-    if (!msg && objError) {
+  } else if (_obj instanceof Error) {
```
davidmarkclements, line 150: what versions did you find there was no slow down? I thought instanceof was expensive? Is that no longer true in 12 and 14?

YouriT: `instanceof` was already there. Now there is even a branching being skipped in opposition to the original code.

davidmarkclements:
So the breaking change will be that if someone is using a custom error serializer and then do logger.info(err) the log output will now be different?

jsumners:
> do we need docs for this?
> 
> pino/browser.js serializer API needs to be synced with this

Yes, we need docs. I forgot to block on that 🤦 

> So the breaking change will be that if someone is using a custom error serializer and then do logger.info(err) the log output will now be different?

The change is that if someone does `logger.info(err)` they will now potentially get serialized output instead of nothing.

YouriT:
> The change is that if someone does `logger.info(err)` they will now potentially get serialized output instead of nothing.
AFAIK there isn't nothing at the moment, there is the error logged without passing it through the serializer. Hence `message`, `stack`, `type` and all other keys of the `Error` object are linked to the root of the log entry.

jsumners:
> > The change is that if someone does `logger.info(err)` they will now potentially get serialized output instead of nothing.
> 
> AFAIK there isn't nothing at the moment, there is the error logged without passing it through the serializer. Hence `message`, `stack`, `type` and all other keys of the `Error` object are linked to the root of the log entry.

Typically, `Error.prototype.toString()` only returns the message.

naseemkullah:
There seems to be no documentation with this PR, so I'll just ask here if that is ok:

does this mean we do not have to call the error object `err` anymore for pino to treat it special and serialize it as en error?

naseemkullah:
There seems to be no documentation with this PR, so I'll just ask here if that is ok:

does this mean we do not have to call the error object `err` anymore for pino to treat it special and serialize it as an error?

i.e. after this pr is merged:

```js
const foo = new Error('hello');
const bar = {a: 'b'};
logger.error({foo, bar}, foo.message);
```

will result in `foo` being serialized?

jsumners:
No. This PR addresses `log.error(Error('message'))`.

jsumners:
I thought I asked first?
https___i pinimg com_originals_55_0d_2f_550d2f2403def50331d02c031fae22b0

mcollina:
You did! I'm passing that to @davidmarkclements!

mcollina:
Any objection in shipping this in v6? @jsumners @davidmarkclements @delvedor @watson and possibly any others?

jsumners:
If we use the justification that this fixes expected behavior, then we can do a minor release. Ideally, we'd issue a major out of an abundance of caution. But I am okay either way.

mcollina:
I'm ok either way. We'll see if we can bump the major in Fastify.

delvedor (approved):
I think it can be shipped as a minor, maybe with a note in the docs for showing how to get the old behavior?

mcollina:
> I think it can be shipped as a minor, maybe with a note in the docs for showing how to get the old behavior?

You can't get the old behavior, that's part of the problem.

xamgore:
When `nestedKey` is defined and `Error` is passed to the logger, no fields will be displayed in the output. Does this PR fix it? Do I correctly understand @watson is the only person left to review the code?

```ts
export const logger = pino({
  nestedKey: 'payload',
})

process.on('uncaughtException', final(logger, (err, finalLogger) => {
  finalLogger.error(err, 'uncaughtException')
  process.exit()
}))

throw Error('wow') // { level: 50, pid, hostname, payload: {}, message: "uncaughtException"}
```

mcollina:
Landed in the `next` branch, prepping for v7.

PR #904, opened 2020-09-06
Add timestamp option to browser env, closes #903

### LINKED ISSUE #903: "timestamp" option is ignored in Browser environment
Hi, thank you for the great library.

I've run the following code.

I expected time should become ISO String, but actually the output is still epochTime.

```typescript
import pino from 'pino'

const logger = pino({
  timestamp: () => `,"time":"${new Date(Date.now()).toISOString()}"`,
  browser: {
    asObject: true,
  },
})

logger.info('Running app')
```

image

is this intensional? Thanks in advance.

acro5piano:
This PR makes `timestamp` option work in the same way as the node.js environment.

https://github.com/pinojs/pino/issues/903

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -311,7 +311,7 @@ function getTimeFunction (opts) {
   if (opts.timestamp === false) {
     return nullTime
   }
-  return Date.now
+  return epochTime
 }
 
 function mock () { return {} }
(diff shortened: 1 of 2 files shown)


jsumners (approved):
LGTM

mcollina (approved):
lgtm

acro5piano:
Thanks for merging it!

PR #913, opened 2020-09-19
Add TS types

### LINKED ISSUE #910: Proposal: bundle type definitions in Pino itself
## References
I found some prior talkings about this: https://github.com/pinojs/pino/issues/344

## Background
I am opening this issue since we have currently a problem in the Fastify codebase: https://github.com/fastify/fastify/pull/2550

## TLDR;
Fastify relies on Pino for logging and therefore also Fastify's typing has to reference Pino's one. The issue is that at the present day they aren't maintained by Pino's team and we are not going to include 3rd party typings in Fastify codebase. 

## Possible solution
If the Pino core team agree I can actively maintain typings inside this repo with the help of some other fastify core team members:
- https://github.com/fastify/fastify/pull/2550#discussion_r487243232
- https://github.com/fastify/fastify/pull/2550#discussion_r487271985

 I am opening this issue because far as I can see, we don't have any other viable solution for the issue https://github.com/fastify/fastify/pull/2550

kibertoad:
fixes #910

Note that there is definitely room for improvements test-wise and some explicit `tsd` assertions definitely would be helpful, but I think everything else can be done incrementally - this provides a usable foundation.


kibertoad:
@jsumners I would assume that me and @fox1t would need maintainer permissions if we move forward with the "at least 2 dedicated maintainers for TS" plan with 30 days response SLA. Considering that both of us are already members of fastify org, hopefully this shouldn't be a controversial request.

kibertoad:
If nobody minds, I can also add types to https://github.com/pinojs/pino-std-serializers so that there are no third-party types used altogether (other than official Node ones).

package.json
```diff
@@ -3,9 +3,12 @@
   "version": "6.6.1",
   "description": "super fast, all natural json logger",
   "main": "pino.js",
+  "type": "commonjs",
+  "types": "pino.d.ts",
   "browser": "./browser.js",
   "files": [
```
jsumners, line 9: @mcollina and @davidmarkclements I thought we removed this difficult to maintain property from the `package.json`?

kibertoad: `fastify` uses `.npmignore` instead. If that is the way, I can use same approach here as well.

mcollina: I didn't remember we used this here even. Let's not change it in this PR.

pino.d.ts
```diff
@@ -0,0 +1,832 @@
...
+    }
+
+    interface redactOptions {
+        paths: string[];
+        censor?: string | ((v: any) => any);
+        remove?: boolean;
+    }
+}
+
+declare class SonicBoom extends EventEmitter {
```
mcollina, line 794: This shipped in sonic-boom.

package.json
jsumners: Agreed. Just asking because it got added in `pino-std-serializers` with this types stuff.

pino.d.ts
kibertoad: Changed to use it.

package.json
kibertoad: @jsumners pino-std-serializer currently bundles all the garbage such as travis config, which felt not ideal. I can remove it from there if you feel strongly about it.

jsumners: Looks like we actually settled on `files` 🤦‍♂️ 

https://github.com/pinojs/pino/issues/136

jstewmon (commented):
Several type and interface definitions are not exported, which can be inconvenient for users. Interfaces need to be exported to support module augmentation, and, in general, exporting all type and interface definitions offers convenience to users because they can more easily use them to declare types in their own code that interoperates with the package.

For example, `LevelChangeEventListener` is useful for package users to declare the type of a function which will be passed as a listener.

72636c:
Thanks for this work! Unfortunately the default-exported types in `sonic-boom@1.2.0` are now incompatible with `@types/pino`. Would it be worth patching `@types/pino` and `@types/sonic-boom` to switch to the default export while this is in flux?

```
node_modules/@types/pino/index.d.ts(194,122): error TS2709: Cannot use namespace 'SonicBoom' as a type.
node_modules/@types/pino/index.d.ts(202,57): error TS2709: Cannot use namespace 'SonicBoom' as a type.
```

mcollina:
From the amount of feedbacks we are getting from _just_ changing sonic-boom types, I think we must do this as a semver-major change.

mcollina:
I'll revert in sonic-boom as well. Let's treat them all as semver-major.

pino.d.ts
```diff
@@ -0,0 +1,793 @@
...
+    const version: string;
+
+    type SerializedError = pinoStdSerializers.SerializedError;
+    type SerializedResponse = pinoStdSerializers.SerializedResponse;
+    type SerializedRequest = pinoStdSerializers.SerializedRequest;
+
+    /**
+     * Provides functions for serializing objects common to many projects.
+     */
+    const stdSerializers: {
```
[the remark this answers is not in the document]
kibertoad: Done

kibertoad:
> 
> 
> Several type and interface definitions are not exported, which can be inconvenient for users. Interfaces need to be exported to support module augmentation, and, in general, exporting all type and interface definitions offers convenience to users because they can more easily use them to declare types in their own code that interoperates with the package.
> 
> For example, `LevelChangeEventListener` is useful for package users to declare the type of a function which will be passed as a listener.

Please take a look at the latest version, I exported the namespace, now these definition can be reached from within the namespace. Alternatively I think we can try to get rid of namespace altogether and do it more similarly to how fastify does it - but that would be a way more intrusive and breaking change.

pino.js
```diff
@@ -218,4 +218,16 @@ pino.stdTimeFunctions = Object.assign({}, time)
...
+ * These export configurations enable JS and TS developers
+ * to consumer pino in whatever way best suits their needs.
+ * Some examples of supported import syntax includes:
+ * - `const pino = require('pino')`
+ * - `const { P } = require('pino')`
+ * - `import * as pino from 'pino'`
+ * - `import { P } from 'pino'`
+ * - `import pino from 'pino'`
+ */
+pino.P = pino
```
fox1t, line 231: Why `P` and not `pino`?

pino.d.ts
```diff
@@ -0,0 +1,771 @@
...
+//                 Michel Nemnom <https://github.com/Pegase745>
+//                 Igor Savin <https://github.com/kibertoad>
+// TypeScript Version: 2.7
+
+/// <reference types="node"/>
+
+import { EventEmitter } from 'events';
+import * as pinoStdSerializers from 'pino-std-serializers';
+
+export default P
```
fox1t, line 22: It would be much better to use Fastify approach and have the "infamous tripet" exported in JS and here a default + named export without using namespace at all. It is the only known way we have to support all of the tsconfigurations, faux modules, bundlers and so on ( :( )

fox1t (changes requested):

fox1t:
@kibertoad I think we should go with the fastify apporach even if it takes a longer to be merged. It is surely more stable and better done.

kibertoad:
@fox1t You are right. Since we are going for semver major either way, might as well use the opportunity to refine the API.

pino.js
kibertoad: to preserve naming of original `@types/pino`, but you are right, we don't really need to do that.

mcollina:
After the current incident, it seems that a significant amount of pino users are served by the types in `@types/pino`. Moving them to these will break them badly. I'll need to let this bounce a bit in my head before making a call. By adding types to pino, we will be adding _significant more issue traffic_ that this project currently gets. Logging should be "boring" and work well.

jsumners:
> By adding types to pino, we will be adding _significant more issue traffic_ that this project currently gets. Logging should be "boring" and work well.

Agreed. And increasing the issue load around here is not helpful.

fox1t:
Yea, there are plenty of devs using pino types. I agree we should not break them. If we don't see any benefit from porting them in this package and allign them to fastify ones (triplet and so on) there is no point in breaking anything.

kibertoad:
@mcollina There are multiple directions where we can go, though. If we want to avoid breaking changes, we can, there is nothing stopping us from adopting existing DefinitelyTyped typings verbatim so that there is zero chance to break anything for anyone. And we'll still get a better workflow and more agility for how we address future challenges - when there will be pino 7 on the horizon, we could seize the opportunity and implement breaking changes _then_, cleaning up the types and making them more robust.
And obviously there are two sides to "lots of people use `@types/pino`" as well - it means that any improvements that we make will have greater impact as well, and I am very confident that whatever challenges people may encounter with types, we would be able to address them faster and better than what is currently the case with DefinitelyTyped ad-hoc maintainership.

davidmarkclements:
is anyone still wanting to maintain types in Pino? If not I'll close this in a week or so.

kibertoad:
@davidmarkclements Yes. Is next semver major coming up? Can we get back to getting this merged?

davidmarkclements:
@kibertoad major depends on progress in https://github.com/pinojs/pinole

kibertoad:
@mcollina Can we start prepping by releasing semver major for all the downstream deps on which types pino rely like sonic-boom?

mcollina:
> Can we start prepping by releasing semver major for all the downstream deps on which types pino rely like sonic-boom?

I would wait until pinole is finished.

mcollina:
@kibertoad are you still up to support TS in pino in v7? We are possibly approaching a release after https://github.com/pinojs/pino/pull/1003 lands.

kibertoad:
@mcollina Yup. Is it a good time to start going over downstream deps again, releasing semver majors with types?

mcollina:
Yes, that's the idea. This PR should target the `next` branch.

kibertoad:
@mcollina Some changes leaked from `master` that are not yet in `next` into this PR. Can I create a PR syncing `next` with `master` to simplify things?

package.json
```diff
@@ -87,13 +92,18 @@
...
     "through2": "^4.0.0",
+    "tsd": "^0.15.1",
+    "typescript": "^4.2.4",
     "winston": "^3.3.3"
   },
   "dependencies": {
     "fast-redact": "^3.0.0",
     "fast-safe-stringify": "^2.0.7",
-    "pino-std-serializers": "^3.1.0",
+    "pino-std-serializers": "kibertoad/pino-std-serializers#feat/types",
```
kibertoad, line 102: this will be replaced with released version after it is available

mcollina:
> @mcollina Some changes leaked from `master` that are not yet in `next` into this PR. Can I create a PR syncing `next` with `master` to simplify things?

This should be fixed now.

Note that there is also https://github.com/pinojs/pino/pull/1003 that would need typings.

kibertoad:
@mcollina Sure. Would you like these types added in this PR, or should I wait until #1003 lands and create separate PR then?

test/types/pino.test-d.ts
```diff
@@ -0,0 +1,267 @@
...
+
+const log2: P.Logger = pino({
+    name: "myapp",
+    safe: true,
+    serializers: {
+        req: pino.stdSerializers.req,
+        res: pino.stdSerializers.res,
+        err: pino.stdSerializers.err,
+    },
+});
```
mcollina, line 32: Does this use tabs for indentation? Maybe can we have this conform to the rest?

kibertoad:
Will update the PR today

test/types/pino.test-d.ts
kibertoad: It already uses spaces.

pino.d.ts
```diff
@@ -0,0 +1,872 @@
...
+        dest?: string | number | DestinationObjectOptions | DestinationStream | NodeJS.WritableStream,
+    ): SonicBoom;
+
+    /**
+     * Create an extreme mode destination. This yields an additional 60% performance boost.
+     * There are trade-offs that should be understood before usage.
+     * @param [fileDescriptor]: File path or numerical file descriptor, by default 1
+     * @returns A Sonic-Boom  stream to be used as destination for the pino function
+     */
+    function extreme(fileDescriptor?: string | number): SonicBoom;
```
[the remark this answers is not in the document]
kibertoad: Removed everything referring to extreme mode

package.json
```diff
@@ -76,7 +81,7 @@
     "import-fresh": "^3.2.1",
     "log": "^6.0.0",
     "loglevel": "^1.6.7",
-    "pino-pretty": "^4.1.0",
+    "pino-pretty": "^4.8.0",
```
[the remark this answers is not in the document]
kibertoad: Fixed

kibertoad:
@mcollina Ready for re-review!

package.json
```diff
@@ -19,8 +22,9 @@
     "docs": "docsify serve",
     "browser-test": "airtap --local 8080 test/browser*test.js",
     "lint": "eslint .",
-    "test": "npm run lint && tap --100 test/*test.js test/*/*test.js",
+    "test": "npm run test:types && npm run lint && tap --100 test/*test.js test/*/*test.js",
```
mcollina, line 25: can you move the type tests to the end?

mcollina (changes requested):
Multistream is missing

package.json
kibertoad: done

kibertoad:
@mcollina Typed multistream to the best of my ability based on API docs, could you please check if typing makes sense?

pino.d.ts
```diff
@@ -0,0 +1,876 @@
+// Type definitions for pino 6.3
+// Project: https://github.com/pinojs/pino.git, http://getpino.io
```
mcollina, line 2: I think this header should be changed/updated.

mcollina (approved):
lgtm

mcollina:
@kibertoad I've invited you to this org to make sure you can help maintain these going forward.

PR #924, opened 2020-10-26
Merge error flows. (#923)

### LINKED ISSUE #923: Merge error flows instead of doing a special serialisation handling
I see you are planning to fix the error serialisation in #896 in the upcoming v7.

I submitted a bug a while ago #889 which addressed the loosing of the `message` property when a `msg` is present. Your recommendation was to use `{ err }`. I asked a question there but never got an answer.

I see in #896 you are detecting the error and running the serialiser manually. Why not just wrap the object in `{ err }` if it is an error and merge the 2 flows?

This will have the following advantages:
* less edge case programming
* `message` property is not lost
* logs with errors will be easily detected (especially in stuff like Elasticsearch)

mihai1voicescu:
Merge the 2 error flows:
* if obj is of type Error wrap it in `{ err }`
* if msg not present set to error message (if present)

Update tests. Update API doc

Closes #923

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -148,15 +148,15 @@ function write (_obj, msg, num) {
   if (_obj === undefined || _obj === null) {
     obj = mixin ? mixin() : {}
   } else if (_obj instanceof Error) {
-    const serializers = this[serializersSym]
-    obj = Object.assign(mixin ? mixin() : {}, serializers.err(_obj))
-    // backward compat, set message
-    obj.message = undefined
+    obj = Object.assign(mixin ? mixin() : {}, { err: _obj })
     if (!msg) {
       msg = _obj.message
     }
   } else {
     obj = Object.assign(mixin ? mixin() : {}, _obj)
+    if (!msg && _obj.err) {
+      msg = _obj.err.message
+    }
   }
 
   const s = this[asJsonSym](obj, msg, num, t)
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -143,23 +143,20 @@ function setBindings (newBindings) {
 function write (_obj, msg, num) {
   const t = this[timeSym]()
   const mixin = this[mixinSym]
-  const objError = _obj instanceof Error
   var obj
 
   if (_obj === undefined || _obj === null) {
     obj = mixin ? mixin() : {}
-  } else {
-    obj = Object.assign(mixin ? mixin() : {}, _obj)
-    if (!msg && objError) {
+  } else if (_obj instanceof Error) {
+    const serializers = this[serializersSym]
+    obj = Object.assign(mixin ? mixin() : {}, serializers.err(_obj))
+    // backward compat, set message
+    obj.message = undefined
+    if (!msg) {
       msg = _obj.message
     }
-
-    if (objError) {
-      obj.stack = _obj.stack
-      if (!obj.type) {
-        obj.type = 'Error'
-      }
-    }
+  } else {
+    obj = Object.assign(mixin ? mixin() : {}, _obj)
   }
 
   const s = this[asJsonSym](obj, msg, num, t)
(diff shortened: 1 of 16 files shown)


mcollina (approved):
lgtm, good for me.

lib/proto.js
```diff
@@ -148,15 +148,15 @@ function write (_obj, msg, num) {
...
-    obj = Object.assign(mixin ? mixin() : {}, serializers.err(_obj))
-    // backward compat, set message
-    obj.message = undefined
+    obj = Object.assign(mixin ? mixin() : {}, { err: _obj })
     if (!msg) {
       msg = _obj.message
     }
   } else {
     obj = Object.assign(mixin ? mixin() : {}, _obj)
+    if (!msg && _obj.err) {
```
jsumners, line 157: It seems like `obj = { err: { foo: 'bar' } }` would make it to this branch. Is that correct? If so, is it still acceptable for `msg = undefined`?

mcollina: @mihai1voicescu ping

mihai1voicescu: To reach this branch it should already be set to `undefined` so setting it again to `undefined ` should be OK (I could also check if the message property exists but this should be a rare occurrence).

Also maybe we should check that `msg === undefined` instead of `!msg` because someone may want to pass values like boolean `false` or number `0` for some reason.

What do you guys think?

jsumners: What I'm saying is that you're detecting in `log.info({ err: {} }, 'a message')` if the `'a message'` string is `undefined` and then assuming the object assigned to `err` has a `message` property. What happens with that assumption?

mihai1voicescu: If the assumption (`err has a message property`) proves false, then the msg (`'a message'`) will be set from `undefined` to `undefined`, which should be OK.

Sorry for the late responses, having some problems with the notifications 😞

jsumners: Can we cover that case in the tests?

mihai1voicescu: Sure.

Should I also swap to checking `msg === undefined` instead of `!msg`? Seems more correct.

jsumners: You can try it. If the tests pass without having to change them, I'd accept it. But if they don't, then I'd save that for another PR that targets v7.

mihai1voicescu: Both these changes

> Merge the 2 error flows:
> 
> if obj is of type Error wrap it in { err }
> if msg not present set to error message (if present)

should imply a major API change(v7)

The tests have already changed, because when passing an Error object it now gets transformed to `{ err }`.

The old behaviour was to set the `msg` property to `Error.message` if it was missing (`undefined`) and obj is of type Error. Now the 2 flows merge, this means they can either both default to `Error.message` or none will, which both imply modifying the tests and the behaviour. The correct choice, imho, is to default it to Error.message.

**If I understood correctly you just want to split these 2 changes into 2 PRs**, but both of them imply major API changes, altering the tests and are linked together. Also, splitting them implies making test changes in the first PR just so I can throw them away in the next PR, which seems wasteful.

jsumners: Oh, I forgot this is already targeting v7. Proceed with the change.

jsumners (approved):
LGTM

mcollina (approved):
lgtm

mcollina:
can you please resolve the conflicts so we can land?

mihai1voicescu:
Please note that I replaced 'type should be retained, even if type is a property' test with 'type should be detected based on constructor' as the std err serializer does not seem to take in account type any more, but rather check the constructor name.

Check https://github.com/pinojs/pino/pull/924/files#diff-ae2bafab3d0742ad51d99945e79114e8ad74e0b6dcfe291797b3433f59f8b130R38

Is this OK?

mcollina (approved):
lgtm, good for me

PR #926, opened 2020-11-13
pass merging object to mixins

madelinecameron:
Allows for mixins to take in an object and add information contextually. 

For instance, if your `mergingObject` has an `error` key (a [JS Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)), I want to add `{ message: error.message, stack: error.stack }`. This will make mixins more useful or having to pull from higher in the scope.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -99,7 +99,7 @@ logger.info('hello') // Will throw an error saying info in not found in logger o
 Default: `undefined`
 
 If provided, the `mixin` function is called each time one of the active
-logging methods is called. The function must synchronously return an
+logging methods is called. The first and only parameter is the value `mergeObject` or an empty object. The function must synchronously return an
 object. The properties of the returned object will be added to the
 logged JSON.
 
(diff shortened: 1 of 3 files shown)


madelinecameron:
Will add a test shortly, sorry about that. :)

mcollina (changes requested):

mcollina (approved):
lgtm

mcollina:
This is also missing some docs update

madelinecameron:
Yep, was wondering about that, I'll do that. Thanks!

madelinecameron:
Took a crack at integrating it to documentation, let me know any feedback. :)

docs/api.md
```diff
@@ -124,7 +124,7 @@ const mixin = {
 }
 
 const logger = pino({
-    mixin() {
+    mixin({ description }) {
```
mcollina, line 127: let's name the paramenter `obj`, without destructuring

jsumners (approved):

PR #929, opened 2020-11-16
doc: fix wrong link to logger.level

mmarchini:
Link was pointing to the `level` hook instead of the `level` instance property.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -15,7 +15,7 @@
   * [logger.child()](#child)
   * [logger.bindings()](#bindings)
   * [logger.flush()](#flush)
-  * [logger.level](#level)
+  * [logger.level](#logger-level)
   * [logger.isLevelEnabled()](#islevelenabled)
   * [logger.levels](#levels)
   * [logger\[Symbol.for('pino.serializers')\]](#serializers)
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

docs/api.md
```diff
@@ -15,7 +15,7 @@
   * [logger.child()](#child)
   * [logger.bindings()](#bindings)
   * [logger.flush()](#flush)
-  * [logger.level](#level)
+  * [logger.level](#loggerlevel-string-gettersetter)
```
jsumners, line 18: Actually, according to the manually created identifiers, the code is correct:

https://github.com/pinojs/pino/blob/6c42f14cf9bea30ad4f3460f04662403b4e1fa8a/docs/api.md#L708

The problem is that there are a couple other "level" subheadings prior to the one this TOC link should refer to. These other subheadings do not have explicit identifiers, so the Markdown interpreter is inserting conflicting anchor identifiers.

The true fix here is to go through and add explicit identifiers for all sub-headings that do not have them.

The lazy fix here is to update the explicit identifier on line 708 to something else and adjust the TOC accordingly.

jsumners (changes requested):

docs/api.md
mmarchini: What if we change `<a id="level"></a>` to `<a id="logger-level"></a>` instead? This way we fix the existing link without relying on weak methods (identifier generated by markdown or line identifier). I don't have bandwidth to go through all sub-heading right now (if someone else wants to and has the bandwidth, feel free to pick this up :) ).

jsumners: Either fix is okay. We just need one of them to happen.

mmarchini: Updated to `<a id="logger-level"></a>`

jsumners (approved):
LGTM

PR #943, opened 2020-12-26
Add pino-dev as an ecosystem package in the docs

dnjstrom:
Been working on an alternative prettifier with built-in support for pino-http and pino-debug that is getting stable enough for general use so figured I'd propose adding it to the docs.

Is it correct to assume that this would also update the http://getpino.io/ website? I found this repo https://github.com/pinojs/getpino.io, but it doesn't seem active while this repo has a CNAME file containing `getpino.io`

Thank you for your consideration!


mcollina (approved):
lgtm

jsumners (approved):

Issue #909, opened 2020-09-14
Access log level in logMethod hook


gzaripov:
I want to collect logs number statistics and I want also to categorize them by levels.
For now I cant get level from logMethod hook, so I cant collect info about method types, lets add it to logMethod hook:

```javascript
const hooks = {
  logMethod (inputArgs, method, { level }) {
    metrics.registerEvent(`log_${level}`);
    return method.apply(this, inputArgs)
  }
}
```

mcollina:
It is not needed as you can get it with https://github.com/pinojs/pino/blob/83d80f63888aa0b68479d460706644c58257131c/docs/api.md#levels.

```js
logger.levels.values[method]
```

gzaripov:
@mcollina could you provide exact solution? I try to call `this.levels.values[method]` inside function and I get undefined.
Here is reproduction https://codesandbox.io/s/funny-shannon-e8elg?file=/src/start.js

mcollina:
Actually I was wrong, that's not feasible right now.

What's your use case exactly?

jsumners:
Support would need to be added at: 

https://github.com/pinojs/pino/blob/95856d1198a4067f9916a2b7e636b71f58bebeb2/lib/tools.js#L30-L35

gzaripov:
> What's your use case exactly?

I want to provide more info about what kind of logs we are logging in node server. It will help us to track new problems when error logs count was increased. For now I can track only total amount of logs.

gzaripov:
I suggest this solution:
```
function genLog (level, hook) {
  if (!hook) return LOG

  return function hookWrappedLog (...args) {
    hook.call(this, args, LOG, { level })
  }
```

mcollina:
Please avoid creating a new object, e.g. `level` not `{ level }`.

mojavelinux:
FYI, this appears to have been resolved but never closed.

PR #944, opened 2020-12-30
access log level in hook #909

DoranYun:
For https://github.com/pinojs/pino/issues/909

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -31,7 +31,7 @@ function genLog (level, hook) {
   if (!hook) return LOG
 
   return function hookWrappedLog (...args) {
-    hook.call(this, args, LOG)
+    hook.call(this, args, LOG, level)
   }
 
   function LOG (o, ...n) {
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Good work! Can you please add a unit test?

DoranYun:
> Good work! Can you please add a unit test?

ok.

mcollina (approved):
lgtm

lib/tools.js
```diff
@@ -31,7 +31,7 @@ function genLog (level, hook) {
   if (!hook) return LOG
 
   return function hookWrappedLog (...args) {
-    hook.call(this, args, LOG)
+    hook.call(this, args, LOG, level)
```
jsumners, line 34: @mcollina this will retain a reference to the closure where `level` is defined. Do you foresee any optimization issues with this? Should the function be defined and a property attached to it so it can be `hook.call(this, args, LOG, this.level)`?

mcollina: This is ok for me, the closure context will be kept anyway.

jsumners (approved):
LGTM

PR #946, opened 2021-01-08
Bump up pino-std-serializers dependency

al-bglhk:
It's been several versions since `pino-std-serializers` was aligned with `pino`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "pino",
-  "version": "6.9.0",
+  "version": "6.9.1",
   "description": "super fast, all natural json logger",
   "main": "pino.js",
   "browser": "./browser.js",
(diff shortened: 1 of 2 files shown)


al-bglhk:
I have updated the test case so it will now refer to `request.socket` instead of `request.connection`. And skipped tests that expect `request.connection` to be optional, because it is always available. 

- request.connection has been deprecated since Node13
- request.socket interface is compatible with request.connection since v0.3.0, or at least v10

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "pino",
-  "version": "6.9.0",
+  "version": "6.9.1",
```
mcollina, line 3: please revert this

mcollina:
@jsumners did you recall why we did not bump this?

jsumners:
> @jsumners did you recall why we did not bump this?

Likely because it is a major version increase or we just didn't think about it. But, it being a major version increase, I wonder if this PR should target v7?

mcollina:
It's just a drop of old node.js versions: https://github.com/pinojs/pino-std-serializers/pull/40.

jsumners (approved):

Issue #947, opened 2021-01-08
Update standard to v16


mcollina:
This means solving a few `var` to `let` situation. Note that `for (let i` is slower than `for (var i` and as a result it should be kept in place.

PR #950, opened 2021-01-13
Replacing var with let/const, somewhere left var because scope, somew…

salesh:
#947 
Replacing var with let/const, somewhere left var because scope somewhere left var because "for" + resolved all standard issues


mcollina (approved):
lgtm

mcollina:
Benchmarks seems ok.

benchmarks/utils/runbench.js
```diff
@@ -90,7 +90,8 @@ function runBenchmark (name, done) {
 }
 
 function sum (arr) {
-  var result = 0
+  let result = 0
+  /* eslint no-var: off */
```
jsumners, line 94: If we were to directly use `eslint` with https://github.com/standard/eslint-config-standard instead of the straight `standard` package, could we update our `.eslintrc` to include this rule? I really despise having these comments scattered all over the code just to tweak a development tool. I understand that sometimes they are unavoidable (e.g. `/* istanbul skip else */`), but the fewer we have to have the better.

In general, I'm really annoyed by the latest couple releases of `standard` imposing some really drastic rules. But this "no more var" one is asinine and dangerous.

browser.js
```diff
@@ -1,11 +1,11 @@
 'use strict'
 
-var format = require('quick-format-unescaped')
+const format = require('quick-format-unescaped')
```
jsumners, line 3: @davidmarkclements are there browsers we support that do _not_ support `let` and `const`?

lib/levels.js
```diff
@@ -55,7 +55,8 @@ function genLsCache (instance) {
   const formatter = instance[formattersSym].level
   const { labels } = instance.levels
   const cache = {}
-  for (const label in labels) {
+  /* eslint no-var: off */
+  for (var label in labels) {
```
jsumners, line 59: It's `for (let x = 0; x < 10; x += 1)` that is slower. I believe in this case we can do:

```js
for (const label in labels)
```

Correct, @mcollina ?

pino.js
```diff
@@ -138,10 +138,14 @@ function pino (...args) {
     [stringifySym]: stringify,
     [formattersSym]: allFormatters
   })
-  const chindings = base === null ? '' : (name === undefined)
-    ? coreChindings(base) : coreChindings(Object.assign({}, base, { name }))
+  const chindings = base === null
+    ? ''
+    : (name === undefined)
+        ? coreChindings(base)
+        : coreChindings(Object.assign({}, base, { name }))
```
jsumners, line 145: Ugh. I didn't realize we had nested ternaries in this code. I'd rather eliminate the nesting. The indentation helps, but it's still difficult to reason through.

lib/levels.js
mcollina: Yes, correct.

jsumners: So everywhere we have a `for...in` in this code, please remove the eslint comment and adjust to a `for(const .. in ..)`.

browser.js
mcollina: I would land this and see if anybody complains. IE is finally out of the door, so I would not bother.

benchmarks/utils/runbench.js
salesh: @jsumners tried now

`{
    "extends": "standard",
    "rules": {
        "no-var": "off"
    }
}`

and 

`{
    "extends": "standard",
    "rules": {
        "no-var": 0
    }
}`

But none is working, Am I missing something, did they change something in standard 16? Even tried .eslintrc.json

jsumners: Yes, we would have to remove the `standard` package and add the `eslint` package. Then update the `package.json` scripts accordingly.

pino.js
salesh: Done

lib/levels.js
salesh: Done

benchmarks/utils/runbench.js
salesh: Done

eslintrc.json
```diff
@@ -0,0 +1,17 @@
+{
```
jsumners, line 1: Why was this renamed? Doesn't `eslint` read `.eslintrc`?

eslintrc.json
```diff
@@ -0,0 +1,17 @@
...
+    },
+    "extends": [
+        "standard"
+    ],
+    "parserOptions": {
+        "ecmaVersion": 12
+    },
+    "rules": {
+        "no-var": "off"
+    }
```
jsumners, line 16: Do we really need anything other than the `"extends"` and `"rules"` sections?

package.json
```diff
@@ -63,6 +64,11 @@
     "bole": "^4.0.0",
     "bunyan": "^1.8.14",
     "docsify-cli": "^4.4.1",
+    "eslint": "^7.17.0",
+    "eslint-config-standard": "^16.0.2",
+    "eslint-plugin-import": "^2.22.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-promise": "^4.2.1",
```
jsumners, line 71: Are these necessary?

eslintrc.json
salesh: https://eslint.org/docs/user-guide/migrating-to-7.0.0#runtime-deprecation-warnings-for-eslintrc-config-files

salesh: There are few more options but json is one of suggested

package.json
salesh: Didn't know if you want extra funct. I will remove

eslintrc.json
salesh: Didn't know if you want extra funct. I will remove

package.json
salesh: @jsumners I did an investigation and all of these 3 need to be in devDependencies - won't work without it, this package is dependent on them.

eslintrc.json
salesh: @jsumners ESLint needs parserOptions to know which one should use - **ESLint allows you to specify the JavaScript language options you want to support. By default, ESLint expects ECMAScript 5 syntax** should I leave  "ecmaVersion": 12 ?. Do we need any "global variables that are predefined" https://eslint.org/docs/4.0.0/user-guide/configuring#specifying-environments  or you want me to remove them?

eslintrc.json
jsumners: That link explicitly says to add a `.eslintrc` file to the project. By default, `eslint` will look for `.eslintrc` in the project directory (the parent directory when running the tool). So let's keep it `.eslintrc` and remove the `--config` in the npm script.

eslintrc.json
jsumners: @mcollina what is your opinion of this configuration? It eliminated needing to scatter inline tool directives throughout the source code, but it is more scaffolding than just "install `standard`". Unfortunately, `standard` explicitly ignores extra configuration by design. So we can't disable this rule while only using the `standard` package.

mcollina: I don't really understand why we are moving away from standard. I think it was ok.

jsumners: We still use the standard rule set, but gain the ability to disable its `var` rule in one place instead of having to do it on every line where we want to keep using `var`.

mcollina: Ok!

mcollina (approved):
lgtm

eslintrc.json
jsumners: Alrighty. @salesh let's rename the rc back to what it was and update the npm script. Then I'm good with this PR.

jsumners (approved):
LGTM

PR #928, opened 2020-11-16
address destructuring mergeObject in docs

madelinecameron:
Follow-on to https://github.com/pinojs/pino/pull/926 to address [this comment](https://github.com/pinojs/pino/pull/926#discussion_r523279100) by @mcollina: 

> let's name the paramenter obj, without destructuring

PR #857, opened 2020-06-03
Issue 855 clarify mixin docks

baterson:
Hello! This is a clarification for the problem from https://github.com/pinojs/pino/issues/855

PR #952, opened 2021-01-15
docs: fix mixin docs

### LINKED ISSUE #951: readme mixin doc: wrong sample code
The section on mixins in the readme has the following code:

```js
const mixin = {
    appName: 'My app'
}

const logger = pino({
    mixin(obj) {
        return {
          description: obj.description
        }
    }
})

pino.info({
    description: 'Ok'
}, 'Message 1')
// {"level":30,"time":1591195061437,"pid":16012,"hostname":"x","appName":"My app","description":"Ok" "msg":"Message 1"}
pino.info('Message 2')
// {"level":30,"time":1591195061437,"pid":16012,"hostname":"x","appName":"My app","description":"Ok","msg":"Message 2"}
// Note: the second log contains "description":"Ok" text, even if it was not provided.
```

but (after fixing the last two `pino` references to `logger`), the prophecy in the logs doesn't actually come to be.

What was this bit of code intended to illustrate?

https://github.com/pinojs/pino/blob/master/docs/api.md#mixin-function

hraban:
Restores the original semantics introduced by afb639c (#857).

Reverts the changes to this documentation from 6e50d72 (#928) and
6c42f14 (#926).

Fixes #951.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -124,18 +124,16 @@ const mixin = {
 }
 
 const logger = pino({
-    mixin(obj) {
-        return {
-          description: obj.description
-        }
+    mixin() {
+        return mixin;
     }
 })
 
-pino.info({
+logger.info({
     description: 'Ok'
 }, 'Message 1')
 // {"level":30,"time":1591195061437,"pid":16012,"hostname":"x","appName":"My app","description":"Ok" "msg":"Message 1"}
-pino.info('Message 2')
+logger.info('Message 2')
 // {"level":30,"time":1591195061437,"pid":16012,"hostname":"x","appName":"My app","description":"Ok","msg":"Message 2"}
 // Note: the second log contains "description":"Ok" text, even if it was not provided.
 ```


mcollina (approved):
lgtm, thanks!

jsumners (approved):
Honestly, I just have to trust you here. I really don't know much about the "mixin" feature.

LGTM

hraban:
> Honestly, I just have to trust you here. I really don't know much about the "mixin" feature.
> 
> LGTM

Fair. Without getting into the specific feature, at all, on a more basic level: this patch actually outputs what the comments say it should. That's what the original patch did, too. The two subsequent patches were a regression in those terms.

I think the confusion came from the point that this snippet, specifically, is about what _not_ to do. The next patches seemed to be intended to give instructions on how to use mixins. Which breaks the expected output.

For posterity: to test this patch (and the state before it): run the snippet locally. You'll see what happens.

PR #953, opened 2021-01-16
docs: clarify options.timestamp usage

cmawhorter:
I mistook the meaning of `,"timestamp"...` and description and was wondering why my log entries looked like this: `{"level":302021-01-16T22:08:04.469Z,"pid":11224, ...`.

the way this works was a little surprising but i'm sure makes sense for perf reasons.


mcollina (approved):
lgtm

Yes, it's for perf reason as it is one of the most sensitive areas of the code.

jsumners (approved):

PR #970, opened 2021-02-16
add support for bigint

alexey-sh:


jsumners (changes requested):
This is at a minimum missing tests.

mcollina:
I'm not sure we should land this as JSON.stringify() does not support this correctly either.

alexey-sh:
>I'm not sure we should land this as JSON.stringify() does not support this correctly either.

I'm not sure there is a way to make it faster than JSON.stringify.

alexey-sh:
>This is at a minimum missing tests.

I added some tests. Please have a look

test/basic.test.js
```diff
@@ -509,6 +509,51 @@ test('handles objects with null prototype', async ({ same }) => {
...
+    hostname,
+    level: 30,
+    bigInt: { a: { b: '1n' } }
+  })
+})
+
+test('object and format bigInt property', async ({ same }) => {
+  const stream = sink()
+  const instance = pino(stream)
+  instance.info({ answer: 42n }, 'foo %s', 1n)
```
jsumners, line 545:
```suggestion
  instance.info({ answer: 42n, foo: { bar: BigInt(Number.MAX_SAFE_INTEGER) + 10n } }, 'foo %s', 1n)
```

test/basic.test.js
```diff
@@ -509,6 +509,51 @@ test('handles objects with null prototype', async ({ same }) => {
...
+  const result = await once(stream, 'data')
+  delete result.time
+  same(result, {
+    pid,
+    hostname,
+    level: 30,
+    msg: 'foo 1',
+    answer: '42n'
```
jsumners, line 553:
```suggestion
    answer: '42n',
    foo: { bar: '9007199254741001n' }
```

test/basic.test.js
```diff
@@ -509,6 +509,51 @@ test('handles objects with null prototype', async ({ same }) => {
...
+  const stream = sink()
+  const instance = pino(stream)
+  instance.info({ answer: 42n }, 'foo %s', 1n)
+  const result = await once(stream, 'data')
+  delete result.time
+  same(result, {
+    pid,
+    hostname,
+    level: 30,
+    msg: 'foo 1',
```
jsumners, line 552: Why doesn't this have the `n` suffix?

jsumners:
<details>
<summary>Current master branch bench</summary>

```
Running DEEP-OBJECT benchmark

benchPinoDeepObj*10000: 3801.850ms
benchPinoAsyncDeepObj*10000: 3609.519ms
benchPinoNodeStreamDeepObj*10000: 4111.749ms
```

</details>

<details>
<summary>PR branch bench</summary>

```
Running DEEP-OBJECT benchmark

benchPinoDeepObj*10000: 4437.022ms
benchPinoAsyncDeepObj*10000: 4350.699ms
benchPinoNodeStreamDeepObj*10000: 4622.513ms
```

</details>

lib/tools.js
```diff
@@ -396,7 +399,12 @@ function final (logger, handler) {
 
 function stringify (obj) {
   try {
-    return JSON.stringify(obj)
+    return JSON.stringify(obj, (key, value) => {
+      if (typeof value === 'bigint') {
+        return value.toString() + 'n'
+      }
+      return value
+    })
```
mcollina, line 407: This is going to slow us down significanfly. I'm -1 to us landing this feature.

mcollina (changes requested):

lib/tools.js
jsumners: The deep object benchmarks confirm this.

test/basic.test.js
alexey-sh: tbh idk. is it important? would you like me to add the suffix?

lib/tools.js
alexey-sh: >This is going to slow us down significanfly.

have you checked the case then JSON.strigify throws an error? I bet it goes down a lot.

>I'm -1 to us landing this feature.

and -{a lot of users} because luck of bigint support.

––––

I guess it's possible to cofigure pino to skip bigint check. does it sound fine for you guys?

test/basic.test.js
jsumners: The other outputs in this PR have the suffix. I'm asking about it from a consistency standpoint. How will an interpreter know that it is a `BigInt` without the suffix?

lib/tools.js
jsumners: Thanks for the pull request. We have had a discussion and have determined that `BigInt` needs better support in general. This implementation negatively impacts our benchmarks even when a `BigInt` is not present.

Issue #978, opened 2021-03-04
First class doc about arguments of logging methods


glensc:
Currently, in [https://github.com/pinojs/pino#usage](https://github.com/pinojs/pino/tree/ca600fc423fdf402ab67ae4bbd63e76db36cdd8e#usage) there is not even covered how non-string arguments and objects must be passed.

So I wrote:
```ts
logger.error("Here be dragons: ", e);
logger.error("Dragons be gone", { e });
```

But this (with prettier) results in:

```
[1614854677187] ERROR (75398 on rocinante): Here be dragons:
[1614854677190] ERROR (75398 on rocinante): Dragons be gone
```

I was expecting it to behave like [`console.log`](https://developer.mozilla.org/en-US/docs/Web/API/Console/log#syntax):
```js
console.log(obj1 [, obj2, ..., objN]);
console.log(msg [, subst1, ..., substN]);
```

however, seems pino expects object (only one?) to be specified as first argument:

```ts
logger.error(e, "Here be dragons: ");
logger.error({ e, f: e }, "Dragons be gone", "forever");
```

Even the code above itself seems oddly reversed,
it outputs like the console.log style code was shown above:

```
[1614854750142] ERROR (75439 on rocinante): Here be dragons:
    Error:
        at logger.test.ts:6:11
        ...
[1614854948285] ERROR (75537 on rocinante): Dragons be gone
    e: {}
    f: {}
```

mcollina:
> I was expecting it to behave like console.log:

What did you make it expect so?

FYI, this has not parity with `console.log()`, it's a different library.

jsumners:
The "Usage" section of the main Readme is an example. Do you expect the full documentation there? The full documentation is at https://github.com/pinojs/pino/blob/ca600fc423fdf402ab67ae4bbd63e76db36cdd8e/docs/api.md#logger

glensc:
@jsumners hence the issue, I read frontpage I see a partial example I go away. There's not even 'read more' link right now. Maybe it's cleverly hidden, so people don't go away? start using it and then find the awkward api, and then it's too late to go away?

@mcollina why would I expect something else? People start with `console.log` in their code, and later find need more control where the logs go, so start looking for frameworks. other logging frameworks. I've come contact to, follow a similar principle, first is message and then optional arguments, arguments can be objects.

In my opinion, the example of library readme use must include example of how to print any objects and `Error` objects.

jsumners:
Pull requests that improve documentation are always welcome.

[...]

PR #979, opened 2021-03-04
doc/api: update consequtive parameters hook to take all arguments

glensc:
the existing example accounted for only two arguments.

this makes the method work more like `console.log(a,b,c,...);`

refs:
- https://github.com/pinojs/pino/issues/978

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -544,13 +544,18 @@ const pinoOptions = {
 }
 
 function logMethod (args, method) {
-  if (args.length === 2) {
-    args[0] = `${args[0]} %j`
+  if (args.length >= 2) {
+    args[0] = args[0] + " %j".repeat(args.length - 1);
   }
   method.apply(this, args)
 }
 
 const logger = pino(pinoOptions)
+```
+
+```js
+logger.info('hello', 'world', 'and', 'hola mundo')
+// {"level":30,"time":1614876433202,"pid":92820,"hostname":"x","msg":"hello 'world' 'and' 'hola mundo'"}
 ```
 
 * See [`message` log method parameter](#message)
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -544,15 +544,20 @@ const pinoOptions = {
 }
 
 function logMethod (args, method) {
-  if (args.length === 2) {
-    args[0] = `${args[0]} %j`
+  if (typeof args[0] === "string" && args.length >= 2) {
+    args[0] += ", %j".repeat(args.length - 1);
```
jsumners, line 548: This does not match the lead in text. The lead in very specifically shows two arguments.

glensc: I don't understand your response. What is lead?

Also, I'm not aware of previous pino versions of what the original snippet tried to solve.

jsumners: https://dictionary.cambridge.org/us/dictionary/english/lead-in

"something that introduces something else, such as the words and music that are used to introduce a television program"

The example you are modifying in this PR is not in isolation. The text before it describes what is being shown in the example. The change in this PR makes this example not match the lead-in text.

glensc: The lead in text you indicate is following:
> Since pino v6, we do not automatically concatenate and cast to string consecutive  parameters:
>
> logger.info('hello', 'world')
>

there's nowhere written in there that only two parameters must exist, consecutive means unspecified amount. so the code **does** match!

@jsumners what do you want from me I don't understand?

jsumners: The example is showing how one would replicate a very specific case that worked prior to v6. It is not a general example for a variadic implementation.

I see two possibilities:

1. Drop the change
2. Finish the PR by making the lead-in text match the example

Issue #983, opened 2021-03-11
After upgrade the msg field is missing from the logs


budarin:
After last upgrade pino deps have noticed that msg field is missing from all my logs

I do log like this:
`logger.info({ msg: 'Log something...', params, options })`

I see in loki
`{..., msg: "", params, options }`

my pino configuration
```
const formatters = {
    level(label) {
        return { level: label };
    },
};
const options = {
    formatters,
    ...(__DEV__ && !CI && prettyPrint),
    level: "info",
    base: null,
};

 let logger!;

if (__DEV__) {
    logger = pino(options, pino.destination(1));
}

if (__PROD__) {
    logger = pino(
        options,
        pino.destination({
            minLength: 4096,
            sync: false,
        }),
    );

    setInterval(function () {
        logger.flush();
    }, logFlushInterval).unref();
}

module.exports = logger;
```

"pino": "^6.11.1",
"pino-pretty": "^4.7.1",

What is changed ? I have not found any in releases about it

jsumners:
What was the prior version and the version you upgraded to?

budarin:
@jsumners 

"pino": "^6.11.1",
"pino-pretty": "^4.7.1",

judging by the logs, the versions of pink-pretty have changed, but this module is not enabled in prod mode, although in dev mode is also `msg=""`

jsumners:
What does `pino-pretty` have to do with this issue? Which version of `pino` were you using before the upgrade that worked as expected?

budarin:
![image](https://user-images.githubusercontent.com/8055157/110851373-fffcf900-82c1-11eb-9193-df3ae81dd055.png)

budarin:
I noticed a missing one today, and the latest pino updates were only in the pino-pretty part for the last 3 days

jsumners:
I am unable to help you if you cannot provide answers to my questions. Showing me screenshots of logs does not tell me the version of the libraries you were using that worked as you expected. Without that information, I cannot begin to determine any cause.

budarin:
I need time to investigate what changes and when led to this - I will roll back in git to find when everything was ok

budarin:
@jsumners 
while tracing code  `pino/lib/tools.js`
```
function LOG (o, ...n) {
    if (typeof o === 'object') {
      let msg = o
      if (o !== null) {
        if (o.method && o.headers && o.socket) {
          o = mapHttpRequest(o)
        } else if (typeof o.setHeader === 'function') {
[...]

PR #984, opened 2021-03-12
Pin quick-format-unescaped

jsumners:
This is a temporary fix for #983 until a resolution is provided in the upstream library. Right now, new installs of Pino that have not locked the dependency tree before two days ago will be broken if this PR is not shipped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "fast-safe-stringify": "^2.0.7",
     "flatstr": "^1.0.12",
     "pino-std-serializers": "^3.1.0",
-    "quick-format-unescaped": "^4.0.1",
+    "quick-format-unescaped": "~4.0.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
     "fast-safe-stringify": "^2.0.7",
     "flatstr": "^1.0.12",
     "pino-std-serializers": "^3.1.0",
-    "quick-format-unescaped": "^4.0.1",
+    "quick-format-unescaped": "~4.0.1",
```
jsumners, line 97:
```suggestion
    "quick-format-unescaped": "4.0.1",
```

mcollina (approved):
lgtm

mcollina:
can you open a repro upstream?

jsumners:
> can you open a repro upstream?

Yeah.

PR #987, opened 2021-03-18
docs: make clearer pino.stdTimeFunctions.isoTime is using UTC

giuscri:


docs/api.md
```diff
@@ -947,7 +947,7 @@ The `pino.stdTimeFunctions` object provides a very small set of common functions
 * `pino.stdTimeFunctions.epochTime`: Milliseconds since Unix epoch (Default)
 * `pino.stdTimeFunctions.unixTime`: Seconds since Unix epoch
 * `pino.stdTimeFunctions.nullTime`: Clears timestamp property (Used when `timestamp: false`)
-* `pino.stdTimeFunctions.isoTime`: ISO 8601-formatted time in UTC
+* `pino.stdTimeFunctions.isoTime`: ISO 8601-formatted time **in UTC** (`process.env.TZ` will be ignored)
```
jsumners, line 950: What is `process.env.TZ` and why would it have any effect?

giuscri: > In POSIX systems, a user can specify the time zone by means of the TZ environment variable.

https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html

jsumners: Yes, but what does that have to do with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString ? The method should be fairly clear that it maps to the standard `Date.toISOString`.

giuscri: > The method should be fairly clear that it maps to the standard Date.prototype.toISOString.

Why you think it's the case? The docs are talking only about ISO 8601 strings, but not about the mapping to `Date.prototype.toISOString`

jsumners: Because we are talking about JavaScript datetimes and the only method there is as I linked. I think the real change here is to make it clear that the supporting method is `Date.prototype.toISOString`.

davidmarkclements: agreed - JS does not have any notion of TZ env var, there's probably date formatting env settings that are ignored on windows too. We can point out that this use toISOString and that covers all possible implicit things that it doesn't support.

PR #999, opened 2021-04-01
Update to tap v15

jsumners:
The alias deprecations was not fun. Refactoring all of the destructured `is` to `equal` was a nightmare (looking at you @davidmarkclements 😒 ). Doing a find/replace all on `t.is` was far easier.


mcollina (approved):
lgtm

Not fun at all. There are like 50-something projects that I mantain that needs attention... it's going to be tough!

PR #1000, opened 2021-04-06
docs: fix Event: levelChange example

laSinteZ:
Right now, example in docs regarding levelChange event is incorrect. It would display "newValue was changed to oldValue", which is not desired I believe.


jsumners (approved):
LGTM

mcollina (approved):
lgtm

PR #1003, opened 2021-04-07
pino.transport()

mcollina:
This add  `pino.transport()` based on [`thread-stream`](http://npm.im/thread-stream).

In theory, this supersedes #937 as we internalize the pinole part.

__Supports Node v12.20.0 onwards__.

Things that are missing:

* [ ] Docs
* [x] transport example/boilerplate
* [x] Support for transport packages
* ~~Support for pure function transports (is this actually needed?)~~
* [x] More tests
* [ ] pino-specific benchmarks
* [x] test support for transport packages on windows


jsumners:
> * Support for pure function transports (is this actually needed?)

I think so.

lib/transport.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const { createRequire } = require('module')
+const caller = require('get-caller-file')
```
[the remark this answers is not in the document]
mcollina: This is what pinole used before. I don't think there is a different way to achieve that.

lib/transport.js
```diff
@@ -7,11 +7,12 @@ const { isAbsolute } = require('path')
 const ThreadStream = require('thread-stream')
 
 function transport (filename, workerData) {
+  /* istabul ignore start */
```
jsumners, line 10:
```suggestion
  /* istanbul ignore else */
```

lib/transport.js
```diff
@@ -7,11 +7,12 @@ const { isAbsolute } = require('path')
...
   if (!(isAbsolute(filename) || filename.indexOf('file://') === 0)) {
     const callerRequire = createRequire(caller())
     filename = callerRequire.resolve(filename)
-    console.log(filename)
   }
+  /* istabul ignore end */
```
jsumners, line 15:
```suggestion
```

lib/transport.js
```diff
@@ -0,0 +1,28 @@
+'use strict'
+
+const { createRequire } = require('module')
+const caller = require('get-caller-file')
+const { isAbsolute } = require('path')
+
+const ThreadStream = require('thread-stream')
+
+function transport (filename, workerData) {
+  /* istabul ignore start */
```
jsumners, line 10:
```suggestion
  // Unreachable on Windows, so we ignore coverage here.
  /* istanbul ignore start */
```

lib/transport.js
```diff
@@ -0,0 +1,28 @@
...
+  if (!(isAbsolute(filename) || filename.indexOf('file://') === 0)) {
+    const callerRequire = createRequire(caller())
+    filename = callerRequire.resolve(filename)
+  }
+  /* istabul ignore end */
```
jsumners, line 15:
```suggestion
  /* istanbul ignore end */
```

mcollina:
> > * Support for pure function transports (is this actually needed?)

> I think so. 

Why is that so? I'm not really understanding the need of it.
Processing one log line at a time is far from fast and it will generate a significant build-up in memory.

jsumners:
> Processing one log line at a time is far from fast

I suppose I didn't quite follow what was meant by the item.

mcollina:
pinole expects a transport to be a pure function: https://github.com/pinojs/pinole/blob/f347c04d4a6c61e05e6ac821e510548aab928458/lib/worker.mjs#L18-L28.
Then it uses `await` to process every incoming chunk of data one at a time.

package.json
```diff
@@ -87,7 +87,8 @@
     "tap": "^15.0.1",
     "tape": "^5.0.0",
     "through2": "^4.0.0",
-    "winston": "^3.3.3"
+    "winston": "^3.3.3",
+    "transport": "./test/fixtures/transport"
```
jsumners, line 91: What is this line for?

mcollina: The transport logic supports loading modules from the scope of the parent. Instead of calling `npm link` using `spawnSync` (which did not work on Windows), I'm just adding it as a devDependency here to avoid doing anything in the test.

lib/transport.js
```diff
@@ -0,0 +1,29 @@
+'use strict'
+
+const { createRequire } = require('module')
+const caller = require('get-caller-file')
+const { isAbsolute } = require('path')
+
+const ThreadStream = require('thread-stream')
+
+function transport (filename, workerData) {
+  if (!(isAbsolute(filename) || filename.indexOf('file://') === 0)) {
```
davidmarkclements, line 10: also support declaring transports per log level 

`transport({ error: 'transport-a', info: 'transport-b'})` this needs more thought on how to support various configs (e.g. error and below etc)

jsumners: That's PR #1004.

davidmarkclements: yeah, but I think we can make it easier by using multistream internally here. It can still be provided separately (as is current), but we can improve this API so that it's opaque for most cases

package.json
```diff
@@ -87,14 +87,17 @@
     "tap": "^15.0.1",
     "tape": "^5.0.0",
     "through2": "^4.0.0",
+    "transport": "./test/fixtures/transport",
```
davidmarkclements, line 90: not a fan of adding this noise here, I know it's not a mistake, but it *looks* like a mistake. 

an alternative is to pass a `--loader` flag to the worker thread that points to test/fixtures/loader.mjs which would then resolve `transport` to /test/fixtures/transport.js within the worker thread

test/fixtures/transport/index.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
```
davidmarkclements, line 2: can we test for an esm transport as well (just call it index.mjs)

test/fixtures/transport/index.js
```diff
@@ -0,0 +1,12 @@
...
+const fs = require('fs')
+const { once } = require('events')
+
+async function run (opts) {
+  const stream = fs.createWriteStream(opts.dest)
```
davidmarkclements, line 7: @mcollina and I spoke on the phone, we'd like to improve transport definition further by allowing for auto line splitting and parsing for simpler transport implementations (every transport is solving the same problem, let's solve it here if possible). 

We discussed `for await of` but there's concerns that this serial processing can saturate worker thread

lib/transport.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+const { createRequire } = require('module')
```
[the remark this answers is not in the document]
davidmarkclements: nice catch

lib/transport.js
davidmarkclements: I'd also like to see something like:

`pino(pino.transport(config), pino.transport(anotherConfig))`

which allows for multiple worker threads 

So with multistream, plus the above there's ultimate flexibility

davidmarkclements: I think this could open the door for intense - say trace log - processing in a dedicated worker thread while another worker thread could handle the usual use cases

lib/transport.js
```diff
@@ -0,0 +1,29 @@
...
+  if (!(isAbsolute(filename) || filename.indexOf('file://') === 0)) {
+    const callerFile = caller()
+    const callerRequire = createRequire(callerFile)
+    filename = callerRequire.resolve(filename)
+  }
+
+  const stream = new ThreadStream({
+    filename,
+    workerData,
+    sync: true // TODO should this be configurable?
```
davidmarkclements, line 19: should other worker thread options be configurable (e.g. supplying a --loader flag)

package.json
mcollina: I prefer not to rely on experimental features.  This way seemed the most "black box" testing that I could imagine.
The previous solution of `spawnSync` did not work on Windows.

lib/transport.js
mcollina: I think the best thing we could do is to support:

```
var streams = [
  { level: 'info', destination: 'stdout' },
  { level: 'warn', module: 'pino-socket', opts: { ... } },
  { level: 'error', module: 'pino-elasticsearch', opts: { ... } }
]

pino.transport(streams)
```

Which will create a new thread with all those transports configured.

lib/transport.js
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  stream.on('ready', function () {
+    stream.unref()
+  })
+
+  return stream
```
davidmarkclements, line 26: can we support returning an async function as well a stream, the function takes an a log line object and is called for each log line object

```js
return async (parsedLogline) {
  // do something with the log line 
}
```

test/fixtures/transport/index.js
mcollina: https://github.com/pinojs/pino-abstract-transport should solve your requirements Dave.

lib/transport.js
mcollina: I don't think this should be supported at this level. However I wrote https://github.com/pinojs/pino-abstract-transport which handles all the cases.

lib/transport.js
mcollina: now is fixed, check it out

package.json
```diff
@@ -95,6 +95,7 @@
     "fast-safe-stringify": "^2.0.7",
     "flatstr": "^1.0.12",
     "get-caller-file": "^2.0.5",
+    "pino-abstract-transport": "^0.1.0",
```
jsumners, line 98: This commit adds this dependency but doesn't use it? I'd also prefer we bump it to 1.0.0 before shipping.

mcollina: I'm working on it, I committed it too soon.

mcollina:
@jsumners @davidmarkclements I've just pushed some WIP code on how to create a multitransport worker. I'm not sure that the API we exposed to the end user is good - feedbacks are welcome.

package.json
davidmarkclements: can we at least modify the package.json dynamically rather than check this in?

test/fixtures/transport/index.js
davidmarkclements: 1. doesn't this hit the same issues
2. I'd rather have an standard API in core - this will also making documenting it much easier

lib/multitransport/worker.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+const pino = require('../../pino.js')
+const build = require('pino-abstract-transport')
+
+module.exports = async function ({ transports }) {
+  transports = await Promise.all(transports.map(async (t) => {
```
davidmarkclements, line 7: what about error handling?

suggest `Promise.allSettled` so that multiple failures can be surfaced rather than the first to error

lib/multitransport/index.js
```diff
@@ -0,0 +1,8 @@
+'use strict'
```
davidmarkclements, line 1: I think we can just called this transport.js ?

and the worker, lib/transport-worker.js

lib/transport.js
davidmarkclements: what about:

```js
pino({
  info: pino.destination(1),
  warn: pino.transport('pino-socket'),
  error: [pino.transport('pino-socket'), pino.transport('pino-elasticsearch')]
})
```

davidmarkclements: I know there's logistical issues here but I'm thinking in terms of nicest API first and working backwards

davidmarkclements: This solves the logistics and makes it nicer - `pino.transport` can return a "transport group" - which is a function that registers transports to run in the same worker thread (hence group):

```js
const transport = pino.transport()
pino({
  info: 1,
  warn: transport('pino-socket', opts),
  error: [transport('pino-socket', opts), transport('pino-elasticsearch', moreOpts)]
})
```

davidmarkclements: For putting levels together:

```
const transport = pino.transport()
pino({
  '>=30': 1,
  20: transport('pino-socket', opts),
  '<20': [transport('pino-socket', opts), transport('pino-elasticsearch', moreOpts)]
})
```

davidmarkclements: The beauty here is also that if you pass arguments to `pino.transport` it can work as currently defined - return a stream that can be passed to `pino` like pino.destination does

lib/multitransport/worker.js
bmeck: wouldn't this mean that long pending errors hold the rest in memory? (probably fine)

lib/multitransport/index.js
mcollina: Well, we have `transport()` already in this PR. are you recommending we merge the two things? It's going to be possible!

lib/multitransport/worker.js
mcollina: I tend to side with @bmeck here.

davidmarkclements: in that it's probably fine or in that you don't want to long pending errors to rest in memory?

it's a tension between devex and memory usage then, I guess it depends how bad the memory usage can get vs the value of someone seeing multiple errors in one run rather than fixing one issue to find another etc (especially given that these sorts of things may not be easily reproduced, and more errors in production would be valuable)

lib/multitransport/index.js
davidmarkclements: yes. it's trivium but I think we should keep lib flat

test/fixtures/transport/index.js
davidmarkclements: @mcollina bump :point_up:

test/fixtures/transport/index.js
davidmarkclements: I'll take this action

package.json
davidmarkclements: I'll take this action

lib/transport.js
davidmarkclements: @mcollina thoughts? :point_up:

lib/multitransport/worker.js
mcollina: I think the output for multiple errors would be really confusing. We `throw` the first, so it gets logged to stderr via `console.error` later on.

lib/multitransport/index.js
mcollina: This is done.

package.json
mcollina: Great!

test/fixtures/transport/index.js
mcollina: 1. it does not have the same issue because the user can do as they like. They can use `for await`, they can use streams, essentially anything.
2. It has to be a separate module to avoid dependency cycles. Moving it to a separate package avoids the transports having a direct dependency on pino.

test/fixtures/transport/index.js
mcollina: we are already testing an esm transport.

lastthyme (approved):

lib/transport.js
davidmarkclements: fair enough. - it also matches how pino.destination works

lib/multitransport/worker.js
davidmarkclements: you can throw a single `AggregateError` instead - since it's a native Error constructor familiarity with multiple errors as standardized through `AggregateError` will only increase

jsumners: > you can throw a single `AggregateError` instead - since it's a native Error constructor familiarity with multiple errors as standardized through `AggregateError` will only increase

I had no idea this exists. According to https://node.green it is only available from Node 15.14 and above.

mcollina:
@jsumners @davidmarkclements this is almost done. Everything works when sending the data to pino-elasticsearch with https://github.com/pinojs/pino-elasticsearch/pull/82.

lib/multitransport/worker.js
davidmarkclements: it can be polyfilled https://www.npmjs.com/package/es-aggregate-error

jsumners:
> @jsumners @davidmarkclements this is almost done. Everything works when sending the data to pino-elasticsearch with [pinojs/pino-elasticsearch#82](https://github.com/pinojs/pino-elasticsearch/pull/82).

Exellent. I'm going to put this in draft mode so I can see when it is truly ready for re-review.

davidmarkclements:
I think we drop the "Support for pure function transports (is this actually needed?)" and do that separately (if at all)

lib/multitransport/worker.js
davidmarkclements: let's punt this to a next iteration for now, if demand for better error handling becomes apparent I think that's the way to do it.

test/fixtures/transport/index.js
davidmarkclements: Let's punt this to a future iteration (if at all) so we can get this landed

package.json
davidmarkclements: looks like it works without needing this in `package.json` at all

davidmarkclements: scratch that, I'm an idiot

davidmarkclements:
@mcollina I've modified the API slightly, let me know what you think (https://github.com/pinojs/pino/pull/1003/commits/21abdac531a83ab64bf1d86d28f1df8a2f34f041)

`pino.transport(name, opts)` -  this is a more normative API and it feels cleaner to document. I get the workerData, workerOpts param trade-offs but I think that we should have `opts.workerData` for specific workerData and special case `opts.destination` to create a convention for transports that matches multitransport destination (either that or rename multitransport `t.destination`  to `t.dest`)


I've also added a test for error handling, such as it is. I think we need to propagate the error to the main thread (in thread stream, issue here https://github.com/pinojs/thread-stream/issues/20) and then update that error handling test (and fixture)

jsumners:
> `pino.transport(name, opts)`

@davidmarkclements @mcollina would you guys be opposed to making the signature `pino.transport({ name, opts })`? It'll give us named parameters and flexibility for expansion in the future and avoid the mystery/complexity of positional parameters.

davidmarkclements:
@jsumners good suggestion, I have no issue with it. @mcollina  ?

mcollina:
> @jsumners good suggestion, I have no issue with it. @mcollina ?

I have none.

However how do you implement the multitransport case?

```js
pino.transport({
  workerData: { hello: 'world' },
  destinations: [{ 
    level: 'warn',  
    destination: file
  }, { 
    level: 'info',  
    module: '/Users/matteo/repositories/pino-elasticsearch/lib.js',  
    opts: {    node: 'http://localhost:9200'  }
  }, {  
    level: 'info',  
    prettyPrint: true
  }]
})

**Comment by jsumners** [comment]
> However how do you implement the [multitransport case](https://github.com/pinojs/pino/pull/1003/files#diff-9939b0546354fb8ac6901a43bfc6d8d8e5fc161c00f4b89733dc3edcaf2ab2e9R49-R51)?

Looking at the code, my suggestion will _already_ remove complexity introduced by positional parameters. The `.transport()` code is already having to inspect each parameter to determine its shape and what purpose each serves. Make it a singular object parameter and we get:

```js
function transport ({
  transportScript: 'file://...',
  workerData: { hello: 'world' },
  destinations: [],
}) {}

function multitransport (input) {
  if (Array.isArray(input)) {
    return transport({
      transportScript: './worker.js',
      transports: input // or whatever
    })
  }
  
  return transport(input)
}
```

mcollina:
I've pushed a commit that fixes support for v12. I'm going to work on the API now.

mcollina:
I've refactored the API. Let me know what you think!

test/transport.test.js
```diff
@@ -18,7 +18,10 @@ test('pino.transport with file', async ({ same, teardown }) => {
     os.tmpdir(),
     '_' + Math.random().toString(36).substr(2, 9)
   )
-  const transport = pino.transport(join(__dirname, 'fixtures', 'to-file-transport.js'), { destination })
+  const transport = pino.transport({
+    src: join(__dirname, 'fixtures', 'to-file-transport.js'),
+    opts: { destination }
```
glensc, line 23: would be cleaner to spell out `options` for public API, same for `source`.

altho `source` would be too abstract, perhaps a more specific name here.

lib/worker.js
```diff
@@ -10,11 +10,11 @@ const { once } = require('events')
 
 /* istanbul ignore file */
 
-module.exports = async function (transports) {
-  transports = await Promise.all(transports.map(async (t) => {
+module.exports = async function ({ destinations }) {
+  destinations = await Promise.all(destinations.map(async (t) => {
```
jsumners, line 14: `Promise.all` could throw and I'm not seeing where it is handled. Should this be `Promise.allSettled`?

jsumners:
> I've refactored the API. Let me know what you think!

Looks good. Just one question about `Promise.all` usage.

lib/worker.js
mcollina: It's handled by https://github.com/pinojs/thread-stream/blob/62d24c681c673aad919bd79fed24e6b75bb92d28/lib/worker.js#L91.

I would still need to throw... in that case what error should I throw if I have more than one?

test/transport.test.js
mcollina: I have been at a loss for good names for these options.. suggestions are welcomed.

@jsumners @davidmarkclements wdyt about using long names?

jsumners: I am fine with using full names.

lib/worker.js
jsumners: So that I'm clear: that `process.on('unhandledRejection')` covers _only_ the worker child process?

mcollina: yes

jsumners: Then I am good with the code as-is.

test/transport.test.js
mcollina: Long names it is!

mcollina:
@davidmarkclements could you take a look? I think we have made some good progress so far.

example-transports.js
```diff
@@ -0,0 +1,61 @@
...
+const { join } = require('path')
+
+const file = join(tmpdir(), `pino-${process.pid}-example`)
+
+const transport = pino.transport([{
+  level: 'warn',
+  destination: file
+}, {
+  level: 'info',
+  module: '/Users/matteo/repositories/pino-elasticsearch/lib.js',
```
davidmarkclements, line 14: should this just be

```js
module: `pino-elasticsearch`
```

example-transports.js
```diff
@@ -0,0 +1,61 @@
+'use strict'
```
davidmarkclements, line 1: now we need an examples folder I think

examples/basic.js
examples/transports.js

lib/transport.js
```diff
@@ -0,0 +1,82 @@
...
+      /* istanbul ignore next */
+      return {
+        ...dest,
+        module: undefined,
+        source: callerRequire.resolve(dest.module)
+      }
+    })
+  }
+
+  // TODO we cannot test this on Windows.. might not work.
```
davidmarkclements, line 73: why can't we test this on windows?

lib/transport.js
```diff
@@ -0,0 +1,82 @@
...
+  let source = fullOptions.source
+
+  if (
+    (source && destinations) ||
+    (source && module) ||
+    (module && destinations)) {
+    throw new Error('Only one of source, destinations or module can be specified')
+  }
+
+  // TODO validate only one of the options can be set
```
davidmarkclements, line 53: stray todo? isn't this done just above?

test/transport.test.js
```diff
@@ -0,0 +1,350 @@
...
+    level: 30,
+    msg: 'hello'
+  })
+})
+
+test('pino.transport with file (no options + error handling)', async ({ equal }) => {
+  const transport = pino.transport({
+    source: join(__dirname, 'fixtures', 'to-file-transport.js')
+  })
+  // TODO: when thread stream passess error handling to main, mop up the console.error
```
davidmarkclements, line 43: stray todo

davidmarkclements:
The API looks really cool, I think there's a low hanging fruit here as well:

If we

* allow options to also be set via environment variables
* allow a stream to be the default export

We can support something like

```js
const myStream = await import('my-stream') // point is  transports we make use of TLA
const { PINO_TRANSPORT_OPTS } = process.env
export default myStream(JSON.parse(PINO_TRANSPORT_OPTS))
```

This can perhaps be a separate PR though.

lib/transport.js
mcollina: Something related to symlink I think. I don't have a windows machine handy to fix it right now.

lib/transport.js
mcollina: Yes, this must be removed.

test/transport.test.js
mcollina: yes

davidmarkclements:
When you have an options API that has both `source` and `destination` it seems like they're connected - you moving something from the source to the destination. Our API is saying load the source CODE execute it and then pass this `destination` option value into the exported function of that code. I think it's confusing to call it `source`. 

Why can't `module` also allow for a file path?

* `target`

mcollina:
The problem with using the same option name for both files and packages was some code that I did not like very much and it looked too brittle. I would love to use better names for these.

mcollina:
> The API looks really cool, I think there's a low hanging fruit here as well:
> 
> 
> 
> If we
> 
> 
> 
> * allow options to also be set via environment variables
> 
> * allow a stream to be the default export
> 
> 
> 
> We can support something like
> 
> 
> 
> ```js
> 
> const myStream = await import('my-stream') // point is  transports can make use of TLA
> 
> const { PINO_TRANSPORT_OPTS } = process.env
> 
> export default myStream(JSON.parse(PINO_TRANSPORT_OPTS))
> 
> ```
> 
> 
> 
> This can perhaps be a separate PR though.

I don't think this is feasible because the transports files are not top level.

lib/worker.js
```diff
@@ -0,0 +1,95 @@
...
+        if (--expected === 0) {
+          cb(err)
+        }
+      }
+    }
+  })
+
+  function process (stream) {
+    const multi = pino.multistream(destinations)
+    // TODO manage backpressure
```
davidmarkclements, line 83: can we close our this PR before this todo is done?

example-transports.js
```diff
@@ -0,0 +1,61 @@
...
+
+const pino = require('./')
+const { tmpdir } = require('os')
+const { join } = require('path')
+
+const file = join(tmpdir(), `pino-${process.pid}-example`)
+
+const transport = pino.transport([{
+  level: 'warn',
```
davidmarkclements, line 10: can we support

```
levels: ['warn', 'error', 'fatal']
```

I think this could be fairly easily achieved - we would just map the levels array to 

```
[{ level: 'warn', ...otherStuff }, { level: 'error', ...otherStuff }, {level: 'fatal, ...otherStuff }]
```

example-transports.js
```diff
@@ -0,0 +1,61 @@
...
+  destination: file
+}, {
+  level: 'info',
+  module: '/Users/matteo/repositories/pino-elasticsearch/lib.js',
+  opts: {
```
davidmarkclements, line 15: should be `options` ?

davidmarkclements:
I'm not sure about the implicit API for pretty printing and file writing via the `destinations` option

I think these should be referred to as built transports and be declared similarly to external or path based transports:

```js
const transport = pino.transport([{
  level: 'warn',
  module: '#pino/writer',
  options: { destination: file }
},  {
  level: 'info',
  module: '#pino/pretty',
  prettyPrint: true
}])
```

I think maybe `module` should be called `target` (and we should get rid of `source`):

```js
const transport = pino.transport([{
  level: 'warn',
  target: '#pino/writer',
  options: { destination: file }
}, 
{
  level: 'trace',
  target: './path/to/local/file.mjs',
  options: {
     consistent: true
  }
},
{
  level: 'info',
  target: 'pino-elasticsearch',
  options: {
    node: 'http://localhost:9200'
  }
},
 {
  level: 'info',
  target: '#pino/pretty',
  prettyPrint: true
}])
```

davidmarkclements:
> I don't think this is feasible because the transports files are not top level.

That doesn't stop TLA - if the files are `await import`ed  it shouldn't make any difference?

lib/worker.js
mcollina: This is something that we should talk about. I'm not 100% convinced we should implement this at all: a slow or stalled destination could block all log processing.

The current behavior is to have memory filled in. At some point the worker would crash, which will terminate the process. How do we want to handle this case?

davidmarkclements:
> The problem with using the same option name for both files and packages was some code that I did not like very much and it looked too brittle. I would love to use better names for these.

Suggesting `target` for all cases:

* if `target` begins with `.` or `/` it's a path (or `file://`)
* if `target` begins with `#` its a built-in
* otherwise `target` is a module

mcollina:
> > I don't think this is feasible because the transports files are not top level.
> 
> 
> 
> That doesn't stop TLA - if the files are `await import`ed  it shouldn't make any difference?

Oh you are right! I've learned something new today.

davidmarkclements:
Should `level` and built-in transport also be available for the single transport case?

Can we get of `destinations` and instead accept either an options object or an array of options objects?

davidmarkclements:
Taken a first pass at docs (written to the suggestions made rather than current state). Feedback welcome :D

mcollina:
I pushed target and targets rename

docs/transports.md
```diff
@@ -1,79 +1,163 @@
...
-Logs can now be consumed using shell piping:
+We would set up our transport by creating a transport stream with `pino.transport`
+and passing it to the `pino` function:
 
-```sh
-node my-app-which-logs-stuff-to-stdout.js | node my-transport-process.js
+```js
+const pino = require('pino')
+const transport = pino.transport({
+  target: './my-transport.mjs'
```
mcollina, line 49: This must be a full path as we do not support relative paths (worker_threads do not support them).

docs/transports.md
```diff
@@ -1,79 +1,163 @@
...
+[Structured Clone Algorithm][sca] which is used to (de)serializing objects between threads.
+
+What if we wanted to use both transports, but send only error logs to `some-file-transport` while
+sending all logs to `./my-transport.mjs`. We can use the `pino.transport` function's `destinations` option:
+
+```js
+const pino = require('pino')
+const transport = pino.transport({
+  destinations: [
+    { target: './my-transport.mjs', level: 'error' },
```
mcollina, line 98: use full path

docs/transports.md
```diff
@@ -1,79 +1,163 @@
...
-const myTransport = through.obj(function (chunk, enc, cb) {
-  // do the necessary
-  console.log(chunk)
-  cb()
-})
 
-pump(process.stdin, split(JSON.parse), myTransport)
+[worker-thread]: https://nodejs.org/dist/latest-v14.x/docs/api/worker_threads.html
+
+## v7+ Transports
```
mcollina, line 22: We need to document `#pino/pretty` and `#pino/file`.

lib/worker.js
davidmarkclements: this is path way to log loss... 

can we have a fallback? log to file to clear memory?

davidmarkclements: this is currently a problem with process transports as well though right - a solution for this would make the switch to v7 even more compelling

docs/transports.md
davidmarkclements: in the near future we should also add #pino/redact (but that needs transport chaining support)

test/transport.test.js
```diff
@@ -0,0 +1,368 @@
...
+    hostname,
+    level: 30,
+    msg: 'hello'
+  })
+})
+
+test('pino.transport with target and targets', async ({ fail, equal }) => {
+  try {
+    pino.transport({
+      target: 'a/file',
```
davidmarkclements, line 280:
```suggestion
      target: '/a/file',
```

test/transport.test.js
```diff
@@ -0,0 +1,368 @@
...
+      targets: [{
+        target: 'a/file'
```
davidmarkclements, line 282:
```suggestion
        target: '/a/file'
```

test/transport.test.js
```diff
@@ -0,0 +1,368 @@
...
+    os.tmpdir(),
+    '_' + Math.random().toString(36).substr(2, 9)
+  )
+  const dest2 = join(
+    os.tmpdir(),
+    '_' + Math.random().toString(36).substr(2, 9)
+  )
+  const transport = pino.transport({
+    targets: [{
+      level: 'info',
```
davidmarkclements, line 198: how does this work with custom levels?

davidmarkclements: can the level option be supported as a top level option as well?

docs/transports.md
mcollina: I would do that in a separate PR.

test/transport.test.js
mcollina: > how does this work with custom levels?

I literally have no clue, I'm not convinced it would work as it is. I would recommend to document that it does not work with custom levels right now.

miladezzat (approved):

mcollina:
@jsumners I think this is ready to go. Could you take a look so we can ship it?

kibertoad:
@mcollina please ping me to revise types after this lands

docs/api.md
```diff
@@ -905,6 +906,98 @@ A `pino.destination` instance can also be used to reopen closed files
...
+For more on transports, how they work, and how to create them see the [`Transports documentation`](/docs/transports.md).
+
+* See [`Transports`](/docs/transports.md)
+* See [`thread-stream` ⇗](https://github.com/mcollina/thread-stream)
+
+#### Options
+
+* `target`:  The transport to pass logs through. This may be an installed module name, an absolute path or a built-in transport (see [Transport Builtins](#transport-builtins))
+* `options`:  An options object which is serialized (see [Structured Clone Algorithm][https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm]), passed to the worker thread, parsed and then passed to the exported transport function.
+* `worker`: [Worker thread](https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options) configuration options. Additional, the `worker` option supports `worker.autoEnd`. If this is set to `false` logs will not be flushed on process exit, it is then up to the developer to call `transport.end()` to flush logs.
```
jsumners, line 953:
```suggestion
* `worker`: [Worker thread](https://nodejs.org/api/worker_threads.html#worker_threads_new_worker_filename_options) configuration options. Additionally, the `worker` option supports `worker.autoEnd`. If this is set to `false` logs will not be flushed on process exit. It is then up to the developer to call `transport.end()` to flush logs.
```

docs/api.md
```diff
@@ -905,6 +906,98 @@ A `pino.destination` instance can also be used to reopen closed files
...
+* `targets`: May be specified instead of `target`, must be an array of transport configurations. Transport configurations include the aforementioned `options` and `target` options plus a `level` option which will send only logs above a specified level to a transport.
```
jsumners, line 954:
```suggestion
* `targets`: May be specified instead of `target`. Must be an array of transport configurations. Transport configurations include the aforementioned `options` and `target` options plus a `level` option which will send only logs above a specified level to a transport.
```

docs/api.md
```diff
@@ -905,6 +906,98 @@ A `pino.destination` instance can also be used to reopen closed files
...
+  options: { destination: '/path/to/file' }
+})
+pino(transport)
+```
+
+The `options.destination` property may also be a number to represent a file descriptor. Typically this would be `1` to write to STDOUT or `2` to write to STDERR. If `options.destination` is not set, it defaults to `1` which means logs will be written to STDOUT.
+
+The difference between using the `#pino/file` transport builtin and using `pino.destination` is that `pino.destination` runs in the main thread, whereas `#pino/file` sets up `pino.destination` in a worker thread.
+
+##### ``#pino/pretty`
```
jsumners, line 983:
```suggestion
##### `#pino/pretty`
```

docs/transports.md
```diff
@@ -1,79 +1,163 @@
...
+const transport = pino.transport({
+  target: 'some-file-transport',
+  options: { destination: '/dev/null' }
+})
+pino(transport)
 ```
-$ node an-app.js | jq 'select(.level == 50)'
+
+Note here we've specified a module by package rather than by relative path. The options object we provide
+is serialized and injected into the transport worker thread, then passed to the modules exported function.
```
jsumners, line 87:
```suggestion
is serialized and injected into the transport worker thread, then passed to the module's exported function.
```

docs/transports.md
```diff
@@ -1,79 +1,163 @@
...
+This means that the options object can only contain types that are supported by the 
+[Structured Clone Algorithm][sca] which is used to (de)serializing objects between threads.
+
+What if we wanted to use both transports, but send only error logs to `some-file-transport` while
+sending all logs to `my-transport.mjs`. We can use the `pino.transport` function's `destinations` option:
```
jsumners, line 92:
```suggestion
sending all logs to `my-transport.mjs`? We can use the `pino.transport` function's `destinations` option:
```

docs/api.md
```diff
@@ -905,6 +906,98 @@ A `pino.destination` instance can also be used to reopen closed files
 * See [Reopening log files](/docs/help.md#reopening)
 * See [Asynchronous Logging ⇗](/docs/asynchronous.md)
 
+<a id="pino-transport"></a>
+### `pino.transport(options) => ThreadStream`
```
jsumners, line 910: There should be a callout here, and in the transports specific docs, to notify the reader about CPU core count. Ideally, a core for the main app + a core per worker thread (transport) should be available.

lib/file-target.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const { Transform, pipeline } = require('stream')
+const pino = require('../pino')
+const { once } = require('events')
+
+module.exports = async function (opts) {
+  // TODO remove the transform
```
jsumners, line 8: TODO?

lib/file-target.js
```diff
@@ -0,0 +1,22 @@
...
+  const stream = new Transform({
+    objectMode: true,
+    autoDestroy: true,
+    transform (chunk, enc, cb) {
+      cb(null, chunk.toString())
+    }
+  })
+
+  // TODO figure out why sync: false does not work here
```
jsumners, line 17: TODO?

lib/pretty-target.js
```diff
@@ -0,0 +1,32 @@
...
+    autoDestroy: true,
+    transform (chunk, enc, cb) {
+      const line = pretty(chunk.toString())
+      cb(null, line)
+    }
+  })
+
+  // TODO figure out why sync: false does not work here
+  // TODO remove the istanbul ignore, add a test using a
+  // child_process
```
jsumners, line 21: TODO?

lib/transport.js
```diff
@@ -0,0 +1,80 @@
...
+    workerData,
+    workerOpts,
+    sync: true // TODO should this be configurable?
+  })
+
+  stream.on('ready', function () {
+    stream.unref()
+
+    if (workerOpts.autoEnd !== false) {
+      // TODO possibly use FinalizationGroup to automatically remove
```
jsumners, line 21: TODO?

mcollina: I prefer to not resolve this here. There is a space for a tiny little utility for doing exactly this, however it's better developed in isolation. It's more a note for improvements in the future.

This means we are leaking memory if a transport is not closed. I think it's ok.

mcollina:
I've addressed most comments @jsumners

jsumners (approved):

mcollina:
@kibertoad could you please send a PR to update the types?

kibertoad:
@mcollina I take it api is final and I can add types?

kibertoad:
on it!

mcollina:
Yes it should be. It might change further in the -rc cycle once we gather feedback, but we are at a point where it's better to give this in the hands of users.

kibertoad:
@mcollina Are you sure that multiple transport example is correct?

When I'm trying this:

```
const transports = pino.transport([
    {
        level: 'info',
        target: '#pino/pretty',
        options: { some: 'options for', the: 'transport' }
    },
    {
        level: 'trace',
        target: '#pino/file',
        options: { destination: './' }
    }
])
```

It explodes on `    if (isAbsolute(origin) || origin.indexOf('file://') === 0) {`, because neither `target`, nor `targets` are set.

mcollina:
good spot: https://github.com/pinojs/pino/pull/1049

PR #1004, opened 2021-04-07
Add pino.multistream

mcollina:
Source code imported from https://github.com/pinojs/pino-multi-stream.

Following up from the introduction of #1003 and #937, I think we can include this in pino as well.

A user might like to send the data to a remote transport and also write the logs to disk.

I plan to make `pino.multistream()` to be usable _within_ a custom transport, so that we do not spawn up more than one thread.


docs/api.md
```diff
@@ -938,6 +939,73 @@ finalLogger.info('exiting...')
...
+  level: 'debug' // this MUST be set at the lowest level of the
+                 // destinations
+}, pino.multistream(streams))
+
+log.debug('this will be written to /tmp/debug.stream.out')
+log.info('this will be written to /tmp/debug.stream.out and /tmp/info.stream.out')
+log.fatal('this will be written to /tmp/debug.stream.out, /tmp/info.stream.out and /tmp/fatal.stream.out')
+```
+
+In order for `multistream` to work, the log level _____must__ be set to the lowest level used in the streams array.
```
jsumners, line 967:
```suggestion
In order for `multistream` to work, the log level __must__ be set to the lowest level used in the streams array.
```

docs/api.md
```diff
@@ -938,6 +939,73 @@ finalLogger.info('exiting...')
...
+
+* `levels`:  Pass custom log level definitions to the instance as an object.
+
++ `dedupe`: Set this to `true` to send logs only to the stream with the higher level. Default: `false`
+
+    `dedupe` flag can be useful for example when using pino-multi-stream to redirect `error` logs to `process.stderr` and others to `process.stdout`:
+
+    ```js
+    var pino = require('pino')
+    var multistream = require('pino-multi-stream').multistream
```
jsumners, line 979:
```suggestion
    var multistream = pino.multistream
```

docs/help.md
```diff
@@ -180,11 +166,10 @@ var streams = [
 
 const logger = pino({
   name: 'my-app',
-  level: 'info',
-}, multistream(streams))
+  level: 'debug', // must be the lower level of all streams
```
jsumners, line 169:
```suggestion
  level: 'debug', // must be the lowest level of all streams
```

lib/multistream.js
```diff
@@ -0,0 +1,138 @@
+'use strict'
+
+const metadata = Symbol.for('pino.metadata')
+
+const defaultLevels = {
```
jsumners, line 5: We should be able to pull these from `lib/levels` now, right?

multistream.js
```diff
@@ -0,0 +1,98 @@
+'use strict'
```
jsumners, line 1: This file is a duplicate of `benchmarks/multistream.js`?

jsumners (commented):
I'm okay with this if you are. I rather liked it being a separate module because we cold say "it exists, but we do not recommend it" and get out of undesirable issues. But I can see that we are not going to be able to get rid of this feature, so bringing into core probably makes sense.

lib/multistream.js
mcollina: done

multistream.js
mcollina: removed, sorry about it

mcollina:
@jsumners @davidmarkclements could you take a final look?

jsumners (approved):
LGTM

PR #1012, opened 2021-04-14
fix: prettyPrint options - misleading error message with wrong option value

### LINKED ISSUE #1007: Misleading error message on wrong options value
Hey 👋 

When passing wrong option value (`ignore` as an array):

```js
pino({
  prettyPrint: {
    ignore: ['hostname']
  }
})
```

The error in console is:

```console
Error: Missing `pino-pretty` module: `pino-pretty` must be installed separately
```

I think it's quite misleading, would be nice to have an options error or smth.

pino - 6.11.2
pino-pretty - 4.7.1

Thanks

lamanabie:
Fixes #1007

### DIFF
--- a/test/pretty.test.js
+++ b/test/pretty.test.js
@@ -66,6 +66,13 @@ test('throws when prettyPrint is true but pino-pretty module is not installed', 
   require.cache[require.resolve('pino-pretty')].exports = prettyFactory
 })
 
+test('throws when prettyPrint has invalid options', async ({ throws, equal }) => {
+  throws(() => pino({ prettyPrint: { ignore: ['hostname'] } }))
+  try { pino({ prettyPrint: { ignore: ['hostname'] } }) } catch ({ message }) {
+    equal(message, 'opts.ignore.split is not a function')
+  }
+})
+
 test('can send pretty print to custom stream', async ({ equal }) => {
   const dest = new Writable({
     objectMode: true,
--- a/test/pretty.test.js
+++ b/test/pretty.test.js
@@ -50,7 +50,7 @@ test('does not throw error when enabled with stream specified', async ({ doesNot
   doesNotThrow(() => pino({ prettyPrint: true }, process.stdout))
 })
 
-test('throws when prettyPrint is true but pino-pretty module is not installed', async ({ throws, equal }) => {
+test('throws when prettyPrint is true but pino-pretty module is not installed', async ({ throws }) => {
   // pino pretty *is* installed, and probably also cached, so rather than
   // messing with the filesystem the simplest way to generate a not found
   // error is to simulate it:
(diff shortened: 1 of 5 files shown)


test/pretty.test.js
```diff
@@ -66,6 +66,13 @@ test('throws when prettyPrint is true but pino-pretty module is not installed',
   require.cache[require.resolve('pino-pretty')].exports = prettyFactory
 })
 
+test('throws when prettyPrint has invalid options', async ({ throws, equal }) => {
+  throws(() => pino({ prettyPrint: { ignore: ['hostname'] } }))
+  try { pino({ prettyPrint: { ignore: ['hostname'] } }) } catch ({ message }) {
+    equal(message, 'opts.ignore.split is not a function')
+  }
```
simoneb, line 73: this test could be rewritten in a slightly better way:

- if no exception is thrown, this test will pass. it should fail instead
- the second assertion implies the first, which is then redundant

simoneb (commented):
Do we have a test to check that the missing pino-pretty error is thrown sometimes?

mcollina (changes requested):
Code is good but the test can be improved as @simoneb said.

test/pretty.test.js
lamanabie: I did try run test with correct prettyPrint options and test correctly failed when no exception is thrown 
```
 ✖ expected to throw

  test/pretty.test.js                                                                      
  68 |                                                                                     
  69 | test('throws when prettyPrint has invalid options', async ({ throws, equal }) => {  
> 70 |   throws(() => pino({ prettyPrint: { ignore: 'hostname' } }))                       
     | --^                                                                                 
  71 |   try { pino({ prettyPrint: { ignore: 'hostname' } }) } catch ({ message }) {       
  72 |     equal(message, 'opts.ignore.split is not a function')                           
  73 |   }                                                                                 

  test: test/pretty.test.js throws when prettyPrint has invalid options
```
Thats why is there that first assertion, that validate if exception is thrown
```javascript
 throws(() => pino({ prettyPrint: { ignore: ['hostname'] } }))
```
Second assertion validate correct error message

I write this test case similar to another test case in test/pretty.test.js, that handle this almost same way
```javascript
test('throws when prettyPrint is true but pino-pretty module is not installed', async ({ throws, equal }) => {
  // pino pretty *is* installed, and probably also cached, so rather than
  // messing with the filesystem the simplest way to generate a not found
  // error is to simulate it:
  const prettyFactory = require('pino-pretty')
  require.cache[require.resolve('pino-pretty')].exports = () => {
    throw Error('Cannot find module \'pino-pretty\'')
  }
  throws(() => pino({ prettyPrint: true }))
  try { pino({ prettyPrint: true }) } catch ({ message }) {
    equal(message, 'Missing `pino-pretty` module: `pino-pretty` must be installed separately')
  }

  require.cache[require.resolve('pino-pretty')].exports = prettyFactory
})
```

simoneb: you are right that the test fails if no exception is thrown the first assertion will fail, but if the error message is wrong the second will not fail. please fix that one.

lamanabie: I did run test with different expected message: 
```javascript
test('throws when prettyPrint has invalid options', async ({ throws, equal }) => {
  throws(() => pino({ prettyPrint: { ignore: ['hostname'] } }))
  try { pino({ prettyPrint: { ignore: ['hostname'] } }) } catch ({ message }) {
    equal(message, 'this will fail')
  }
})
```
Test failed because error message does not equal
```
 FAIL  test/pretty.test.js
 ✖ should be equal

  test/pretty.test.js                                                                   
  70 |   throws(() => pino({ prettyPrint: { ignore: ['hostname'] } }))                  
  71 |   try { pino({ prettyPrint: { ignore: ['hostname'] } }) } catch ({ message }) {  
> 72 |     equal(message, 'this will fail')                                             
     | ----^                                                                            
  73 |   }                                                                              
  74 | })                                                                               
  75 |        
```

I believe test case is written correctly and handle all edge cases.

simoneb, line 73: Yes, it's correct but poorly written. Try a more idiomatic way to write the same:

```suggestion
  throws(() => pino({ prettyPrint: { ignore: ['hostname'] } }), 'opts.ignore.split is not a function')
```
see https://node-tap.org/docs/api/asserts/#tthrowsfn-expectederror-message-extra

lamanabie: I did rewrite all throws in tests that did follow bad pattern.

simoneb (approved):
Thanks for improving the existing tests too!

simoneb:
@mcollina please take a look. some tests have been changed to avoid a weird pattern of asserting an error would be thrown and then try to throw it manually just to assert on the error message

mcollina (approved):
lgtm

jsumners (approved):
LGTM

Issue #1006, opened 2021-04-10
level-change event fired when child is created


laSinteZ:
When I create a child instance of a logger, `level-change` event is fired. This, for example, causes 2 events per request in `pino-http` (can provide simple repro). I am not sure if it's desired behaviour, since it isn't described in the documentation and can be a bit confusing. If it is a desired behaviour, I can help expanding the documentation.

Example code:
```
const pino = require("pino");
const logger = pino();

logger.on("level-change", () => {
  console.log("level-change event had happened");
});

logger.info("Hello, world. No events.")
logger.child({});
logger.child({});
logger.child({});
logger.child({});
```

Output:
```
{"level":30,"time":1618034735938,"pid":16304,"hostname":"MacBook-Air.local","msg":"Hello, world. No events."}
level-change event had happened
level-change event had happened
level-change event had happened
level-change event had happened
```

jsumners:
It happens because we need to set the level on the child instance: https://github.com/pinojs/pino/blob/84fa44b965e5b5bb2d51eb56583ef6b61e0244ef/lib/proto.js#L124

Which in turn invokes: https://github.com/pinojs/pino/blob/84fa44b965e5b5bb2d51eb56583ef6b61e0244ef/lib/levels.js#L84-L111

I _suppose_ an optional second parameter can be passed to signal omitting the event.

mcollina:
I think this is a _bug_, as we should call reinitialize the EventEmitter state when creating a child. However, I'm not sure this is worth the overhead!

laSinteZ:
Well, I guess at least it's worth to document it, since I was quite surprised to see those events polluting logs (even though I've solved it with simple if newVal !== prevVal in my handler)

mcollina:
If you would like to try fixing it, I'd welcome a PR. The problem is caused by child loggers "inheriting" their parent event listeners.

mcollina:
A PR updating the doc is also fine.

laSinteZ:
@mcollina wouldn't it be effectively a breaking change if child loggers WON'T inherit event listeners though? I believe someone could rely on this functionality.

mcollina:
> @mcollina wouldn't it be effectively a breaking change if child loggers WON'T inherit event listeners though? I believe someone could rely on this functionality.

version numbers are there to be bumped. We have the next branch that we accumulate all our breaking changes and then ship them.

PR #1014, opened 2021-04-18
[Docs] document that `.child()` call triggers `level-change` event.

laSinteZ:
As discussed [in the issue 1006](https://github.com/pinojs/pino/issues/1006#issuecomment-817205040) I've created an update to docs. Unfortunately, I have no bandwidth to fix this bug right now, but I guess letting people know about it is also good.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -841,8 +841,7 @@ logger.level = 'trace' // trigger event
 ```
 
 Please note that due to a [known bug](https://github.com/pinojs/pino/issues/1006), every `logger.child()` call will
-fire a `level-change` event, with `levelLabel` equal to `previousLevelLabel` and `levelValue` equal to `previousLevelValue`.
-You might want to ignore these events, using something like:
+fire a `level-change` event. You might want to ignore these events, using something like:
 
 ```js
 const logger = require('pino')()
--- a/docs/api.md
+++ b/docs/api.md
@@ -841,7 +841,7 @@ logger.level = 'trace' // trigger event
 ```
 
 Please note that due to a [known bug](https://github.com/pinojs/pino/issues/1006), every `logger.child()` call will
-fire a `level-change` event. You might want to ignore these events, using something like:
+fire a `level-change` event. These events can be ignored by writing an event handler like:
 
 ```js
 const logger = require('pino')()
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -840,6 +840,20 @@ logger.on('level-change', (lvl, val, prevLvl, prevVal) => {
 logger.level = 'trace' // trigger event
 ```
 
+Please note that due to a [known bug](https://github.com/pinojs/pino/issues/1006), every `logger.child()` call will
+fire a `level-change` event, with `levelLabel` equal to `previousLevelLabel` and `levelValue` equal to `previousLevelValue`.
+You might want to ignore these events, using something like:
+
+```js
+const logger = require('pino')()
+logger.on('level-change', (lvl, val, prevLvl, prevVal) => {
```
mcollina, line 849: I would remove the arrow function and rely on `this === logger` to check.

laSinteZ: Thank you, I did not think of this (no pun intended) :)

mcollina (approved):
lgtm

laSinteZ:
@mcollina pardon me for one more commit, but my statement wasn't correct:
for example, `logger.child({ level: "trace" });` call would result in an event with lvl = trace, prevLvl = info

docs/api.md
```diff
@@ -840,6 +840,20 @@ logger.on('level-change', (lvl, val, prevLvl, prevVal) => {
 logger.level = 'trace' // trigger event
 ```
 
+Please note that due to a [known bug](https://github.com/pinojs/pino/issues/1006), every `logger.child()` call will
+fire a `level-change` event. You might want to ignore these events, using something like:
```
jsumners, line 844:
```suggestion
Please note that due to a [known bug](https://github.com/pinojs/pino/issues/1006), every `logger.child()` call will
fire a `level-change` event. These events can be ignored by writing an event handler like:
```

docs/api.md
```diff
@@ -840,6 +840,20 @@ logger.on('level-change', (lvl, val, prevLvl, prevVal) => {
...
+
+```js
+const logger = require('pino')()
+logger.on('level-change', function (lvl, val, prevLvl, prevVal) {
+  if (logger === this) {
+    console.log('%s (%d) was changed to %s (%d)', prevLvl, prevVal, lvl, val)
+  }
```
jsumners, line 851:
```suggestion
  if (logger !== this) {
    return
  }
  console.log('%s (%d) was changed to %s (%d)', prevLvl, prevVal, lvl, val)
```

jsumners (approved):
LGTM

PR #1016, opened 2021-04-20
Drop flatstr

mcollina:
Since Node v12, this is not needed anymore.

So long, and thanks for all the fish `flatstr`!


jsumners (approved):
LGTM

davidmarkclements:
Are we absolutely certain it's not needed? 

😭

davidmarkclements (approved):
approving on the basis it's def not needed for any minor/patch versions of 12

mcollina:
It's not needed!

PR #1035, opened 2021-05-31
Revert "Add TS types"

davidmarkclements:
Reverts pinojs/pino#913

* npm test fails because types-test isn't correctly referenced
* npm test fails because types-test fails
* let's not add `test-ci` - the reason this passed CI is because it wasn't checking the changes in the PR

Please make a new PR to resolve this, reverting immediately


kibertoad:
@davidmarkclements Will do, sorry

kibertoad:
@davidmarkclements So should I, or should I not add `test-types` inside `test-ci`? Sounded kinda unclear as to why we may not want it there.

PR #1039, opened 2021-06-02
fix(deps): upgrade to sonic-boom 2.0.1

JustinBeckwith:
This has been a fun journey.  This all started with me trying to upgrade `sonic-boom` in one of our own repositories, but found typescript type conflicts between the v1 types in `@types/sonic-boom` and the built-in types in v2.  This was surfaced because we're also using pino, which brings in v1 types.  The PR that started this path is here:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53559

It's unclear exactly what caused the semver major bump between v1 and v2 of sonic-boom, but all of the tests here seem to just pass with no other changes.  A closer look would be appreciated :)  Thank you folks!

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,6 @@
     "flatstr": "^1.0.12",
     "pino-std-serializers": "^3.1.0",
     "quick-format-unescaped": "^4.0.3",
-    "sonic-boom": "^1.0.2"
+    "sonic-boom": "^2.0.1"
   }
 }


jsumners:
How is this a fix? What is broken in the current version of `pino` that upgrading to this next major version of `sonic-boom` will fix?

JustinBeckwith:
@jsumners generally I've approached dependency updates as `fix` with conventional commits, because it makes a material change to the package that would lead to different bits shipped in a new version.  This would unblock problems we're having with TypeScript types and conflicts with sonic-boom v1 and v2.  I am happy to use whichever prefix makes the most sense for your project!

jsumners:
@mcollina and @kibertoad I leave this to you.

kibertoad:
`It's unclear exactly what caused the semver major bump between v1 and v2 of sonic-boom`

It was actually the introduction of incompatible TS types that did it.

@JustinBeckwith Could you also please submit PR to DefinitelyTyped that would obsolete their types and point out that they are embedded now?

JustinBeckwith:
@kibertoad awesome!  That's actually where I started 😆   https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53559

kibertoad:
@JustinBeckwith What will happen to people who currently depend on DefinitelyTyped typings if we merge and release this? Aren't DT Pino types incompatible with `sonic-boom` 2.0.x, and hence would result in broken builds, unless people upgrade to Pino types from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/53559?
I'm not sure if anything needs to be done in Pino master. `next` branch already embeds the whole chain of types and should be fine. Ensuring that no pieces break during piecemeal plugging of types into master is likely to be risky.

kibertoad:
@JustinBeckwith Can't the conflict that you have be addressed by simply removing DT types for sonic-boom from your project? Are they coming transitively from somewhere?

mcollina:
The other reason is to cut/remove old versions. sonic-boom v2 will be part of pino v7.

miladezzat (approved):

miladezzat:
I think, it's good to update deps all the time

jsumners:
It is a breaking change and we are not ready to release a new major version at this time.

mojavelinux:
I'm interested in using new features added in sonic-boom 2 (full disclaimer: some of those features are features I contributed). I'd be willing to evaluate a prerelease of pino. Are there any plans to begin releasing prereleases of the v7 line? (If this is not the right forum to ask this question, kindly direct me to another channel).

mcollina:
You should be able to pass in a sonic-boom v2 stream in place of `pino.destination`.

---

We will ship a prerelase as soon as the new transports have landed.

mojavelinux:
> You should be able to pass in a sonic-boom v2 stream in place of pino.destination.

Aha! Thanks for the tip.

PR #1046, opened 2021-06-23
Better describe base options in docs

tomasstrejcek:
There is possible bug in documentation how the base options works, as with recommended `null` values the log contained
.. "pid":null,"hostname":null ..

When i supplied `undefined` values, this removed those two properties entirely from the log;
also the is no `name` property so maybe the mention of it should be dropped too?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -269,7 +269,7 @@ Default: `{pid: process.pid, hostname: os.hostname}`
 
 Key-value object added as child logger to each log line.
 
-Set to `null` to avoid adding `pid`, `hostname` and `name` properties to each log.
+Set to `undefined` to avoid adding `pid`, `hostname` properties to each log.
 
 #### `enabled` (Boolean)
 


mcollina (approved):
lgtm

mcollina:
good spot, thanks!

kstolp:
I think this might have been a mistake of misinterpreting the documentation.

The original documentation: 
```
#### `base` (Object)
Default: `{pid: process.pid, hostname: os.hostname}`

Key-value object added as child logger to each log line.

Set to `null` to avoid adding `pid`, `hostname` and `name` properties to each log.
```

If you misread this and set `base` to equal `{pid: null, hostname: null}`, then you will get the result @tomasstrejcek was reporting.

However, if you read this and set `base` to equal `null`, then everything functions correctly: `pid`, `hostname`, and `name` properties are omitted from each log line.

There still exists the logic to handle `null` and to omit the `name` property when `base === null`.
https://github.com/pinojs/pino/blob/v6.12.0/pino.js#L142-L149

@mcollina Thoughs?

PR #1050, opened 2021-07-04
Add TS types for transports

kibertoad:
refs #1003

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "lint": "eslint .",
     "test": "npm run lint && tap test/*test.js test/*/*test.js && npm run test-types",
     "test-ci": "npm run lint && tap --no-check-coverage test/*test.js test/*/*test.js --coverage-report=lcovonly && npm run test-types",
-    "test-types": "tsc && tsd",
+    "test-types": "tsc && tsd && ts-node test/types/pino.ts",
     "cov-ui": "tap --coverage-report=html test/*test.js test/*/*test.js",
     "bench": "node benchmarks/utils/runbench all",
     "bench-basic": "node benchmarks/utils/runbench basic",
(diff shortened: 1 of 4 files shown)


pino.d.ts
```diff
@@ -200,6 +201,32 @@ declare namespace P {
...
+    interface TransportSingleOptions<TransportOptions = Record<string, any>> extends TransportBaseOptions<TransportOptions>{
+        target: string
+    }
+
+    interface TransportMultiOptions<TransportOptions = Record<string, any>> extends TransportBaseOptions<TransportOptions>{
+        targets: readonly TransportTargetOptions<TransportOptions>[]
+    }
+
+    // ToDo https://github.com/pinojs/thread-stream/issues/24
+    type ThreadStream = any
```
kibertoad, line 224: Considering that ThreadStream is at a pre-release version and might potentially change a lot, I don't think it makes much sense to in-line its type definitions in pino, especially since it is likely to be simply fed into pino and not manipulated directly.
I'll work on https://github.com/pinojs/thread-stream/issues/24 when there is green light to do so, and replace this type with a reference to bundled type, but I don't think it should be a blocker here.

kibertoad:
@mcollina On a somewhat unrelated note, considering that thread-steam is at a pre-release version, shouldn't it be pinned at patch, not minor version, as breaking changes within minor versions are legal within semver?

mcollina:
thread-stream is an internal library that should not be seen from the outside.

mcollina (approved):
lgtm

package.json
```diff
@@ -92,8 +92,9 @@
     "tap": "^15.0.1",
     "tape": "^5.0.0",
     "through2": "^4.0.0",
-    "tsd": "^0.15.1",
-    "typescript": "^4.2.4",
+    "ts-node": "^10.0.0",
```
davidmarkclements, line 95: why are we adding this?

PR #1053, opened 2021-07-07
Use on-exit-leak-free for transport on-exit

mcollina:
Uses WeakRef, WeakMap and FinalizationGroup to implement a leak-free exit listener.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -939,6 +939,10 @@ const transports = pino.transport({
 pino(transports)
 ```
 
+If `WeakRef`, `WeakMap` and `FinalizationRegistry` are available in the current runtime (v14.5.0+), then the thread
+will be automatically terminated in case the stream or logger goes out of scope.
+The `transport()` function adds a listener to `process.on('exit')` to ensure the worker is flushed and all data synced
+before the process exits.
 
 For more on transports, how they work, and how to create them see the [`Transports documentation`](/docs/transports.md).
 
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -6,6 +6,27 @@ const { join, isAbsolute } = require('path')
 
 const ThreadStream = require('thread-stream')
 
+function setupOnExit (stream) {
+  /* istanbul ignore next */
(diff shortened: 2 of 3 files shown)


lib/transport.js
```diff
@@ -6,6 +6,27 @@ const { join, isAbsolute } = require('path')
 
 const ThreadStream = require('thread-stream')
 
+function setupOnExit (stream) {
+  /* istanbul ignore next */
+  if (global.WeakRef && global.WeakMap && global.FinalizationRegistry) {
```
jsumners, line 11: Looks like v14.17.3 is the first version with full support for all of these. Do we need to make this clear in the documentation?

davidmarkclements: can we just release 7 for min Node version 14.7.3?

davidmarkclements: I say remove the checks and do a version check in index.js, throw and say use Pino 6 for older node versions

davidmarkclements (changes requested):
no need for the checks imo, just drop support for older node in v7

mcollina:
We need to support Node.js v12, it's still LTS and we do not want to force one of our consumers (Fastify) to drop it.

davidmarkclements (approved):
my bad, I forgot 12 has a long life left

docs/api.md
```diff
@@ -939,6 +939,10 @@ const transports = pino.transport({
 pino(transports)
 ```
 
+If `WeakRef`, `WeakMap` and `FinalizationRegistry` are available in the current runtime (v14.5.0+), then the thread
+will be automatically terminated in case the stream or logger goes out of scope.
+The `transport()` function adds a listener to `process.on('exit')` to ensure the worker is flushed and all data synced
+before the process exits.
```
mcollina, line 945: @jsumners docs added

jsumners:
> my bad, I forgot 12 has a long life left

"long" is a bit generous. EOL is April 2022.

jsumners (approved):
Looks good to me.

Issue #1062, opened 2021-07-09
Handling complex circular objects


mcollina:
Pino claims to be "safe" and reliable: I would argue one of our principles is that __logging a line should not crash your process__.

Yet we did that and it was fixed in https://github.com/davidmarkclements/fast-safe-stringify/pull/52 and https://github.com/pinojs/pino/pull/1061. The approach of those fixes is to "give up" and avoid serializing such a complex circular object and produce a string instead.

Here is an example:

```js
const s = new PassThrough()
s.resume()
s.write('', () => {})
const obj = { s, p: new Proxy({}, { get () { throw new Error('kaboom') } }) }
const instance = pino()
instance.info({ obj })
```

This is due to the approach that we take in https://github.com/pinojs/pino/blob/14ab3782cb034f82fd7c0013189162c4a0a8c8f3/lib/tools.js#L400-L406: in case of complex objects (like the one above) that cannot be serialized by `JSON.stringify()` due to circular objects, we call `fast-safe-stringify()`.

I would note that this is _not in the hot path_ as the performance of serializing such a complex object is really bad anyway. Moreover, logging a line should not crash a process. The reason for the problem was that `fast-safe-stringify` replaces the keys with circular references.  It's a pretty risky approach to take in something that is not a hot path and needed for the developer experience.

One approach to avoid this kind of problems in the future is to use https://github.com/BridgeAR/safe-stable-stringify from @BridgeAR instead of fss. We could potentially look into some other safe-but-slow algorithms.

mcollina:
WDYT @jsumners @davidmarkclements ?

jsumners:
Ship it.

davidmarkclements:
this makes sense to me

davidmarkclements:
closed by #1066

PR #1065, opened 2021-07-10
Replace fast-safe-stringify with json-stringify-safe

mcollina:
Fixes https://github.com/pinojs/pino/issues/1062

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -5,7 +5,7 @@
 const format = require('quick-format-unescaped')
 const { mapHttpRequest, mapHttpResponse } = require('pino-std-serializers')
 const SonicBoom = require('sonic-boom')
-const stringifySafe = require('safe-stable-stringify')
+const stringifySafe = require('json-stringify-safe')
 const {
   lsCacheSym,
   chindingsSym,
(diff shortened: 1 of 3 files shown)


jsumners (approved):

mcollina:
Should we ship this in v7 or v6?

climba03003:
Will it be the problem https://github.com/BridgeAR/safe-stable-stringify/issues/17 ?
Currently, `level` will place first for logging and it is easier to trace. With this update, it may becomes `hostname` or something unpredictable because of custom `bindings`.

jsumners:
> Should we ship this in v7 or v6?

7

mcollina:
> Will it be the problem [BridgeAR/safe-stable-stringify#17](https://github.com/BridgeAR/safe-stable-stringify/issues/17) ?
> Currently, `level` will place first for logging and it is easier to trace. With this update, it may becomes `hostname` or something unpredictable because of custom `bindings` when `JSON.stringify` failed.

I replaced it with json-stringify-safe instead. It's an old module which is incredibly stable and trusted. The algorithm is quite simple, we could even think of vendoring some iteration of that code.

jsumners (approved):

mcollina:
I moved this to `next` in https://github.com/pinojs/pino/pull/1066. It was simpler to redo as there were a bit of conflicts to rebase.

PR #1066, opened 2021-07-11
Replace fast-safe-stringify with json-stringify-safe

mcollina:
Fixes https://github.com/pinojs/pino/issues/1062

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -5,7 +5,7 @@
 const format = require('quick-format-unescaped')
 const { mapHttpRequest, mapHttpResponse } = require('pino-std-serializers')
 const SonicBoom = require('sonic-boom')
-const stringifySafe = require('fast-safe-stringify')
+const stringifySafe = require('json-stringify-safe')
 const {
   lsCacheSym,
   chindingsSym,
@@ -437,7 +437,11 @@ function stringify (obj) {
   try {
     return JSON.stringify(obj)
   } catch (_) {
-    return stringifySafe(obj)
+    try {
+      return stringifySafe(obj)
+    } catch (_) {
+      return '"[unable to serialize, circular reference is too complex to analyze]"'
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


lib/tools.js
```diff
@@ -437,7 +437,11 @@ function stringify (obj) {
   try {
     return JSON.stringify(obj)
   } catch (_) {
-    return stringifySafe(obj)
+    try {
+      return stringifySafe(obj)
+    } catch (_) {
+      return '"[unable to serialize, circular reference is too complex to analyze]"'
```
kibertoad, line 443: why this double wrap of error message?

test/basic.test.js
```diff
@@ -593,19 +593,18 @@ test('children with same names render in correct order', async ({ equal }) => {
   equal(a, 3, 'last logged object takes precedence')
 })
 
-// https://github.com/pinojs/pino/pull/251 - use this.stringify
-test('use `fast-safe-stringify` to avoid circular dependencies', async ({ same }) => {
+test('use `json-stringify-safe` to avoid circular dependencies', async ({ same }) => {
```
kibertoad, line 596: is the implementation detail relevant to the test? shouldn't it be about the observable behaviour?

lib/tools.js
mcollina: Because it can still throw. If even the "safe" options throws, there is something really odd in that object and we should give up, not blow up the error to the user.

test/basic.test.js
mcollina: Unfortunately it is. See the line changed in the output.

lib/tools.js
kibertoad: Sorry for not being clear. I meant th `'"[["'` wrap of message itself.

kibertoad (approved):

lib/tools.js
mcollina: It follows the same pattern of `[Circular]` used for circular references

jsumners (approved):

PR #1059, opened 2021-07-08
feat: child redact option

climba03003:
Fixes: #831

I think child redact option is a important piece in my application, so I want to push forward for the child `options` discussed in the issue mention above.
I am relatively new in `pino` and I hope the method adopted do not have anything wrong.

I have run benchmark in my computer against the master bench and it should claim that no performance drop.

<details><summary>Before Benchmark</summary>
<p>

```
Running BASIC benchmark

benchBunyan*10000: 565.999ms
benchWinston*10000: 415.691ms
benchBole*10000: 228.661ms
benchDebug*10000: 365.646ms
benchLogLevel*10000: 335.88ms
benchPino*10000: 327.115ms
benchPinoAsync*10000: 164.915ms
benchPinoNodeStream*10000: 247.776ms
benchBunyan*10000: 457.424ms
benchWinston*10000: 362.67ms
benchBole*10000: 213.013ms
benchDebug*10000: 393.197ms
benchLogLevel*10000: 381.517ms
benchPino*10000: 350.057ms
benchPinoAsync*10000: 155.095ms
benchPinoNodeStream*10000: 275.198ms

Running OBJECT benchmark

benchBunyanObj*10000: 631.586ms
benchWinstonObj*10000: 591.859ms
benchBoleObj*10000: 317.796ms
benchLogLevelObject*10000: 593.543ms
benchPinoObj*10000: 335.924ms
benchPinoAsyncObj*10000: 142.843ms
benchPinoNodeStreamObj*10000: 245.006ms
benchBunyanObj*10000: 491.808ms
benchWinstonObj*10000: 386.373ms
benchBoleObj*10000: 230.106ms
benchLogLevelObject*10000: 506.045ms
benchPinoObj*10000: 352.597ms
benchPinoAsyncObj*10000: 133.298ms
benchPinoNodeStreamObj*10000: 248.173ms

Running DEEP-OBJECT benchmark

benchBunyanDeepObj*10000: 1.637s
benchWinstonDeepObj*10000: 2.896s
benchBoleDeepObj*10000: 2.190s
benchLogLevelDeepObj*10000: 6.290s
benchPinoDeepObj*10000: 2.989s
benchPinoAsyncDeepObj*10000: 2.666s
benchPinoNodeStreamDeepObj*10000: 2.528s
benchBunyanDeepObj*10000: 1.664s
benchWinstonDeepObj*10000: 2.776s
benchBoleDeepObj*10000: 2.044s
benchLogLevelDeepObj*10000: 6.460s
[...]

PR #1067, opened 2021-07-12
refactor: deprecate child bindings apply options

climba03003:
Deprecate apply options through bindings for child logger.

### DIFF
--- a/benchmarks/child-creation.bench.js
+++ b/benchmarks/child-creation.bench.js
@@ -60,6 +60,13 @@ const run = bench([
       child.info({ hello: 'world' })
     }
     setImmediate(cb)
+  },
+  function benchPinoCreationWithOption (cb) {
+    const child = plogDest.child({ a: 'property' }, { redact: [] })
+    for (var i = 0; i < max; i++) {
+      child.info({ hello: 'world' })
+    }
+    setImmediate(cb)
   }
 ], 10000)
 
--- a/browser.js
+++ b/browser.js
@@ -113,13 +113,17 @@ function pino (opts) {
     set(setOpts, logger, 'trace', 'log')
   }
 
-  function child (bindings) {
+  function child (bindings, childOptions) {
     if (!bindings) {
       throw new Error('missing bindings for child Pino')
     }
-    const bindingsSerializers = bindings.serializers
-    if (serialize && bindingsSerializers) {
-      var childSerializers = Object.assign({}, serializers, bindingsSerializers)
+    childOptions = childOptions || {}
+    if (serialize && bindings.serializers) {
+      childOptions.serializers = bindings.serializers
+    }
+    const childOptionsSerializers = childOptions.serializers
+    if (serialize && childOptionsSerializers) {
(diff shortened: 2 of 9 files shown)


benchmarks/child-creation.bench.js
```diff
@@ -60,6 +60,13 @@ const run = bench([
       child.info({ hello: 'world' })
     }
     setImmediate(cb)
+  },
+  function benchPinoCreationWithOption (cb) {
+    const child = plogDest.child({ a: 'property' }, { redact: [] })
+    for (var i = 0; i < max; i++) {
```
kibertoad, line 66: why not let?

browser.js
```diff
@@ -113,13 +113,17 @@ function pino (opts) {
...
-    const bindingsSerializers = bindings.serializers
-    if (serialize && bindingsSerializers) {
-      var childSerializers = Object.assign({}, serializers, bindingsSerializers)
+    childOptions = childOptions || {}
+    if (serialize && bindings.serializers) {
+      childOptions.serializers = bindings.serializers
+    }
+    const childOptionsSerializers = childOptions.serializers
+    if (serialize && childOptionsSerializers) {
+      var childSerializers = Object.assign({}, serializers, childOptionsSerializers)
```
kibertoad, line 126: why not let?

kibertoad:
@climba03003 Was this intended for master and not next?

climba03003:
> 
> 
> @climba03003 Was this intended for master and not next?

It should comes with #1059 and nothing is breaking. So, this PR can be landed on both `master` and `next` without problem.

browser.js
climba03003: Do you means `var childSerializers` here? I do not know why it is `var` in the beginning, so I will keep it.

benchmarks/child-creation.bench.js
climba03003: the looping is copy from other test. As I remember, `for-var` loop out perform `for-let` loop. I think that is the reason.

mcollina (approved):
lgtm

kibertoad:
@climba03003 next likely needs a slight TS types updates

kibertoad (approved):

browser.js
kibertoad: yes. yeah, it's a nitpick, just no good reason to use var anymore, so eventually we should convert.

benchmarks/child-creation.bench.js
kibertoad: huh, interesting. TIL

climba03003:
One of the fields is missing `bindings.level`. I am thinking of, how to test the deprecate warning.

kibertoad:
@climba03003 How about putting a spy on console? might need an additional lib, though

climba03003:
> 
> 
> doesn't this also need an assert to check the overall amount of asserts in test to ensure this is called?

`deprecations.test.js` is a tricky test files, it all rely on `pass` to prevent error. If we use the fixed test-case amount. It will becomes all failed. As the `onWarning` functions will receive all the other warning at the same time.

jsumners (approved):

PR #1073, opened 2021-07-21
refactor: update typings

climba03003:
Update typings in favor of #1059 and #1067

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -683,6 +683,13 @@ declare namespace P {
     interface ChildLoggerOptions {
         level?: Level | string;
         serializers?: { [key: string]: SerializerFn };
+        customLevels?: { [key: string]: number };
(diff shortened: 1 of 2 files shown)


pino.d.ts
```diff
@@ -677,9 +677,19 @@ declare namespace P {
...
 
     interface Bindings {
+        [key: string]: any;
+    }
+
+    interface ChildLoggerOptions {
         level?: Level | string;
         serializers?: { [key: string]: SerializerFn };
-        [key: string]: any;
+        customLevels?: { [key: string]: number };
```
climba03003, line 686: I do not know if it is intended not to include both `customLevels` and `formatters`, as it do not support in `broswer`.

mcollina (approved):
lgtm

kibertoad (approved):

PR #1089, opened 2021-08-18
Make ThreadStream sync: false (the default)

mcollina:
There is no reason to be `sync: true` on this and it slows everything dow.

### DIFF
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -29,12 +29,11 @@ function setupOnExit (stream) {
   }
 }
 
-function buildStream (filename, workerData, workerOpts) {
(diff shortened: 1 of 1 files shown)


lib/transport.js
```diff
@@ -29,12 +29,11 @@ function setupOnExit (stream) {
   }
 }
 
-function buildStream (filename, workerData, workerOpts) {
+function buildStream (filename, workerData, workerOpts, streamOpts = {}) {
```
jsumners, line 32: Looks like `streamOpts` is added but not used?

jsumners (approved):
Looks good to me.

PR #1091, opened 2021-08-18
Automatically flush the logs on exit in Node v14+

mcollina:
This PR removes the main developer experience hurdle in using asynchronous logging: the use of `pino.final()`.
Thanks to `on-exit-leak-free` we can automatically flush on exit without leaking event emitter listeners (in most cases).
This uses the same mechanism I experimented with for `pino.transport()`.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -368,6 +368,12 @@ function setupOnExit (stream) {
 }
 
 function autoEnd (stream, eventName) {
+  // This check is needed only on some platforms
+  /* istanbul ignore next */
+  if (stream.destroyed) {
+    return
+  }
+
   if (eventName === 'beforeExit') {
     // We still have an event loop, let's use it
     stream.flush()
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -330,6 +330,10 @@ function hasBeenTampered (stream) {
 function buildSafeSonicBoom (opts) {
   const stream = new SonicBoom(opts)
   stream.on('error', filterBrokenPipe)
+  // if we are sync: false, we must flush on exit
+  if (!opts.sync) {
+    setupOnExit(stream)
+  }
   return stream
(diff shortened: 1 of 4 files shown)


test/exit.test.js
```diff
@@ -37,7 +37,9 @@ test('pino with no args log everything when calling process.exit(0)', async ({ n
   not(actual.match(/world/), null)
 })
 
-test('sync false does not log everything when calling process.exit(0)', async ({ equal }) => {
+const hasWeak = !!global.WeakRef
```
jsumners, line 40: nit: I despise double inverses more than coercing checks.

jsumners (approved):
Looks good to me.

test/exit.test.js
davidmarkclements: eh 🤷 all forms of boolean conversion are a bit shit. `Boolean(WeakRef)` is where you call a class without `new` to cast it. `==` is not good for obvious reasons, `!~~` is too esoteric. I suppose you could use `'WeakRef' in global` instead

davidmarkclements (approved):
outstanding, I love this. CI has an issue, but preapproving

lib/tools.js
```diff
@@ -73,7 +74,7 @@ function asString (str) {
   if (l > 100) {
     return JSON.stringify(str)
   }
-  for (var i = 0; i < l && point >= 32; i++) {
```
jsumners, line 76: Has something changed with `var` vs `let` in loops?

mcollina: I just ran standard --fix on this file. I'll revert, good spot.

jsumners: Use `eslint --fix`. We are using that here so we don't have to liter the whole codebase with pragma directives specifically because of `var/let`.

PR #1092, opened 2021-08-20
Use fastify-warning for deprecation notices

jsumners:
This pull request uses `fastify-warning` instead of `process.emitWarning` directly in order to reduce the verbosity of warning messages. With this change, warnings will be emitted once per warning instead of every time the deprecated usage is invoked.

---
PR donated @knockaway.


kibertoad:
Will this be ported to main as well?

mcollina (approved):
lgtm

PR #1093, opened 2021-08-20
Use fastify-warning in v7

jsumners:
Ports #1092 to the v7 branch.


mcollina (approved):
lgtm

mcollina:
I think we should remove all those things in v7. Wdyt?

jsumners:
I agree. But we should keep the definitions around.

PR #1097, opened 2021-08-24
Document pino.transport() async startup

### LINKED ISSUE #1095: next: worker-thread messages not log when pino used in both main-thread and worker-thread
When I use pino in both main-thread and worker-thread. Only the main-thread can log messages.

Repro Link: https://replit.com/@climba03003/pino-issue-1095

```js
// index.js
const { logger } = require('./logger.js')
const { Worker } = require('worker_threads')

const worker = new Worker('./pino.js')
setTimeout(function() {
  worker.terminate()
  process.exit(0)
}, 3000)

logger.info('main - show log')
logger.info('main - show log')
logger.info('main - show log')
logger.info('main - show log')
logger.info('main - show log')
logger.info('main - show log')
logger.info('main - show log')
logger.info('main - show log')

// logger.js
const pino = require('pino')

const logger = pino({ level: 'trace' }, pino.transport({
  targets: [
    { level: 'info', target: '#pino/file', options: {} }
  ]
}))

module.exports.logger = logger

// pino.js
const { logger } = require('./logger.js')

logger.info('worker - should show')
logger.info('worker - should show')
logger.info('worker - should show')
logger.info('worker - should show')
logger.info('worker - should show')
logger.info('worker - should show')
logger.info('worker - should show')

process.exit(0)
```

mcollina:
Fixes #1095 

Unfortunately the asynchronous startup is a limitation of the current implementation. It might be possible to overcome this in the future but we should document it for now.

This is due to ThreadStream waiting for a `'READY'` message on the MessagePort - we would buffer all data in the meanwhile: https://github.com/pinojs/thread-stream/blob/e88be6efaccb9722b4b3ff9be6aa2cc8c1fedabd/index.js#L222-L225.

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -104,11 +104,33 @@ pino(transport)
 
 For more details on `pino.transport` see the [API docs for `pino.transport`][pino-transport].
 
-
 [pino-transport]: /docs/api.md#pino-transport
 [sca]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
 
 
+### Asynchronous startup
+
+The new transports boot asynchronously and calling `process.exit()` before the transport
+started will cause logs to not be delivered.
+
+```js
+const pino = require('pino')
+const transport = pino.transport({
+  targets: [
+    { target: '/absolute/path/to/my-transport.mjs', level: 'error' },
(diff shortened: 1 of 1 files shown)


jsumners (approved):
Looks good to me.

PR #1099, opened 2021-08-26
Improve TS types

kibertoad:
While working on https://github.com/fastify/fastify/issues/3101 I've noticed following things:

1) `Logger` is not correct, as all members of an object must conform to the string index signature (see https://stackoverflow.com/questions/57371839/merge-typescript-record-or-dictionary-like-type-with-fixed-key-typings)
2) BaseLogger is not so base, better minimal logger typing would be this: https://github.com/fastify/fastify/blob/6ded7ffd767ea0348159435a9d7119a0de859dff/types/logger.d.ts#L45
3) Dropped some obsolete typings for code that does not seem to be present anymore

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   },
   "homepage": "http://getpino.io",
   "devDependencies": {
-    "@types/node": "^16.4.10",
+    "@types/node": "^16.7.2",
     "airtap": "4.0.3",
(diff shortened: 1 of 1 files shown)


kibertoad:
@mcollina Are there any pending changes before we can publish a new RC?

Issue #1102, opened 2021-08-27
pino tried to access pino-pretty, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound


Unbuttun:
While using pino and enabling the pretty print module, while installing or using the application through yarn 2/3 it starts complaining and refusing to start with error.

Example - 
```ts
  const app = await fastify({
    logger: {
      prettyPrint: true
    }
  });
```
  
Now when we start the app, yarn will give this error - 

```shell
pino tried to access pino-pretty, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

Currently, we need to add this into `.yarnrc.yml`
```yml  
packageExtensions:
  pino@*:
    dependencies:
      pino-pretty: "*"
```
  
Stack trace 
```shell
Require stack:
- /1/app/server/.yarn/cache/pino-npm-6.13.1-69c64f98df-8124217b9f.zip/node_modules/pino/lib/tools.js
- /1/app/server/.yarn/cache/pino-npm-6.13.1-69c64f98df-8124217b9f.zip/node_modules/pino/lib/levels.js
- /1/app/server/.yarn/cache/pino-npm-6.13.1-69c64f98df-8124217b9f.zip/node_modules/pino/lib/proto.js
- /1/app/server/.yarn/cache/pino-npm-6.13.1-69c64f98df-8124217b9f.zip/node_modules/pino/pino.js
- /1/app/server/.yarn/cache/fastify-npm-3.20.2-2983538579-ae02705d72.zip/node_modules/fastify/lib/logger.js
- /1/app/server/.yarn/cache/fastify-npm-3.20.2-2983538579-ae02705d72.zip/node_modules/fastify/lib/reply.js
- /1/app/server/.yarn/cache/fastify-npm-3.20.2-2983538579-ae02705d72.zip/node_modules/fastify/fastify.js
- /1/app/server/apps/server/dist/main.js
    at /1/app/server/apps/server/dist/main.js:53:9
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

mcollina:
The default package manager for Node.js is npm. It is your choice to use a package manager whose maintainers have decided to change the behavior of module resolution. Adding pino-pretty to dependencies or peerDependencies will cause a few additional modules to be installed for production usage.. and I am pretty sure that a lot of devs care about package size too.

[...]

PR #1103, opened 2021-08-27
fix: Moves pino-pretty from devDependencies to dependencies

Unbuttun:
Fixes - https://github.com/pinojs/pino/issues/1102

More explanation on the issue.


jsumners:
Thank you for taking the time to submit a contribution, but we do not include `pino-pretty` as a hard dependency for a reason. We do not promote prettification in production and our docs are very clear that you must install the module on your own if you want in-process prettification.

Unbuttun:
Can we add it to peerDependenciesMeta then?

```json
"peerDependenciesMeta": {
  "pino-pretty": {
    "optional": true
  }
}
```

jsumners:
> Can we add it to peerDependenciesMeta then?
> 
> ```json
> "peerDependenciesMeta": {
>   "pino-pretty": {
>     "optional": true
>   }
> }
> ```
> 
> @jsumners

Please follow #1106.

Issue #1105, opened 2021-08-28
Add a transports option to the constructor


mcollina:
The option would need to pass its inputs as pino.transport({ transports }) and use the result as the destination stream. This would be useful for Fastify and others that do not re-expose pino.transport() to their users.

cc @davidmarkclements

Eomm:
I'm working on it!

davidmarkclements:
@mcollina sgtm

Eomm:
This could be closed 👍🏽

PR #1111, opened 2021-08-31
add transports constructor option

Eomm:
Implements #1105 

Opening as a draft to ask what do you prefer if:

- users add unnecessary options like a destination: should pino throw an error?

```js
  const instance = pino({
    name: 'logger',
    transport: {
      target: '#pino/file',
      options: { destination }
    }
  }, '/log/path')
``` 

- are you ok with the new `transports` field name to manage both `pino.transport()` input? ( `{target}` or `{targets:[]}` )

Then I'm going to add more tests and doc as well

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -367,6 +367,23 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
 
 * See [Browser API ⇗](/docs/browser.md)
 
+#### `transport` (Object)
+
+The `transport` option is a shortcut to the [pino.transport()](#pino-transport) function.
+It support the same input options.
+```js
+require('pino')({
+  transport: {
+    target: '/absolute/path/to/my-transport.mjs'
+  }
+})
+```
+
+It is not possible to provide an additional [`destination`](#destination) argument
+when using the `transport` option. In this case an `Error` will be thrown.
+
+* See [pino.transport()](#pino-transport)
+
 <a id="destination"></a>
 ### `destination` (SonicBoom | WritableStream | String | Object)
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -369,7 +369,7 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
 
 #### `transport` (Object)
 
-The `transport` option is a shortcut to the [pino.transport()](#pino-transport) function.
+The `transport` option is a shorthand for the [pino.transport()](#pino-transport) function.
 It supports the same input options:
 ```js
 require('pino')({
@@ -389,7 +389,11 @@ require('pino')({
 })
 ```
 
-It is not possible to provide an additional [`destination`](#destination) argument
+If the transport option is supplied to `pino`, a [`destination`](#destination) parameter may not also be passed as a separate argument to `pino`:
+
+```js
+pino({ transport: {}}, '/path/to/somewhere') // THIS WILL NOT WORK, DO NOT DO THIS
+pino({ transport: {}}, process.stderr) // THIS WILL NOT WORK, DO NOT DO THIS
 when using the `transport` option. In this case an `Error` will be thrown.
 
 * See [pino.transport()](#pino-transport)
--- a/docs/api.md
+++ b/docs/api.md
@@ -370,11 +370,21 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
 #### `transport` (Object)
 
 The `transport` option is a shortcut to the [pino.transport()](#pino-transport) function.
-It support the same input options.
+It support the same input options:
 ```js
 require('pino')({
   transport: {
     target: '/absolute/path/to/my-transport.mjs'
+  }
+})
+
+// or multiple transports
+require('pino')({
+  transport: {
+    targets: [
+      { target: '/absolute/path/to/my-transport.mjs', level: 'error' },
+      { target: 'some-file-transport', options: { destination: '/dev/null' }
+    ]
   }
 })
 ```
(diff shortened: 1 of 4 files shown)


lib/tools.js
```diff
@@ -398,6 +399,9 @@ function createArgsNormalizer (defaultOptions) {
     } else if (opts instanceof SonicBoom || opts.writable || opts._writableState) {
       stream = opts
       opts = null
+    } else if (opts.transports) {
```
mcollina, line 402: I think we should call it transport instead of transports.

mcollina (commented):
Good work! I think it should throw if stream is otherwise defined.

test/transport.test.js
```diff
@@ -380,3 +380,95 @@ test('stdout in worker', async ({ not }) => {
...
+  delete result.time
+  same(result, {
+    pid,
+    hostname,
+    level: 30,
+    msg: 'transport option test'
+  })
+})
+
+test('pino transports options with targets', async ({ teardown, same }) => {
```
mcollina, line 409:
```suggestion
test('pino transport options with targets', async ({ teardown, same }) => {
```

test/transport.test.js
```diff
@@ -380,3 +380,95 @@ test('stdout in worker', async ({ not }) => {
   await once(child, 'close')
   not(strip(actual).match(/Hello/), null)
 })
+
+test('pino transports options with target', async ({ teardown, same }) => {
```
mcollina, line 384:
```suggestion
test('pino transport options with target', async ({ teardown, same }) => {
```

docs/api.md
```diff
@@ -367,6 +367,23 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
...
 * See [Browser API ⇗](/docs/browser.md)
 
+#### `transport` (Object)
+
+The `transport` option is a shortcut to the [pino.transport()](#pino-transport) function.
+It support the same input options.
+```js
+require('pino')({
+  transport: {
+    target: '/absolute/path/to/my-transport.mjs'
```
mcollina, line 377: We also support modules, not just absolute paths.

Eomm: Yes, the transport supports many options and I thought it would be useful a quick example.
Then users may deep down in the `transport` section.

Does it worth adding more examples or ask the user to read the `transport` page?

mcollina: Add a few more examples here.

mcollina (approved):
lgtm

lib/tools.js
```diff
@@ -394,10 +395,16 @@ function createArgsNormalizer (defaultOptions) {
       stream = buildSafeSonicBoom({ dest: opts, sync: true })
       opts = {}
     } else if (typeof stream === 'string') {
+      if (opts && opts.transport) {
+        throw Error('Only one of option.transport or stream can be specified')
```
jsumners, line 399: Are we throwing errors with uppercase strings anywhere else? If not, we should be consistent by:

```suggestion
        throw Error('only one of option.transport or stream can be specified')
```

docs/api.md
```diff
@@ -367,6 +367,33 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
 
 * See [Browser API ⇗](/docs/browser.md)
 
+#### `transport` (Object)
+
+The `transport` option is a shortcut to the [pino.transport()](#pino-transport) function.
```
davidmarkclements, line 372:
```suggestion
The `transport` option is a shorthand for the [pino.transport()](#pino-transport) function.
```

docs/api.md
```diff
@@ -367,6 +367,33 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
...
+It supports the same input options:
+```js
+require('pino')({
```
davidmarkclements, line 375: do we do this require('pino') inline thing elsewhere in the docs?

docs/api.md
```diff
@@ -367,6 +367,33 @@ documented in the [Browser API ⇗](/docs/browser.md) documentation.
...
+  transport: {
+    targets: [
+      { target: '/absolute/path/to/my-transport.mjs', level: 'error' },
+      { target: 'some-file-transport', options: { destination: '/dev/null' }
+    ]
+  }
+})
+```
+
+It is not possible to provide an additional [`destination`](#destination) argument
```
davidmarkclements, line 392: this is confusing, and also in the error message we call it stream, instead of destination argument. 

```suggestion
If the transport option is supplied to `pino`, a [`destination`](#destination) parameter may not also be passed as a separate argument to `pino`:

```js
pino({ transport: {}}, '/path/to/somewhere') // THIS WILL NOT WORK, DO NOT DO THIS
pino({ transport: {}}, process.stderr) // THIS WILL NOT WORK, DO NOT DO THIS
```

docs/api.md
Eomm: There are 18 occurrences of `require('pino')` in this file actually: would you like to migrate the examples to `pino`?

test/transport.test.js
```diff
@@ -380,3 +380,95 @@ test('stdout in worker', async ({ not }) => {
...
+test('transport options with target and stream', async ({ fail, equal }) => {
+  try {
+    pino({
+      transport: {
+        target: {}
+      }
+    }, '/log/null')
+    fail('must throw')
+  } catch (err) {
+    equal(err.message, 'Only one of option.transport or stream can be specified')
```
Eomm, line 472:
```suggestion
    equal(err.message, 'only one of option.transport or stream can be specified')
```

docs/api.md
jsumners: Do they assign the result of `require` or do they use it directly like this example?

Eomm: Both declaration tbh

https://github.com/pinojs/pino/blob/5333d73773f59331c2d5c7962dec3ee00574fc6f/docs/api.md#destination-sonicboom--writablestream--string--object

I can update this example and the other ones if you tell me what do you prefer.
Is the cjs syntax not up to date for users?

jsumners (approved):

Issue #1107, opened 2021-08-28
Test pino.transport() with pnpm and yarn 2/3 pnp


mcollina:
As titled, we need to verify how we load 3-party transports in yarn 2/3 pnp and pnpm. Use pino-elasticsearch for testing if if all works fine.

Eomm:
Working on it!

Eomm:
> Use pino-elasticsearch for testing if all works fine.

Is it not updated to pino@7, or doesn't it need any change?

mcollina:
it has been already updated.

PR #1113, opened 2021-09-01
Package manager CI action

Eomm:
Fix https://github.com/pinojs/pino/issues/1107
A run of this action here: https://github.com/Eomm/pino/actions/runs/1190604660

This PR requires a todo list to recap its status:

### pnpm
 
- [x] <del>no issues ✅</del> flaky test on win and node 16: disabled coverage to let it works

### yarn

- [x] `yarn@2` at this time, it triggers an error when the `typescript` module is higher than `4.3.2`. Asking to reopen this issue https://github.com/yarnpkg/berry/issues/2935 . A _temporary solution_ is to stick the TS module version.
- [x] `yarn@3 / PnP` must skip the test that create at runtime a `node_module/transport` package. This happens because the module does not exist during the installation phase and it is not packaged. Not sure it is a valid check. Asked help [here](https://github.com/yarnpkg/berry/discussions/3369)
- [x] `yarn@3 / PnP` does not works withing `pino-elasticsearch`. This **must be supported to be PnP compatible**.
- [x] `yarn@3 / PnP: disabled typescript test due https://github.com/yarnpkg/berry/issues/3394

#### deep down on `yarn@3 / PnP`

The `pino-elasticsearch` issue is caused by the yarn's `require.resolve()` resolution that produces a string like this `[...]cache/pino-elasticsearch-npm-6.1.0-0c03079478-6915435172.zip/node [...]` (**note** the zip file). 

This string is provided as `fileName` parameter to the `thread-stream` module:

https://github.com/pinojs/thread-stream/blob/main/lib/worker.js#L18

that throws the error:

```
+ENOTDIR: not a directory, stat '/Users/mspigolon/workspace/pino/.yarn/cache/pino-elasticsearch-npm-6.1.0-0c03079478-6915435172.zip/node_modules/pino-elasticsearch/lib.js'
```

Yarn supports this zip path [out of the box](https://yarnpkg.com/advanced/pnpapi#accessing-the-files) (monkey patching it) but it does not support the `import` ESM loader.
Reference: https://github.com/yarnpkg/berry/issues/638 and there is a draft PR to support it https://github.com/yarnpkg/berry/pull/2161

So, to solve the PnP support issue, I see two main paths:

1. update the `thread-stream` module to support `yarn` zip path using the [`@yarnpkg/fslib`](https://www.npmjs.com/package/@yarnpkg/fslib) module.
1. wait for the https://github.com/yarnpkg/berry/pull/2161 but **I'm not sure it will work** because the import is executed by the worker thread and I doubt it is monkey patched


merceyz:
> So, to solve the PnP support issue, I see two main paths:

A third option would be to fall back to `require` if `import` fails

> but I'm not sure it will work because the import is executed by the worker thread and I doubt it is monkey patched

I haven't tested but it should work, the `Worker` is supposed to inherits the options from the parent

> wait for the https://github.com/yarnpkg/berry/pull/2161

That PR is mostly blocked by Node, they unflagged ESM before the loader API was ready

mcollina:
I like

> A third option would be to fall back to require if import fails

a lot.

I'd prefer _not_ to add that additional dependency, but I'm also ok to add it if it's the only option.

merceyz:
> I'd prefer not to add that additional dependency, but I'm also ok to add it if it's the only option.

Adding it wouldn't help much, you can read the file by using the `fs` module but `import` can't without https://github.com/yarnpkg/berry/pull/2161

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,87 @@
...
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Use yarn
+        run: |
+          npm install -g yarn
+          yarn set version berry
+          echo "nodeLinker: pnp" >> .yarnrc.yml 
+          echo "pnpMode: loose" >> .yarnrc.yml 
+          yarn install
+          yarn add --dev transport@link:./test/fixtures/transport
```
merceyz, line 82: I meant you should do this regardless of what is running and get rid of the manual symlinking

mcollina: doing the same with npm across multiple node version is not that easy.

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,92 @@
...
+          # see https://github.com/yarnpkg/berry/issues/2935#issuecomment-911299992
+          yarn add --dev typescript@~4.3.2
+          yarn install
+        env:
+          # needed due the yarn.lock file in pino's .gitignore
+          YARN_ENABLE_IMMUTABLE_INSTALLS: false
+      - name: Tests
+        run: |
+          # yarn doesn't support ESM yet
+          mv ./test/esm/index.test.js ./test/esm/index.test.skip.js
```
merceyz, line 63:
```suggestion
```
It's PnP that doesn't support ESM yet, this is testing with `node_modules` so ESM should work just fine

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2.4.0
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Use yarn
+        run: |
+          npm install -g yarn
+          yarn set version berry
+          echo "nodeLinker: pnp" >> .yarnrc.yml 
+          echo "pnpMode: loose" >> .yarnrc.yml
```
merceyz, line 85: Why are you testing in loose mode? This will allow undeclared dependencies to go through

.github/workflows/package-manager-ci.yml
Eomm: This file load `mjs` files that trigger the error

I have changed, skipping the test in the file reading the `isYarnPnp` flag.

Sorry for the wrong comment

.github/workflows/package-manager-ci.yml
Eomm: There are some packages that do not have added their packages as dependency or peer ones. One of them is `tap`.
Fixing all of them seems very slow.

I'm trying to track `packageExtensions` but this could be massive actually

There is no option to generates the `packageExtensions` during a `yarn run ..`?

Eomm:
Quick update:

- yarn/pnp tests are green, the typescript tests are failing to this issue https://github.com/yarnpkg/berry/issues/3394
- pnpm: there is one flaky test on Node v16 I'm working on

mcollina:
I'm ok in landing this without typescript support if it's feasible.

.github/workflows/package-manager-ci.yml
merceyz: With npm you can use the `file:` protocol which has been supported since 2.0.0
https://docs.npmjs.com/cli/v7/configuring-npm/package-json#local-paths

.github/workflows/package-manager-ci.yml
merceyz: The one you currently have for `debug` is unnecessary, it's declared as an optional peer dependency already.

> There are some packages that do not have added their packages as dependency or peer ones. One of them is tap.
Fixing all of them seems very slow.

Sadly there are still packages that don't declare their dependencies correctly but if it's of any consultation it used to be a lot worse

> There is no option to generates the packageExtensions during a yarn run ..?

We haven't made a tool for it yet, no. It's a bit difficult as well since we have no clue what version or what type of dependency it is supposed to be.

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+        uses: actions/setup-node@v2.4.0
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Use yarn
+        run: |
+          npm install -g yarn
+          yarn set version berry && yarn set version 2.4.2
+          echo "nodeLinker: node-modules" >> .yarnrc.yml
+          # see https://github.com/yarnpkg/berry/issues/2935#issuecomment-911299992
+          yarn add --dev typescript@~4.3.2
```
merceyz, line 55:
```suggestion
```
We've released the fix for this issue

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+        node-version: [12, 14, 16]
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2.4.0
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Use yarn
+        run: |
+          npm install -g yarn
```
merceyz, line 51:
```suggestion
```
The GitHub actions runners already have Yarn installed

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+          echo 'nodeLinker: pnp
+          packageExtensions:
+            debug@*:
+              dependencies:
+                supports-color: "*"
+            treport@*:
+              dependencies:
+                tap-yaml: "*"
+          ' >> .yarnrc.yml
+          yarn add --dev typescript@~4.3.2
```
merceyz, line 90:
```suggestion
```

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+        node-version: [12, 14, 16]
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v2.4.0
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: Use yarn
+        run: |
+          npm install -g yarn
```
merceyz, line 79:
```suggestion
```

.github/workflows/package-manager-ci.yml
Eomm: Is it not an older version?
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md#package-management

.github/workflows/package-manager-ci.yml
Eomm: Thanks

To avoid this issue for future typescripts minor releases, where could we read the last working version integrated?

.github/workflows/package-manager-ci.yml
Eomm: Removed the loose mode 👍🏽 

The `debug` is needed otherwise I have some warning that let the tests fail

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+          yarn set version berry && yarn set version 2.4.2
```
merceyz, line 52:
```suggestion
          yarn set version berry && yarn set version 2
```
Since Yarn 3.0 has support for ranges you can run `set version 2` to get the latest 2.x version

.github/workflows/package-manager-ci.yml
merceyz: `1.22.11` is the latest 1.x version but you use `set version ...` so it doesn't really matter

.github/workflows/package-manager-ci.yml
merceyz: We try to update the patch as soon as they make a new release so we don't have it documented anywhere. Also TypeScript doesn't follow semver so their `minor` is actually a `major` release

Eomm: > TypeScript doesn't follow semver so their minor is actually a major release

For this reason, the pino's maintainer may evaluate to pin the dep to `~4.4.0` so it will be dependabot to update it and trigger the new error for new minor changes

Eomm:
PR ready

Updated first post accordingly.

for this issue

> pnpm: there is one flaky test on Node v16 I'm working on

I had to disable the coverage since this test was starving.

What I found is that:

- pnpm spawn tap
- tap spawn a coverage instance and read the stdout
- a tap test spawns a process
- the process spawns a worker_thread
- the tap test doesn't receive any `close` event

I did not investigate further

https://github.com/pinojs/pino/blob/f712a52e05c633246626f6b888298cf933a3b872/test/transport.test.js#L372-L382

mcollina (changes requested):
I think it might be too much to run this for every PR, those are a lot of tests. Maybe could we just run them.

As an example, we run this only on `master`: https://github.com/fastify/fastify/blob/main/.github/workflows/package-manager-ci.yml#L3-L6

I would also run this only on Node.js v14.

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Tests
+        run: yarn run test-ci
+
+  yarn-pnp:
+    name: yarn-pnp package manager on ${{ matrix.node-version }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest]
```
mcollina, line 68: why only on mac?

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+      - name: Tests
+        run: pnpm run test-ci-pnpm
+
+  yarn:
+    name: yarn package manager on ${{ matrix.node-version }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest]
```
mcollina, line 41: why only on mac?

.github/workflows/package-manager-ci.yml
```diff
@@ -0,0 +1,95 @@
...
+        - 'docs/**'
+        - '*.md'
+jobs:
+  pnpm:
+    name: pnpm package manager on ${{ matrix.node-version }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest, windows-latest]
```
mcollina, line 18: why only on mac and linux?

.github/workflows/package-manager-ci.yml
Eomm: Forgot to add windows

Replaced max with ubuntu to test is on a UNIX system

package.json
```diff
@@ -24,6 +24,8 @@
     "lint": "eslint .",
     "test": "npm run lint && tap test/*test.js test/*/*test.js && npm run test-types",
     "test-ci": "npm run lint && tap --no-check-coverage test/*test.js test/*/*test.js --coverage-report=lcovonly && npm run test-types",
+    "test-ci-pnpm": "npm run lint && tap --no-coverage --no-check-coverage test/*test.js test/*/*test.js && npm run test-types",
+    "test-ci-yarn-pnp": "npm run lint && tap --no-check-coverage test/*test.js test/*/*test.js --coverage-report=lcovonly",
```
merceyz, line 28:
```suggestion
    "test-ci-yarn-pnp": "yarn run lint && tap --no-check-coverage test/*test.js test/*/*test.js --coverage-report=lcovonly",
```

package.json
```diff
@@ -24,6 +24,8 @@
     "lint": "eslint .",
     "test": "npm run lint && tap test/*test.js test/*/*test.js && npm run test-types",
     "test-ci": "npm run lint && tap --no-check-coverage test/*test.js test/*/*test.js --coverage-report=lcovonly && npm run test-types",
+    "test-ci-pnpm": "npm run lint && tap --no-coverage --no-check-coverage test/*test.js test/*/*test.js && npm run test-types",
```
merceyz, line 27:
```suggestion
    "test-ci-pnpm": "pnpm run run lint && tap --no-coverage --no-check-coverage test/*test.js test/*/*test.js && pnpm run test-types",
```

test/transport.test.js
```diff
@@ -15,6 +15,32 @@ const writer = require('flush-write-stream')
...
+      throw err
+    }
+  }
+}
+
+async function uninstallTransportModule () {
+  try {
+    await unlink(join(__dirname, '..', 'node_modules', 'transport'))
+  } catch (err) {
+    if (!isYarnPnp) {
```
mcollina, line 38: it's better to avoid the unlink if `isYarnPnp`, i.e. return at the start of the function.

test/transport.test.js
```diff
@@ -15,6 +15,32 @@ const writer = require('flush-write-stream')
...
+  try {
+    await uninstallTransportModule()
+  } catch {}
+  try {
+    await symlink(
+      join(__dirname, 'fixtures', 'transport'),
+      join(__dirname, '..', 'node_modules', 'transport')
+    )
+  } catch (err) {
+    if (!isYarnPnp) {
```
mcollina, line 28: it's better to avoid the unlink if `isYarnPnp`, i.e. return at the start of the function.

test/transport.test.js
```diff
@@ -83,6 +102,29 @@ test('pino.transport with package', { skip: isWin }, async ({ same, teardown })
...
+  transport.on('ready', function () {
+    fail('ready event should not be emitted')
+  })
+  transport.on('error', function (err) {
+    ok(err)
+    equal(err.message, 'Missing node(s) option', 'pino-elastisearch trigger a valid error')
+  })
+
+  pass('pino instance created')
+})
```
mcollina, line 126: I would remove this test. We should ideally cover it with the yarn pnp automatic linking as it's done for transport.

mcollina (changes requested):

test/transport.test.js
Eomm: Do you mean running `yarn add pino-elasticsearch`?

Right now this test is equal to the `pino-pretty` one, since the last merge on master removed the `#pino-pretty` hard coding case

mcollina: then remove the test completely!

mcollina (approved):
lgtm

PR #1116, opened 2021-09-04
Add overload to handle unknown returned by TS-4.4 for errors be default

### LINKED ISSUE #1115: TypeScript 4.4 breaks logging exceptions from catch
https://github.com/Microsoft/TypeScript-wiki/blob/main/Breaking-Changes.md#using-unknown-in-catch-variables

> Users running with the --strict flag may see new errors around catch variables being unknown due to the new --useUnknownForCatchVariables flag, especially if the existing code assumes only Error values have been caught.

This causes issues with the typing in `pino.error(exception)` for example.

```plain
TS2769: No overload matches this call.
Overload 1 of 2, '(obj: object, msg?: string | undefined, ...args: any[]): void', gave the following error.
Argument of type 'unknown' is not assignable to parameter of type 'object'.
Overload 2 of 2, '(msg: string, ...args: any[]): void', gave the following error.<br/>Argument of type 'unknown' is not assignable to parameter of type 'string'.
```

Of course it works, because in this particular scenario the `error` is an actual thrown `Error`. But this change from TypeScript is because you can in theory throw anything. :shrug:

kibertoad:
fixes #1115

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
     "through2": "^4.0.0",
     "ts-node": "^10.2.1",
     "tsd": "^0.17.0",
-    "typescript": "4.4.2",
+    "typescript": "^4.4.2",
     "winston": "^3.3.3"
   },
   "dependencies": {
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
   },
   "homepage": "http://getpino.io",
   "devDependencies": {
-    "@types/node": "^16.7.2",
+    "@types/node": "^16.7.10",
     "airtap": "4.0.3",
     "benchmark": "^2.1.4",
     "bole": "^4.0.0",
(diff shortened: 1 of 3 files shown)


kibertoad:
@climba03003 Could you please take a look?

mcollina (approved):
lgtm

package.json
```diff
@@ -94,7 +94,7 @@
     "through2": "^4.0.0",
     "ts-node": "^10.2.1",
     "tsd": "^0.17.0",
-    "typescript": "^4.4.2",
+    "typescript": "4.4.2",
```
mcollina, line 97: Can you please restore the caret here?

mcollina (changes requested):

package.json
kibertoad: Done. You prefer random early failures when TS updates to explicit updates? There are benefits to both approaches, I still don't know which one makes more sense to me personally.

mcollina: I just think TS is too much maintenance on dependabot/PRs to manage in that way. If our types start breaking suddenly we'll address it, however that has not been a problem in Fastify so the approach seem to work ok.

mcollina (approved):
lgtm

PR #1120, opened 2021-09-08
[v6.x] set max depth and edge limit

### LINKED ISSUE #990: CPU Hang with infinite decirc
Hello,

I experienced downtime from pino logger.
A developer accidentally logged an express `req` in pino, and on catch It made our api hang with full CPU.

Here a small exemple of the expressjs middleware used : 
```js
import { NextFunction, Request, Response } from 'express'
import pino from 'pino'

export const webhook = (
  req: Request,
  res: Response,
  next: NextFunction,
) => {
  const logger = pino({
    name: 'test',
    prettyPrint: {
      colorize: true,
      translateTime: true,
      messageFormat: '{hostname} - {pid} - {msg}',
      ignore: 'hostname,pid',
    },
  })

  logger.error({ req })
}
```

**Node** version : 12.9.0
**Typescript** : 4.1.3
**Pino** : 6.11.2
**Pino-pretty** : 4.7.1
**Express** : 4.17.1

Here is a profiler trace
Capture d’écran 2021-03-24 à 13 51 32

mcollina:
Fixes #990

We might want to expose this as config values.

(Note that we cannot forward-port this change to v7.x, as we switched the serialization engine)

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -25,6 +25,11 @@ const {
   messageKeySym
 } = require('./symbols')
 
+const defaultStringifyOptions = {
+  depthLimit: 100,
+  edgelimit: 200
+}
+
 function noop () {}
 
 function genLog (level, hook) {
@@ -401,7 +406,7 @@ function stringify (obj) {
   try {
     return JSON.stringify(obj)
   } catch (_) {
-    return stringifySafe(obj)
+    return stringifySafe(obj, undefined, undefined, defaultStringifyOptions)
   }
 }
 


kibertoad (approved):

jsumners (approved):

simoneb (approved):

simoneb:
@marcbachmann I think they can be considered independent. The only thing we should try to get some consistency on is the default limits, which are different between the 2 PRs. In #1169 we set [these](https://github.com/pinojs/pino/pull/1169/files#diff-e6c216f4530aee9459ebefff9e14eabea747ef62bedcfc4fcf8803b8c007dcf3R73-R74) because they are fixing the original issue #990 but I'm not 100% sure they are sensible defaults.

PR #1124, opened 2021-09-09
Removed deprecated options in pino.child()

mcollina:


jsumners (approved):
We must make sure to highlight this in the release notes.

kibertoad:
@mcollina Should bindings be completely dropped from typings, or there are still cases when they are used?

mcollina:
I think the typings are already correct.

PR #1125, opened 2021-09-09
Update the `options.level` documentation of the child logger

marcbachmann:
Related PR: https://github.com/pinojs/pino/pull/1124

The in here bring the documentation up to date with the recent deprecation of `bindings.level`.

```
(node:15042) [PINODEP007] PinoWarning: bindings.level is deprecated, use options.level option instead
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -678,7 +678,7 @@ the current log level of the parent at the time they are spawned.
 
 The log level of a child is mutable. It can be set independently
 of the parent either by setting the [`level`](#level) accessor after creating
-the child logger or using the reserved [`bindings.level`](#bindingslevel-string) key.
+the child logger or using the [`options.level`](#optionslevel-string) key.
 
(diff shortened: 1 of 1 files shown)


jsumners (commented):
Shouldn't this target the v6 branch? Isn't this depreciation already removed on the master branch?

mcollina (approved):
lgtm

@jsumners This behavior is _still_ in v7. We introduce the deprecation not so long ago... should we remove it anyway?

jsumners:
> lgtm
> 
> @jsumners This behavior is _still_ in v7. We introduce the deprecation not so long ago... should we remove it anyway?

Yes. No need to carry the baggage for another major since we already alert on the current major.

mcollina:
This PR should target this branch anyway as the doc is not up to date with the deprecation. We also need to cherry-pick this commit to v6.

jsumners (approved):

PR #1127, opened 2021-09-12
fix: make pino.final sync flushes when no handler provided (#1126)

### LINKED ISSUE #1126: `pino.final` without a handler and `stream.flushSync()`
I discovered that pino.final can be used without a handler, which I find very handy. Example:

```js
const handler = (err) => {
  const finalLogger = pino.final(logger);
  
  if (err) finalLogger(err);
  process.exit(err ? 1 : 0);
};
```

But reading the code I found something that might be a bug and perhaps someone here can confirm it.

https://github.com/pinojs/pino/blob/c57c25855d89a2c481914672bbdeebe7ec3087c7/lib/tools.js#L463-L478

`stream.flushSync()` is only called before the handler is executed or after `finalLogger` is used. In the example above there is no handler and `finalLogger` is only called if there is an error, so when the process exits normally there is no call to `stream.flushSync()`.

Perhaps the implemention when there is no handler should be this

```js
  if (!hasHandler) {
    try {
      stream.flushSync()
    } catch (e) {
      // it's too late to wait for the stream to be ready
      // because this is a final tick scenario.
      // in practice there shouldn't be a situation where it isn't
      // however, swallow the error just in case (and for easier testing)
    }
    return finalLogger
  }
```

javiertury:
Fixes #1126

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -461,6 +461,14 @@ function final (logger, handler) {
   })
 
   if (!hasHandler) {
+    try {
+      stream.flushSync()
+    } catch (e) {
+      // it's too late to wait for the stream to be ready
+      // because this is a final tick scenario.
+      // in practice there shouldn't be a situation where it isn't
+      // however, swallow the error just in case (and for easier testing)
+    }
     return finalLogger
   }
 
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -463,7 +463,7 @@ function final (logger, handler) {
   if (!hasHandler) {
     try {
       stream.flushSync()
-    } catch (e) {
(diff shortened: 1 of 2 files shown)


lib/tools.js
```diff
@@ -461,6 +461,14 @@ function final (logger, handler) {
   })
 
   if (!hasHandler) {
+    try {
+      stream.flushSync()
+    } catch (e) {
```
jsumners, line 466:
```suggestion
    } catch {
```

mcollina (approved):
lgtm

jsumners (approved):

PR #1134, opened 2021-09-20
Removed coveralls, removed TS badge

mcollina:


kibertoad:
Why remove coverage?

mcollina:
1. Coveralls is continuosly down
2. We enforce 100% code coverage via tap

kibertoad:
1. They were doing maintenance, it's almost done by now: https://status.coveralls.io/
2. Can we add some badge for that?

mcollina:
@kibertoad they went down on the 18th and it's the 20th and they did not come back yet. I'm not sure we can depend on them.

We do not have a badge for it... I'd be happy for suggestions.

jsumners (approved):

Issue #1138, opened 2021-09-23
Pino.transport seems to be dropping logs?


bitjson:
Thanks for the fantastic library and documentation! I've been happily using it for a few years, and this is the first meaningful issue I've had (after upgrading from `pino@6.11.3` +`pino-multi-stream@5.3.0` to `pino@7.0.0-rc.6` – hopefully I'm just misunderstanding configuration for v7).

I have a server which constantly fetches data from another process and stores it in a database (it attempts to fetch only as fast as its able to store). It's using the following configuration:

```ts
const logPath = 'some/path';
const logLevelStdout = 'trace';

const transport = pino.transport({
  targets: [
    {
      level: 'debug',
      options: { destination: logPath },
      target: 'pino/file',
    },
    {
      level: logLevelStdout,
      options: { destination: 1 },
      target: 'pino/file',
    },
  ],
});

export const logger = pino({ level: logLevelStdout }, transport);

// elsewhere in the codebase:

logger.info('foo');
logger.debug('bar');
logger.trace('baz'); // etc.
```

When running the server in production, I've noticed that **some logging messages never make it to stdout or the log file**. It seems to happen randomly: a particular phase of the fetch/store process will log properly many times, then every few minutes, a few log messages in the sequence will simply be missing.

I know the server is otherwise behaving correctly – if the logging code wasn't hit, the rest of the fetch/store process would be failing. E.g. for a logging message which logs a bit of data right before that data is stored by a database insert, I can later see that the data was stored even though the preceding message was never logged.

Another data point: after the process runs for a while, it seems like logging occasionally stalls for a few minutes, then the logs from those few minutes seems to be dumped all at once. (To both stdout and the file, which I'm `tail -f`ing in another pane.)

I have e2e tests for the server (using [Ava](https://github.com/avajs/ava) in serial mode) which spin it up (using [Execa](https://github.com/sindresorhus/execa)), emulate the external data sources, and connect to a real Postgres instance. Fortunately **the e2e tests drop some logs pretty reliably** – one of a few lines always seems to be dropped right near the beginning of the tests (and the tests otherwise proceed normally, so I think the server is working as expected).

Unfortunately, when I attempt to debug the issue around those lines in VSCode, no logs are dropped. It seems to only happen at full speed.

Does anyone have any recommendations for how to debug this sort of issue?

bitjson:
[...]

PR #1140, opened 2021-09-24
Remove in-between Transform for file.js

mcollina:
I have been investigating #1138 for dropped logs, and I found _some_ during process shutdown. Specifically, it is due to the  `Transform` stream emitting `close` while the destination was still shutting down.

We would need some better mechanism to signal pino "the whole pipeline has been shutdown".

### DIFF
--- a/.taprc
+++ b/.taprc
@@ -1,2 +1,3 @@
 coverage: true
 timeout: 480
+check-coverage: false
--- a/file.js
+++ b/file.js
@@ -1,20 +1,10 @@
 'use strict'
 
-const { Transform, pipeline } = require('stream')
 const pino = require('./pino')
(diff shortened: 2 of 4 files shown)


.taprc
```diff
@@ -1,2 +1,3 @@
 coverage: true
 timeout: 480
+check-coverage: false
```
mcollina, line 3: I'm not sure why but the coverage went down, so I'm disabling this for now.

mcollina:
@Eomm check this out. I feel a few of the other transports would suffer from similar problems.

jsumners (approved):

mcollina:
The fun part about this is that on my system we start seeing it after _8_ million lines written.

PR #1143, opened 2021-09-30
Transport pipeline

### LINKED ISSUE #1142: How can I use pino-socket with pino-syslog?
I have a test app:

```typescript
'use strict';

const pino = require('pino');
  targets: [
    {
      target: 'pino-syslog',
    },
    {
      target: 'pino-socket',
      options: {
        address: '127.0.0.1',
        port: 5140,
        mode: 'tcp',
      },
    },
  ],
});
const logger = pino(transports);

logger.error('Test message');
```

When I run it, the syslog-formatted message logs to stdout:

```
<131>1 2021-09-28T19:15:09Z bmaupin-laptop - 218252 - - {"level":50,"time":1632856509189,"pid":218252,"hostname":"workstation","msg":"Test message"}
```

And the syslog server instead receives an unformatted log:

```
$ nc -kl 5140
{"level":50,"time":1632856509189,"pid":218252,"hostname":"workstation","msg":"Test message"}
```

These are the package versions I'm using:

```
    "pino": "^7.0.0-rc.7",
    "pino-socket": "^4.0.0",
    "pino-syslog": "github:Eomm/pino-syslog#pino-7",
```

Thanks!

mcollina:
Fixes #1142

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -188,6 +188,8 @@ const logger = pino({
     pipeline: [{
       target: './my-transform.js'
     }, {
+      // Use target: 'pino/file' to write to stdout
+      // without any change.
       target: 'pino-pretty
     }]
   }
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -196,7 +196,7 @@ const logger = pino({
 logger.info('hello world')
 ```
 
-__NOTE: there is no "default" destinations for a pipeline but
+__NOTE: there is no "default" destination for a pipeline but
 a terminating target, i.e. a `Writable` stream.__
 
 ### Notable transports
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -143,6 +143,62 @@ module.exports = function (opts) {
 
 (It is possible to use the async iterators with CommonJS and streams with ESM.)
 
+### Creating a transport pipeline
+
+As an example, the following transport returns a `Transform` stream:
+
+```js
+import build from 'pino-abstract-transport'
+import { pipeline, Transform } from 'stream'
+exports default async function (options) {
+  return build(function (source) {
+    const myTransportStream = new Transform({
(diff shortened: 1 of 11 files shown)


mcollina:
cc @jsumners @Eomm

lib/worker-pipeline.js
```diff
@@ -0,0 +1,49 @@
...
+        throw error
+      }
+    }
+    const stream = await fn(t.options)
+    return stream
+  }))
+  const ee = new EE()
+
+  const stream = new PassThrough({
+    destroy (err, cb) {
```
Eomm, line 31:
```suggestion
    autoDestroy: true,
    destroy (err, cb) {
```

mcollina: thx

jsumners (changes requested):
Need doc updates.

mcollina:
@jsumners added docs.

docs/transports.md
```diff
@@ -143,6 +143,62 @@ module.exports = function (opts) {
...
+      target: 'pino-pretty
+    }]
+  }
+})
+
+logger.info('hello world')
+```
+
+__NOTE: there is no "default" destinations for a pipeline but
+a terminating target, i.e. a `Writable` stream.__
```
jsumners, line 200:
```suggestion
__NOTE: there is no "default" destination for a pipeline but
a terminating target, i.e. a `Writable` stream.__
```

Should we expose some usable destination for pipelines that dumps to `stdout`? Something like:

```js
pipeline:[
  {  /* custom transform */ },
  { target: pino.pipelineStdoutTerminator }
]
```

mcollina: We already have that, it's `pino/file`. By default it writes to stdout.

jsumners (approved):

PR #1152, opened 2021-10-12
Fix types for `pino.multistream`

clemyan:
Fixes TypeScript types for `pino.multistream`.

Also, the implementation of `pino.multistream` is actually pretty loose with the first argument. It can handle:

- An object with `stream` and optionally `level` properties (i.e. a `StreamEntry`)
- An object with a `write` method (i.e. a `DestinationStream`)
- An array of the above

However, only array of `StreamEntry` is documented, and the current types reflect that. I don't know if that is intentional, so I have not changed that for now. If not, then the correct type should be `(DestinationStream | StreamEntry)[] | DestinationStream | StreamEntry`

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -246,7 +246,8 @@ declare namespace P {
     }
 
     function multistream(
-        streamsArray: StreamEntry[], opts?: P.MultiStreamOptions
+        streamsArray: (DestinationStream | StreamEntry)[] | DestinationStream | StreamEntry,
+        opts?: P.MultiStreamOptions
     ): MultiStreamRes
 
     /**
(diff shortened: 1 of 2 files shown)


clemyan:
> I'm lost. I think the types are ok as is, but maybe I'm missing some detail.

Well, in terms of the implementation, `pino.multistream` is pretty liberal with it first argument. Each of the 

```
pino.multistream(process.stdout) // Single stream
pino.multistream([process.stdout]) // Array of streams
pino.multistream({ stream: process.stdout, level: 'info' }) // Single StreamEntry
pino.multistream([{ stream: process.stdout, level: 'info' }]) // Array of StreamEntries

// Mixed array
pino.multistream([
	process.stdout,
	{ stream: process.stderr, level: 'error' },
]) 
```

However, only the "Array of StreamEntries" usage is documented. So the TypeScript types in this PR only allows array of StreamEntries. Question is whether the types should allow any or all of the other cases.

clemyan:
I have added the other signatures to the types and tests.

PR #1156, opened 2021-10-14
Add transport option param to types

kibertoad:
refs https://github.com/pinojs/pino-pretty/issues/248

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
   },
   "homepage": "http://getpino.io",
   "devDependencies": {
-    "@types/node": "^16.9.4",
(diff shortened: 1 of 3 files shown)


kibertoad:
@mcollina You mean this?
```
import pino from 'pino'

const logger = pino({
  transport: {
    pipeline: [{
      target: './my-transform.js'
    }, {
      // Use target: 'pino/file' to write to stdout
      // without any change.
      target: 'pino-pretty'
    }]
  }
})

logger.info('hello world')
```

I'll check, probably still needs to be added

mcollina:
yes

Issue #1104, opened 2021-08-28
test pino.transport() with bundlers


mcollina:
A few of our users bundle their applications. While I recommend against doing it, we should verify pino.transport works properly for them too.

Unga:
> While I recommend against doing it

Why do you recommend not doing bundling?

mcollina:
> Why do you recommend not doing bundling?

It's not needed and it complicates debugging and resolving production issues.

Issue #1157, opened 2021-10-14
Pino v7 shows "Cannot find module 'lib/worker.js' when bundling it with Rollup


pan93412:
## Brief Description

`Rollup` can't bundle the `worker.js` of Pino v7 properly. [There may be a Rollup plugin for bundling worker.js](https://www.npmjs.com/package/rollup-plugin-web-worker-loader), but Pino seems like not compatible with this.

```
➜  UnblockNeteaseMusic git:(feat/pino-7) ✗ node app.js

node:internal/event_target:777
  process.nextTick(() => { throw err; });
                           ^
Error: Cannot find module '/Users/pan93412/Projects/UnblockNeteaseMusic/precompiled/lib/worker.js'
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:79:12)
    at MessagePort.<anonymous> (node:internal/main/worker_thread:187:24)
    at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
    at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28)
Emitted 'error' event on Worker instance at:
    at Worker.[kOnErrorMessage] (node:internal/worker:289:10)
    at Worker.[kOnMessage] (node:internal/worker:300:37)
    at MessagePort.<anonymous> (node:internal/worker:201:57)
    at MessagePort.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
    at MessagePort.exports.emitMessage (node:internal/per_context/messageport:23:28) {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
```

## Reproducible code and step

https://github.com/UnblockNeteaseMusic/server/blob/316766433de074ff94f120f515e058de63dae560/src/logger.js

```bash
git clone https://github.com/UnblockNeteaseMusic/server
cd server
git checkout 316766433de074ff94f120f515e058de63dae560
yarn
yarn build
node precompiled/app.js # ⚠️
```

Everspace:
This is also a problem with esbuild for console applications.

davipon:
@Everspace can you give [esbuild-plugin-pino](https://github.com/davipon/esbuild-plugin-pino) a try?
I just wrote it to solve my work project. Like to know if it misses any special case.

[...]

PR #1161, opened 2021-10-15
feat: add bundlers test stack

zekth:
fixes: https://github.com/pinojs/pino/issues/1104

The test will fail as long as https://github.com/pinojs/pino/issues/1157 is not fixed.


mcollina:
I would like to land this but keep PRsa and commit being "green"

zekth:
> Awesome! Do you have a clue on how to fix it?

I have ideas, currently working on a POC.

Issue #1168, opened 2021-10-18
log loss docs outdated?


ronag:
https://github.com/pinojs/pino/blob/master/docs/asynchronous.md#log-loss-prevention

I thought one of the things pino 7 fixed was that we no longer need final?

PR #1171, opened 2021-10-18
Bring the asynchronous logging doc up to v7

mcollina:
Fixes https://github.com/pinojs/pino/issues/1168


mcollina:
cc @ronag

mcollina:
There is no alternative on Node v12 and we still support that, we'll
deprecate in the next cycle.

PR #1172, opened 2021-10-18
Fixes #1167 (dedupe should write to all streams matching the target level)

### LINKED ISSUE #1167: Same level multistream logging is not working with "dedupe"
If I set `dedupe` to true, the option works as stated in the docs, but it also _breaks_ the possibility to have multiple streams for the same level, as only the "first" stream of the given level will be used, not all the streams matching the level. I'm not sure if this is the intended way of working or this is something you might want to change. I understand that the way it currently works exactly means "deduping" but in my (and I think in most) use cases, we would expect dedupe to "send logs only to the stream with the higher level" but to **all** of them.

```javascript
var fs = require('fs')
var pino = require('pino')
var pretty = require('pino-pretty')
var streams = [
  { level: 'info', stream: pretty() },
  { level: 'info', stream: fs.createWriteStream(`./log-info.log`) },
  { level: 'error', stream: fs.createWriteStream(`./log-error.log`) },
]

var multistreamOpts = {
    dedupe: true,
};

var log = pino({
  level: 'info' // this MUST be set at the lowest level of the destinations
}, pino.multistream(streams, multistreamOpts))

log.info('this will be written ONLY to the console by pino-pretty and NOT in ./log-info.log')
log.fatal('this will be written ONLY to ./log-error.log')
```

---

I think changing the [current multistream write function](https://github.com/pinojs/pino/blob/588b4858ebcc65f67ceac40c1c4a10452a84dc75/lib/multistream.js#L45-L73) to the following would do the trick:

```javascript
function write (data) {
  //...
  for (let i = 0; i < streams.length; i++) {
    dest = streams[i]
    if (dest.level <= level) {
      //...
      if (!opts.dedupe || dest.level === level) { //change 'if' condition
        stream.write(data)
      }
    } else {
      break
    }
  }

  //TODO: remove
  //if (opts.dedupe && stream) {
  //  stream.write(data)
  //}
}
```

frzsombor:
Fixes #1167

### DIFF
--- a/test/multistream.test.js
+++ b/test/multistream.test.js
@@ -429,6 +429,48 @@ test('dedupe', function (t) {
   t.end()
 })
 
+test('dedupe when some streams has the same level', function (t) {
+  let messageCount = 0
+  const stream1 = writeStream(function (data, enc, cb) {
+    messageCount -= 1
+    cb()
+  })
+
+  const stream2 = writeStream(function (data, enc, cb) {
+    messageCount += 1
+    cb()
+  })
+
+  const stream3 = writeStream(function (data, enc, cb) {
+    messageCount += 1
+    cb()
+  })
+
+  const streams = [
+    {
+      stream: stream1,
+      level: 'info'
+    },
+    {
+      stream: stream2,
+      level: 'fatal'
+    },
+    {
+      stream: stream3,
+      level: 'fatal'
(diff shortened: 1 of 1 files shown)


frzsombor:
I've added a new test case, hope that it is something you expected to see

PR #1180, opened 2021-10-24
Coerce string integer destinations to file descriptors

jsumners:
It's easy to accidentally write code like the following and expect it "to work":

```js
'use strict'

const pino = require('pino')
  targets: [
    { target: 'pino/file', level: 'debug', options: { destination: '2' } },
    { target: 'pino/file', level: 'info', options: { destination: '1' } }
  ]
})

const log = pino({ level: 'debug' }, transport)

log.error('something bad')
log.info('it works')
```

Notice that the `destination` in each target is a string integer. The clear intention is that they write to `stderr` and `stdout` respectively. This pull request accounts for this.

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -262,7 +262,7 @@ const transport = pino.transport({
 pino(transport)
 ```
 
-The `options.destination` property may also be a number to represent a file descriptor. Typically this would be `1` to write to STDOUT or `2` to write to STDERR. If `options.destination` is not set, it defaults to `1` which means logs will be written to STDOUT.
+The `options.destination` property may also be a number to represent a filedescriptor. Typically this would be `1` to write to STDOUT or `2` to write to STDERR. If `options.destination` is not set, it defaults to `1` which means logs will be written to STDOUT. If `options.destination` is a string integer, e.g. `'1'`, it will be coerced to a number and used as a file descriptor. If this is not desired, provide a full path, e.g. `/tmp/1`.
 
 The difference between using the `pino/file` transport builtin and using `pino.destination` is that `pino.destination` runs in the main thread, whereas `pino/file` sets up `pino.destination` in a worker thread.
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -422,7 +422,7 @@ Default: `pino.destination(1)` (STDOUT)
 The `destination` parameter, at a minimum must be an object with a `write` method.
 An ordinary Node.js `stream` can be passed as the destination (such as the result
 of `fs.createWriteStream`) but for peak log writing performance it is strongly
-recommended to use `pino.destination` to create the destination stream. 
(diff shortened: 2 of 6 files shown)


docs/transports.md
```diff
@@ -262,7 +262,7 @@ const transport = pino.transport({
 pino(transport)
 ```
 
-The `options.destination` property may also be a number to represent a file descriptor. Typically this would be `1` to write to STDOUT or `2` to write to STDERR. If `options.destination` is not set, it defaults to `1` which means logs will be written to STDOUT.
+The `options.destination` property may also be a number to represent a filedescriptor. Typically this would be `1` to write to STDOUT or `2` to write to STDERR. If `options.destination` is not set, it defaults to `1` which means logs will be written to STDOUT. If `options.destination` is a string integer, e.g. `'1'`, it will be coerced to a number and used as a file descriptor. If this is not desired, provide a full path, e.g. `/tmp/1`.
```
mcollina, line 265: The docs for pino.destination should be updated too.

mojavelinux:
I don't understand why we need to coerce a string to a number. Isn't the differentiation between a file path and file descriptor the type? What if you want to write to a file that is named "1". Now you lose that ability. For logs, that's not an unreasonable scenario (though probably rare) since logs are often sequenced in a directory. Can't we make the user get the type right?

docs/transports.md
jsumners: 🤦‍♂️ will do. I literally went back to bed after this.

jsumners:
The included docs show how to write to a file named `'1'`.

mojavelinux:
That just seems really weird to me that we are not accepting what the user is providing. Sure, we can try to analyze the value and figure whether it is one or the other, but the type was already doing that.

jsumners:
No, it wasn't. This is a reduction in surprise. Passing a single integer as a destination is a mistake and most likely an attempt to write to stdout or stderr. If the string is accepted as-is, it is not defined what will actually happen. If it _does_ write to a file, then that file can be in the current working directory or somewhere else entirely. If a file is the intended destination then an actual file path should be specified.

mojavelinux:
We can agree to disagree, but I think the current behavior is the logical one. A integer is a file descriptor (which might be stdout, stderr, or one the application is managing). A string is a file path. As a user of pino, that's how I understood it and that contract makes complete sense to me.

mcollina (approved):
lgtm

mojavelinux:
After sitting on this for awhile, I now agree this is consistent with the *nix convention. I retract my earlier objection. Thanks for making this improvement.

jsumners:
> After sitting on this for awhile, I now agree this is consistent with the *nix convention. I retract my earlier objection. Thanks for making this improvement.

Thank you.

PR #1184, opened 2021-10-25
style(test/types): make implicitly inserted semicolons explicit

Fdawgs:
All of the other statements in the file have explicit semicolons; updated to be consistent.


Fdawgs:
> 
> 
> Why is this different from the rest of the code base?

You've lost me here.

mcollina:
We have no semicolons in the js files and it is enforced by the linter.

Fdawgs:
> 
> 
> We have no semicolons in the js files and it is enforced by the linter.

Ah, okay I see what you mean. Will rejigg this PR to make all the explicit semicolons implicit to match the rest of the codebase.

PR #1199, opened 2021-10-29
Deprecation warning for pino.final() in Node v14+

### LINKED ISSUE #1175: Add a deprecation warning for pino.final() in Node v14+
> Correct! How about we emit the warning if on Node v14+?

Seems acceptable.

_Originally posted by @jsumners in https://github.com/pinojs/pino/issues/1171#issuecomment-947647249_

sameer-coder:
This PR adds 
- Deprecation warning for pino.final() in Node v14+

Closes #1175


mcollina (approved):
lgtm

pino.js
```diff
@@ -100,6 +101,9 @@ function pino (...args) {
     edgeLimit
   } = opts
 
+  const major = Number(process.versions.node.split('.')[0])
+  if (major >= 14) warning.emit('PINODEP009')
```
jsumners, line 105: How is this limited to the usage of `pino.final`? This looks to me like it will warn on any usage of `pino()`.

jsumners (changes requested):

pino.js
sameer-coder: @jsumners  I was under the assumption that this was a one-time warning to be displayed when the logger is initialized rather than printing when`pino.final` actually gets called. I will change the code to move it to `final()`

jsumners: Thank you. If someone isn't directly using `pino.final` then this warning will be nothing more than an annoyance to them.

mcollina: good spot @jsumners

test/final.test.js
```diff
@@ -8,6 +8,35 @@ const { sleep, getPathToNull } = require('./helper')
...
+
+  function onWarning (warning) {
+    isWarningEmitted = true
+    equal(warning.code, 'PINODEP009')
+  }
+
+  if (major >= 14) process.on('warning', onWarning)
+
+  instance.info('hello')
+  await sleep(10)
```
jsumners, line 31: This seems fragile. Instead of using an `async` test, how about using the callback form and issuing `t.end()` when the test has been completed?

sameer-coder: Makes sense. I have changed the code to call `t.end()` in the warning callback instead of using sleep.

test/final.test.js
```diff
@@ -8,6 +8,28 @@ const { sleep, getPathToNull } = require('./helper')
 // will be emitted. Let's raise this so we do not scare everybody.
 process.setMaxListeners(100)
 
+test('should show warning for pino.final on node 14+', async ({ equal, end }) => {
+  const major = Number(process.versions.node.split('.')[0])
+  if (major < 14) end()
+
+  const dest = pino.destination({ dest: getPathToNull(), sync: false })
```
jsumners, line 15:
```suggestion
test('should show warning for pino.final on node 14+', ({ equal, end }) => {
  const major = Number(process.versions.node.split('.')[0])
  if (major < 14) end()

  t.plan(1)
  const dest = pino.destination({ dest: getPathToNull(), sync: false })
```

jsumners (changes requested):
One final thing: let's tell tap how many assertions we expect in the case where the warning should be issued. This is beneficial for clarity in a callback style test.

test/final.test.js
jsumners: @sameer-coder you missed removing the `async` keyword. I believe this is causing the failure https://github.com/pinojs/pino/runs/4080672051?check_suite_focus=true#step:5:1729

sameer-coder: @jsumners I think it's because of `if (major < 14) end()`. The test still continues `plan(1)`. I will add a `return` statement there and also remove the `async` keyword.

jsumners (approved):
Looks good to me.

PR #1203, opened 2021-10-31
Improve docs and deprecation about prettyPrint

mcollina:


examples/basic.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+// const pino = require('..')(require('pino-pretty')())
```
kibertoad, line 3: should this be commented out?

examples/basic.js
```diff
@@ -1,5 +1,6 @@
...
 const pino = require('..')()
```
jsumners, line 4:
```suggestion
const pino = require('..')()
```

or

```suggestion
// Pino's primary usage writes ndjson to `stdout`:
const pino = require('..')()

// However, if "human readable" output is desired,
// `pino-pretty` can be provided as the destination
// stream by uncommenting the following line in place
// of the previous declaration:
// const pino = require('..')(require('pino-pretty')())

```

jsumners (approved):

PR #1204, opened 2021-11-01
Removed console.log() from transport doc

### LINKED ISSUE #1200: Missing log output when using custom transports
Errors don't end up in `stdout` when using pino-pretty or a custom transport after catching en error (e.g. typical express + mongoose app setup)

The following is the `README.md` from [this issue's repro repository](https://github.com/replete/pino-missing-log-output), which explains the problem:

<hr/>

# Missing log output issue

This repo has three express applications, each identical apart from the configuration passed into `pino()`.

The problem is that when using a custom transport, in this scenario, no output is logged.

```js
app.listen(3000, async () => {
	log.info("API listening...");
	await connectDatabase();
});

async function connectDatabase() {
	try {
		// Imagine this is mongoose.connect(), and its thrown an
		// Authentication Failed error, but no errror is logged. 
		throw Error('some error')
		log.info("Connected to Database...");
	} catch (error) {
		log.error(error); // never logged with custom pino transports... 
		process.exit(1);
	}
}
```

### Install
``` bash
git clone git@github.com:replete/pino-missing-log-output.git
cd pino-missing-log-output
npm install
```

### Run
```bash
npm run appWithPino
npm run appWithPinoPretty
npm run appWithPinoCustom
```

## Pino with default configuration:

```js
const log = pino();
```

Running `npm run appWithPino` results in expected behaviour.

```bash
$ npm run appWithPino

> pino-problem@1.0.0 appWithPino
> node appWithPino.js

{"level":30,"time":1635621351990,"pid":83045,"hostname":"Seneca","msg":"App start"}
{"level":30,"time":1635621351996,"pid":83045,"hostname":"Seneca","msg":"API listening..."}
{"level":50,"time":1635621351996,"pid":83045,"hostname":"Seneca","err":{"type":"Error","message":"some error","stack":"Error: some error\n    at connectDatabase (/Users/phil/dev/pino-problem/appWithPino.js:17:9)\n    at Server.<anonymous> (/Users/phil/dev/pino-problem/appWithPino.js:12:8)\n    at Object.onceWrapper (node:events:509:28)\n    at Server.emit (node:events:390:28)\n    at emitListeningNT (node:net:1368:10)\n    at processTicksAndRejections (node:internal/process/task_queues:82:21)"},"msg":"some error"}
```

## Pino configured with pretty-print:

```js
const log = pino({
	transport: {
		target: "pino-pretty",
		options: { destination: 1, colorize: true, ignore: "pid,hostname" },
	},
});
```

Running `npm run appWithPinoPretty` results in missing logs:

```bash
$ npm run appWithPinoCustom

> pino-problem@1.0.0 appWithPinoCustom
> node appWithPinoCustom.js
```

## Pino configured with custom transport:

```js
const log = pino({
	transport: {
		pipeline: [
			{
				// logger output for terminal, colorised by level for easy reading
				target: "./pinocustom.mjs",
				options: {
					destination: 1,
				},
			},
		],
	},
});
```

Running `npm run appWithPinoCustom` results in even more missing logs:

```bash
$ npm run appWithPinoCustom

> pino-problem@1.0.0 appWithPinoCustom
> node appWithPinoCustom.js

```

mcollina:
Fixes #1200

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -23,17 +23,10 @@ and can be used or configured via the options object passed to `pino` on initial
 A transport is a module that exports a default function which returns a writable stream:
 
 ```js
-import { Writable } from 'stream'
+import { createWriteStream } from 'fs'
+
 export default (options) => {
-  const myTransportStream = new Writable({
-    autoDestroy: true, // Needed for Node v12+
-    write (chunk, enc, cb) {
-    // apply a transform and send to stdout
-    console.log(chunk.toString().toUpperCase())
-    cb()
-    }
-  })
-  return myTransportStream
+  return createWriteStream(options.destination)
 }
 ```
 
@@ -55,7 +48,8 @@ The transport code will be executed in a separate worker thread. The main thread
 will write logs to the worker thread, which will write them to the stream returned
 from the function exported from the transport file/module.
 
-The exported function can also be async. Imagine the following transport:
+The exported function can also be async. If we use an async function we can throw early
+if the transform could not be opened. As an example:
 
 ```js
 import fs from 'fs'
@@ -107,6 +101,7 @@ For more details on `pino.transport` see the [API docs for `pino.transport`][pin
 [pino-transport]: /docs/api.md#pino-transport
 [sca]: https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm
 
+<a id="writing"></a>
 ### Writing a Transport
 
 The module [pino-abstract-transport](https://github.com/pinojs/pino-abstract-transport) provides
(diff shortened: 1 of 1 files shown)


docs/transports.md
```diff
@@ -116,11 +111,27 @@ You can see an example using a async iterator with ESM:
 
 ```js
 import build from 'pino-abstract-transport'
+import SonicBoom from 'sonic-boom'
+import { once } from 'events'
 
 export default async function (opts) {
+  // SonicBoom is necessary to avoid loops with the main thread.
+  // It is the same of pino.destination().
+  const destination = new SonicBoom({ dest: opts.destination || 1, sync: false })
```
jsumners, line 120: Why don't we show using `pino.destination` here instead?

mcollina: If I am writing a module to be published on npm I do not want to depend on pino. This should probably be written somewhere.

jsumners: That makes sense. I think these docs as written, though, are from the perspective of someone writing an _application_ that depends on Pino. Thus, they will already have the dependency.

mcollina: Should we mention both?

jsumners: Yeah, but we can follow up with another PR at a later date if you like.

mcollina: This page needs an overhaul I think :(. Let's land this and iterate.

jsumners (approved):
Looks good to me.

Issue #1164, opened 2021-10-17
Runtime issue (packed by esbuild): Cannot find module 'pino-pretty'


Unga:
Hi,

I packed project with esbuild (output is one file index.js).
Configuration of Pino transport:
`
transport: {
     target: 'pino-pretty',
     options: {
       colorize: true,
       ignore: 'time,pid,hostname',
    },
},
`

On runtime I receive next error:
node:internal/modules/cjs/loader:930
  throw err;
  ^
Error: Cannot find module 'pino-pretty'
Require stack:
 /Users/d/test/index.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:927:15)
    at Function.resolve (node:internal/modules/cjs/helpers:99:19)
    at fixTarget (/Users/d/test/index.js:36543:34)
    at transport (/Users/d/test/index.js:36534:26)
    at normalizeArgs (/Users/d/test/index.js:41495:20)
    at Object.pino [as default] (/Users/d/test/index.js:42270:32)
    ...

Code (row 36543) from my packed index.js:
`
function fixTarget(origin) {
 
    if (isAbsolute(origin) || origin.indexOf("file://") === 0) {
      return origin;
    }
    
    switch (origin) {
      case "pino/file":
        return join4(__dirname, "..", "file.js");
      default:
        return callerRequire.resolve(origin);   // row 36543
    }
}`

Any ideas what to do? Thanks

kibertoad:
Have you tried adding pino-pretty as a runtime dependency if you are usibg it?

Unga:
> Have you tried adding pino-pretty as a runtime dependency if you are using it?

Yep. Installed "npm i pino pino-pretty"

kibertoad:
wonder if bundler is not picking it up due to dynamic import. can you somehow force dependency bundling?

Unga:
> wonder if bundler is not picking it up due to dynamic import. can you somehow force dependency bundling?

Yesterday I switched our project from Winston to Pino because Winston has same issue... I decided to try something new :) But when I added pino-pretty to config I got same issue. 
What do you mean "force dependency bundling"?

PR #1209, opened 2021-11-05
feat: Support bundlers.

ShogunPanda:
This PR add supports for using pino in bundled source files.
In particular:

* It uses real-require to perform dynamic requires/import.
* It add support for a global variable __bundlerPathsOverrides which bundler plugins can set to override require paths if needed.

Note that in order for this to fully work, https://github.com/pinojs/thread-stream/pull/48 must be merged and released.

Fixes https://github.com/pinojs/pino/issues/1164
Fixes https://github.com/pinojs/pino/issues/1104

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,12 @@ In many cases, Pino is over 5x faster than alternatives.
 
 See the [Benchmarks](docs/benchmarks.md) document for comparisons.
 
+### Bundling support
+
+Pino supports to being bundled using tools like webpack or esbuild. 
+
+See [Bundling](docs/bundling.md) document for more informations.
+
 <a name="team"></a>
 ## The Team
 
--- a/docs/bundling.md
+++ b/docs/bundling.md
@@ -1,0 +1,30 @@
+# Bundling 
+
+Due to its internal architecture based on Worker Threads, it is not possible to bundle Pino *without* generating additional files.
+
+In particular, a bundler must ensure that the following files are also bundle separately:
+
+* `lib/worker.js` from the `thread-stream` dependency
+* `file.js`
+* `lib/worker.js`
+* `lib/worker-pipeline.js`
+* Any transport used by the user (like `pino-pretty`)
+
+Once the files above have been generated, the bundler must also add information about the files above by injecting a code which sets `__bundlerPathsOverrides` in the `globalThis` object.
+
+The variable is a object whose keys are identifier for the files and the the values are the paths of files relative to the currently bundle files.
+
+Example:
+
+```javascript
+// Inject this using your bundle plugin
+globalThis.__bundlerPathsOverrides = {
+  'thread-stream-worker': pinoWebpackAbsolutePath('./thread-stream-worker.js')
+  'pino/file': pinoWebpackAbsolutePath('./pino-file.js'),
+  'pino-worker': pinoWebpackAbsolutePath('./pino-worker.js'),
+  'pino-pipeline-worker': pinoWebpackAbsolutePath('./pino-pipeline-worker.js'),
+  'pino-pretty': pinoWebpackAbsolutePath('./pino-pretty.js'),
+};
+```
+
+Note that `pino/file`, `pino-worker`, `pino-pipeline-worker` and `thread-stream-worker` are required identifiers. Other identifiers are possible based on the user configuration.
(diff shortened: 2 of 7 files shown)


mcollina (changes requested):
This needs automated tests

jsumners:
Why are we patching around other tool's problems?

ShogunPanda:
We're not patching their problems.
We're being friendly and provide them a mechanism to integrate us without bad monkey-patching.

ShogunPanda:
Tests added.

mcollina (commented):
Could you add docs?

ShogunPanda:
Sure, will add them soon.

jsumners:
> We're not patching around their problems. We're being friendly and provide them a mechanism to integrate us without bad monkey-patching after embracing WorkerThreads caused serious issues.

I don't understand how checking for magic variables (that look to be intended as private) in the global object so that we can then use a patched up `require` function isn't patching around their issues. If that other tool works correctly, then it will work with all regular uses of `require`. None of our usage of `require` in Pino is abnormal.

mcollina:
There is a missing webpack plugin that will make this possible.

Currently pino and transport do not work _at all_ with bundled codebases.

ShogunPanda:
@jsumners Yeah, Matteo is right. Sorry, I should have mentioned that as we speak I'm working on one of the plugin that will add this support to webpack.

jsumners:
> Currently pino and transport do not work _at all_ with bundled codebases.

My question is: how is that the fault of Pino?

mcollina:
> My question is: how is that the fault of Pino?

It's not the fault of pino. This is not a bugfix but an additional feature. People _will_ ask for this and there are valid usecases for bundling their codebase.

jsumners:
> @jsumners Yeah, Matteo is right. Sorry, I should have mentioned that as we speak I'm working on one of the plugin that will add this support to webpack.

This is the slippery slope I'm trying to guard against. This PR addresses `webpack`. What about `parcel`, `esbuild`, `babel`, and whatever else someone has cooked up?  If they are not solved by this PR, then we end up having to add all sorts of hacks for every bundler that comes along.

ShogunPanda:
Nope, the fix is universal. I already experimented for webpack and esbuild.
We're gonna provide the webpack (as is one of the most used, if not the first one) plugin, so that people will have a reference.

mcollina:
The gist of the fix:

1. make sure the bundler generate separate entry points for the workers
2. make sure the pino knows where those files live

This is generic.

ShogunPanda:
@jsumners You can take a look at https://github.com/pinojs/pino-webpack-plugin/pull/1 to see how a bundler plugin might work.

jsumners:
> @jsumners You can take a look at [pinojs/pino-webpack-plugin#1](https://github.com/pinojs/pino-webpack-plugin/pull/1) to see how a bundler plugin might work.

Truthfully, no. I know nothing of webpack nor it's API.

I'm not going to block this work. I just disagree with it for the reasoning stated above.

ShogunPanda:
Docs added as well.

README.md
```diff
@@ -95,6 +95,12 @@ See the [Benchmarks](docs/benchmarks.md) document for comparisons.
 <a name="team"></a>
 ## The Team
 
+### Bundling support
+
+Pino supports to being bundled using tools like webpack or esbuild. 
+
+See [Bundling](docs/bundling.md) document for more informations.
```
jsumners, line 102: Why is this under the `## The Team` heading?

ShogunPanda: Because I misplaced it. :D 
Will fix soon, thanks for noticing!

ShogunPanda: Fixed!

mcollina (approved):
lgtm

simoneb:
@jsumners I'll chime in to provide some context around the underlying issue and why we thought it would be a good idea to address it.

First of all, this problem is introduced by pino@7. Why? Because using service workers requires physical files on the file systems. Hence, even though I can agree that bundling a Node application is not necessarily a great idea this change has broken **all** such usages of pino. And this is "pino's fault" if you wish. 

Obviously, all such usages could just as well decide to not upgrade to pino@7, but this breaking change is not really something that was even foreseen, it was just realized afterwards. Because bundling an application is something that in some cases is actually a valid scenario, we decided to try to see what supporting it would imply.

What it implies is that the original requirement of worker threads to have the file run by the worker exist physically on the file system remains, hence we needed to find a workaround make it happen when using a bundler, which turned out to be tricky but doable. It required the change contained in this PR for pino which can be used for all bundlers (or at least we hope so), and the rest will be done on the bundler's side.

For webpack, which is the only implementation we have so far, we created a plugin. https://github.com/pinojs/pino-webpack-plugin

For the other bundlers, which we are not actively working on, there will be similar approaches that we can follow. Hopefully with the webpack plugin we have paved the way for the other bundlers.

PR #1220, opened 2021-11-15
Avoid duplicate action runs for eg. Dependabot

voxpelli:
Dependabot, and some of you who are maintainers here, pushes to branches in the pino repo itself and unless one tell GitHub Action that it should jobs to only trigger the `push` events on the `main` branch, then it will trigger both the `push` and the `pull_request` events when PR:s gets opened for those, making a Dependabot PR look like:

Skärmavbild 2021-11-15 kl  23 02 06

I recently made this change myself to get rid of this duplicates, thought I would share 👍

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,6 +1,8 @@
 name: Benchmarks
 on:
   push:
+    branches:
+      - main
     paths-ignore:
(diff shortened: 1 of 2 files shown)


jsumners (approved):
26fc85fe50f1e9bb0f0252283e9a99ca

PR #1223, opened 2021-11-16
pass mkdir option on pino/file transport to destination

### LINKED ISSUE #1221: Add mkdir option to pino/file transport
Like SonicBoom, it would be helpful if the pino/file transport provided a `mkdir` option to instruct the transport whether to ensure the directory of the destination file exists before attempting to open it for writing.

Consider the following script:

```js
const pino = require('pino')
const ospath = require('path')
  target: 'pino/file',
  options: {
    destination: ospath.join(__dirname, 'logs/main.log'),
  },
})
logger.error('something bad happened')
```

Assuming this is run in a fresh directory that does not contain a `logs` folder, this code results in the following exception:

```
Error: ENOENT: no such file or directory, open '/tmp/pino-lab/logs/main.log'
Emitted 'error' event on ThreadStream instance at:
    at destroy (/tmp/pino-lab/node_modules/thread-stream/index.js:341:12)
```

The proposed `mkdir` option, if set, would ensure that the `/tmp/pino-lab/logs` directory exists before trying to open the destination file for writing.

```js
  target: 'pino/file',
  options: {
    destination: ospath.join(__dirname, 'logs/main.log'),
    mkdir: true,
  },
})
```

For more context and precedent, see https://github.com/pinojs/sonic-boom/issues/78.

mojavelinux:
If `mkdir` option is specified on the pino/file transport, pass it to the pino.destination.

closes #1221

### DIFF
--- a/file.js
+++ b/file.js
@@ -4,7 +4,9 @@ const pino = require('./pino')
 const { once } = require('events')
 
 module.exports = async function (opts = {}) {
-  const destination = pino.destination({ dest: opts.destination || 1, sync: false })
+  const destOpts = { dest: opts.destination || 1, sync: false }
+  if (opts.mkdir) destOpts.mkdir = true
+  const destination = pino.destination(destOpts)
   await once(destination, 'ready')
   return destination
 }
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -240,6 +240,17 @@ const transport = pino.transport({
 pino(transport)
 ```
 
+By default, the `pino/file` transport assumes the directory of the destination file exists. If it does not exist, the transport will throw an error when it attempts to open the file for writing. The `mkdir` option may be set to `true` to configure the transport to create the directory, if it does not exist, before opening the file for writing.
+
+```js
+const pino = require('pino')
+const transport = pino.transport({
+  target: 'pino/file',
+  options: { destination: '/path/to/file', mkdir: true }
+})
+pino(transport)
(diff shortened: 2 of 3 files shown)


mcollina (commented):
Could you add the documentation for this "new" option? Thanks!

mojavelinux:
Absolutely! Stay tuned.

file.js
```diff
@@ -4,7 +4,9 @@ const pino = require('./pino')
 const { once } = require('events')
 
 module.exports = async function (opts = {}) {
-  const destination = pino.destination({ dest: opts.destination || 1, sync: false })
+  const destOpts = { dest: opts.destination || 1, sync: false }
+  if (opts.mkdir) destOpts.mkdir = true
```
kibertoad, line 8: why not add ternary condition on destOpts instead?

mojavelinux: I'm happy to do it anyway that is acceptable to the project. Adding an undefined property broke an existing test. I wrote it in such a way that the existing test would pass.

mojavelinux:
I've added a paragraph with an example to the transports.md docs as requested. Let me know if there's anything you'd like me to change.

mcollina (approved):
lgtm

mcollina:
@mojavelinux could you send another PR with the `append` flag?

mojavelinux:
Yep! On it.

PR #1224, opened 2021-11-17
build(dependabot): ignore minor and patch github-actions updates

Fdawgs:
GitHub introduced the [ability to ignore specific updates](https://github.blog/changelog/2021-05-21-dependabot-version-updates-can-now-ignore-major-minor-patch-releases/) back in May.
This PR should stop Dependabot flooding PRs with every minor and patch update of GitHub's own actions every day.

Happy to make change for rest of Pino repos if wanted.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,6 +2,10 @@ version: 2
 updates:
 - package-ecosystem: github-actions
   directory: '/'
+  ignore:
+      - dependency-name: 'actions/*'
+        update-types:
+            ['version-update:semver-minor', 'version-update:semver-patch']
   schedule:
     interval: daily
   open-pull-requests-limit: 10
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -17,11 +17,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Code
-        uses: actions/checkout@v2.3.5
+        uses: actions/checkout@v2
         with:
           ref: ${{ github.base_ref }}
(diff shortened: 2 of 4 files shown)


mcollina (approved):
lgtm

mcollina:
amazing work!

mcollina:
Go ahead and roll it out throughout the org.

simoneb:
Just curious, apart from reducing the noise, why would you want to avoid getting any updates other than major?

jsumners:
> Just curious, apart from reducing the noise, why would you want to avoid getting any updates other than major?

We don't "avoid getting any updates". We avoid getting constant bumps to configuration files for explicit version numbers. This change says "I don't care what version of the v2 line of checkout you use, just use the latest v2".

Fdawgs:
> 
> 
> Just curious, apart from reducing the noise, why would you want to avoid getting any updates other than major?

GitHub's own actions follow [this tag style](https://docs.github.com/en/actions/creating-actions/about-custom-actions#using-tags-for-release-management) for releases, so specifying `actions/setup-node@v2` will get the latest v2.x.x of that action.

simoneb:
Ah interesting, I didn't know. From what I understand it isn't generally true that using `v2` means that you're getting whatever is the latest `v2`, it is true only as long as the publisher of the action moves the `v2` tag to refer to whatever is the latest. This is something that GitHub is doing with their own actions it seems, but I'm not sure how widespread a practice this is.

jsumners:
Very true. In my org, we use these `vX` tags for GitHub authored actions. For others, we pin to the full git SHA.

PR #1229, opened 2021-11-18
pass append option on pino/file transport to pino.destination function

### LINKED ISSUE #1222: Add append option to pino/file transport
Like SonicBoom, it would be helpful if the pino/file transport provided an `append` option to control the mode the transport uses when opening the file. The default would remain as "a" (append). If this option is set to `false`, the mode would be changed to "w" (truncate and write).

Consider the following script:

```js
const pino = require('pino')
const ospath = require('path')
  target: 'pino/file',
  options: {
    destination: ospath.join(__dirname, 'main.log'),
  },
})
logger.error('something bad happened')
```

Each time this script is run, a new log message is appended. The proposed `append` option, if set to `false`, would recreate the file anew.

```js
  target: 'pino/file',
  options: {
    destination: ospath.join(__dirname, 'main.log'),
    append: false,
  },
})
```

For more context and precedent, see https://github.com/pinojs/sonic-boom/issues/76.

mojavelinux:
If `append` option is specified on the pino/file transport, and the value is `false`, pass it to the pino.destination function.

closes #1222

### DIFF
--- a/file.js
+++ b/file.js
@@ -5,6 +5,7 @@ const { once } = require('events')
 
 module.exports = async function (opts = {}) {
   const destOpts = { dest: opts.destination || 1, sync: false }
+  if (opts.append === false) destOpts.append = false
   if (opts.mkdir) destOpts.mkdir = true
   const destination = pino.destination(destOpts)
   await once(destination, 'ready')
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -251,6 +251,17 @@ const transport = pino.transport({
 pino(transport)
 ```
 
+By default, the `pino/file` transport appends to the destination file if it exists. The `append` option may be set to `false` to configure the transport to truncate the file upon opening it for writing.
+
+```js
+const pino = require('pino')
+const transport = pino.transport({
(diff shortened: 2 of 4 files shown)


mcollina (approved):
lgtm

file.js
```diff
@@ -5,6 +5,7 @@ const { once } = require('events')
 
 module.exports = async function (opts = {}) {
   const destOpts = { dest: opts.destination || 1, sync: false }
+  if (opts.append === false) destOpts.append = false
   if (opts.mkdir) destOpts.mkdir = true
   const destination = pino.destination(destOpts)
```
jsumners, line 10: Would this not be more comprehensive?

```js
const destOpts = Object.assign({}, opts.destination, { dest: opts.destination || 1, sync: false })
```

mojavelinux: That's certainly a valid approach, but it would expand the scope of this issue. If you prefer to forward all the options to the destination, I won't object. @mcollina, where do you stand on this?

mcollina: go for it, it's better

PR #1233, opened 2021-11-20
Update LTS document

jsumners:
According to the doc itself, we only support actively supported Node.js LTS releases. Thus, v17 is a bonus covered by our CI but not a guarantee in our policy.


mcollina (approved):
lgtm

mcollina:
It's better to keep the current release in CI to avoid suprises once it goes lts

PR #1234, opened 2021-11-20
forward all unknown options passed to pino/file transport to destination

mojavelinux:
As discussed in #1229.

We delete the `destination` property just to avoid any confusion since that's not a properly that pino.destination / SonicBoom recognizes.

### DIFF
--- a/file.js
+++ b/file.js
@@ -4,9 +4,8 @@ const pino = require('./pino')
 const { once } = require('events')
 
 module.exports = async function (opts = {}) {
-  const destOpts = { dest: opts.destination || 1, sync: false }
-  if (opts.append === false) destOpts.append = false
-  if (opts.mkdir) destOpts.mkdir = true
+  const destOpts = Object.assign({}, opts, { dest: opts.destination || 1, sync: false })
+  delete destOpts.destination
   const destination = pino.destination(destOpts)
   await once(destination, 'ready')
   return destination


mcollina (approved):
lgtm

file.js
```diff
@@ -4,9 +4,8 @@ const pino = require('./pino')
 const { once } = require('events')
 
 module.exports = async function (opts = {}) {
-  const destOpts = { dest: opts.destination || 1, sync: false }
-  if (opts.append === false) destOpts.append = false
-  if (opts.mkdir) destOpts.mkdir = true
+  const destOpts = Object.assign({}, opts, { dest: opts.destination || 1, sync: false })
+  delete destOpts.destination
   const destination = pino.destination(destOpts)
```
jsumners, line 9: I don't think this is quite sufficient. Notice that my original suggestion updated the `opts.destination` object, not the `opts` object itself. But, even so, that wouldn't have worked as intended because:

https://github.com/pinojs/pino/blob/0433f54f5ea292b6c01b9d62750986935742150d/pino.js#L192-L199

The `destination` function only accepts a single parameter: the target file descriptor. It then passes that along to `sonic-boom`:

https://github.com/pinojs/pino/blob/0433f54f5ea292b6c01b9d62750986935742150d/lib/tools.js#L335-L342

I believe `pino.destination` needs to be updated to recognize an object has been supplied as the parameter and pass it along appropriately.

PR #1235, opened 2021-11-21
consistently use the term mergingObject in the docs

mojavelinux:
While reading the docs, I found one section that slipped into using mergedObject instead of mergingObject. This PR fixes this inconsistency.


mcollina (approved):
lgtm

PR #1238, opened 2021-11-23
test: make test timeouts configurable

mhdawson:
We saw some timeouts in the tests when we were running to validate in Red Hat containers (for example: https://catalog.redhat.com/software/containers/ubi8/nodejs-12/5d3fff015a13461f5fb8635a).

Our investigation so far points to it being loading/resource issue on the machines we are running versus anything in the tests.

We'd like to be able to extended timeouts in the tests, this PR adds a simple way to do that through the environment.

### DIFF
--- a/test/helper.js
+++ b/test/helper.js
@@ -54,7 +54,7 @@ function sleep (ms) {
 
 function watchFileCreated (filename) {
   return new Promise((resolve, reject) => {
-    const TIMEOUT = 2000
+    const TIMEOUT = process.env.PINO_TEST_WAIT_WATCHFILE_TIMEOUT || 2000
     const INTERVAL = 100
     const threshold = TIMEOUT / INTERVAL
     let counter = 0
@@ -79,7 +79,7 @@ function watchFileCreated (filename) {
 
 function watchForWrite (filename, testString) {
   return new Promise((resolve, reject) => {
-    const TIMEOUT = 2000
+    const TIMEOUT = process.env.PINO_TEST_WAIT_WRITE_TIMEOUT || 2000
     const INTERVAL = 100
     const threshold = TIMEOUT / INTERVAL
     let counter = 0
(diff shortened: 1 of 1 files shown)


kibertoad:
Shouldn't this be documented somewhere?

test/helper.js
```diff
@@ -79,7 +79,7 @@ function watchFileCreated (filename) {
 
 function watchForWrite (filename, testString) {
   return new Promise((resolve, reject) => {
-    const TIMEOUT = 2000
+    const TIMEOUT = process.env.PINO_TEST_WAIT_WRITE_TIMEOUT || 2000
```
mcollina, line 82:
```suggestion
    const TIMEOUT = process.env.PINO_TEST_WAIT_WRITE_TIMEOUT || 10000
```

test/helper.js
```diff
@@ -54,7 +54,7 @@ function sleep (ms) {
 
 function watchFileCreated (filename) {
   return new Promise((resolve, reject) => {
-    const TIMEOUT = 2000
+    const TIMEOUT = process.env.PINO_TEST_WAIT_WATCHFILE_TIMEOUT || 2000
```
mcollina, line 57:
```suggestion
    const TIMEOUT = process.env.PINO_TEST_WAIT_WATCHFILE_TIMEOUT || 10000
```

mcollina (commented):
I'd recommend we increase the default to 10s.

jsumners (approved):
Looks good to me with suggestions applied. Will probably solve some flakiness in our own CI.

mcollina:
I'm not sure we should document this.. but if we want to 
the place is https://github.com/pinojs/pino/blob/master/CONTRIBUTING.md.

kibertoad:
`I'm not sure we should document this`

How would people know it's there without randomly stumbling upon it in code then?

mcollina:
A timeout error would point to that block of code.

mhdawson:
@mcollina thanks, accepted your suggestions.

From the output I'll confirm it was pretty easy to see that the failures were a timeout issue and where the timeouts where.

mcollina (approved):
lgtm

Issue #1193, opened 2021-10-27
New bundled TypeScript typings do not expose various top-level exports


thw0rted:
As I commented in #910, the bundled TypeScript type definitions do not allow `import {destination, stdSerializers} from "pino"`, which is perfectly legal code.

Looking at `pino.js` (as shipped in v7.0.5), I see top level exports defined for the following variables:

````text
destination
transport
multistream
final
levels
stdSerializers
stdTimeFunctions
symbols
version
````

Since these can be imported from the module, the typings should include definitions for them, to accurately describe the shape of the JS module.

While you're at it, consider removing the namespace "P" that wraps all your type and interface declarations, and exporting them directly.  Namespace bundling is no longer considered a best practice in TS, and prevents the use of the (relatively recently introduced) `import type` syntax.

This has real-world consequences.  If I want to write a library that takes a `DestinationStream`, writing `import pino from "pino"; export function withStream(str: pino.DestinationStream) { ... }` results in including the entire `pino` library at runtime, even though I'm not actually constructing a logger or using any of pino's runtime logic, while the exact same code with `import type {DestinationStream} from "pino"` instead, only requires the library at compile-time, and doesn't bundle the pino implementation.

KoltesDigital:
`import type {Logger} from "pino";` worked in 6.x but not in 7.x anymore. Bringing it back should be a _patch_ ... and of course I'd like to have it released as soon as possible so that I don't have to temporarily switch to P 🤗 . I guess many users of this library are in this case. I also think you can leave P with a deprecated mention, I guess few users switched.

FTR I found this thread because I'm upgrading a project's dependencies and tsc complains about non-existing Logger type, whereas it worked before.

Retro64:
I stumbled over the same lost exported top-level Logger type when upgrading to the new version. I agree it should be a patch, as adding another top-level export does not mean the export within the namespace P needs to be dropped. It - for sure - becomes redundant, but won't be breaking (as it is for now for people, blindly upgrading the dependency to 7.x).

kibertoad:
Released in 7.5.0

PR #1246, opened 2021-11-30
fix: type regression on fastify

climba03003:
Fixes the type regression on https://github.com/fastify/fastify/pull/3506

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -18,7 +18,7 @@
 
 import type { EventEmitter } from "events";
 // @ts-ignore -- gracefully falls back to `any` if not installed
-import type { PrettyOptions } from "pino-pretty";
+import type { PrettyOptions as PinoPrettyOptions } from "pino-pretty";
 import type { SonicBoom, SonicBoomOpts } from "sonic-boom";
 import type { WorkerOptions } from "worker_threads";
 
@@ -26,9 +26,6 @@ import * as pinoStdSerializers from "pino-std-serializers";
 
 
 //// Non-exported types and interfaces
-
-type SerializerFn = (value: any) => any;
-type WriteFn = (o: object) => void;
 
 // ToDo https://github.com/pinojs/thread-stream/issues/24
 type ThreadStream = any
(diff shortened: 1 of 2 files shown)


pino.d.ts
```diff
@@ -192,10 +105,98 @@ interface LoggerExtras extends EventEmitter {
 
 declare namespace pino {
     //// Exported types and interfaces
+    
+    interface BaseLogger {
```
kibertoad, line 109: why move it from top level into a namespace?

climba03003: Because it never export outside.

kibertoad: can't we export it instead?

climba03003: Then the old way of import is missing this exportation.
e.g.
```
import pino from 'pino'
pino.BaseLogger
```

climba03003:
I have tested even without my changes, it failed the test.

I have no idea why it failed by `@types/jsdom` type check and it is something that we never import.

mcollina (approved):
lgtm

pino.d.ts
kibertoad: We generally want to move away from relying on namespace. See https://github.com/pinojs/pino/issues/1193#issuecomment-966899139
I agree that types should be fixed if we lost the way to do something we could do before, though.

climba03003: I agree that it should move away from namespace, but the current version is too late to do it.
Current version should support the import syntax that can be used in `7.0.0`.

The complete removal should be happen in next major `8.0.0`.

mcollina:
@climba03003 can you rebase? 
@kibertoad fixed this in https://github.com/pinojs/pino/pull/1251.

mcollina:
@kibertoad are you good with this landing?

climba03003:
CI fail is not related to TypeScript changes.

mcollina:
This can land as soon as @kibertoad approves. https://github.com/pinojs/pino/pull/1253 is to improve that test.

pino.d.ts
kibertoad: I understand that we can't remove it yet, but can we at least already provide alternative for people who want to do direct named imports? That was the gist of #1193 and it seems like the right direction.
Other than that, PR LGTM.

climba03003: As I mention on line 768, it has the top-level export mapped.

pino.d.ts
```diff
@@ -756,10 +759,13 @@ export type LevelChangeEventListener = pino.LevelChangeEventListener;
 export type LogDescriptor = pino.LogDescriptor;
 export type Logger = pino.Logger;
 export type SerializedError = pino.SerializedError;
+export type SerializerFn = pino.SerializerFn;
 export type SerializedRequest = pino.SerializedRequest;
 export type SerializedResponse = pino.SerializedResponse;
+export type WriteFn = pino.WriteFn;
 
 // Interfaces
+export interface BaseLogger extends pino.BaseLogger {}
```
climba03003, line 768: @kibertoad I think here is what you mention?

pino.d.ts
climba03003: This PR can import the type I moved inside `namespace` with the below format.

```ts
// named import
import { BaseLogger } from 'pino'
// namespace import / default import
import pino from 'pino'
pino.BaseLogger
// pino named import
import { pino } from 'pino'
pino.BaseLogger
// old P named import
import { P } from 'pino'
P.BaseLogger
```

If anything is missing, please let me know.

kibertoad (approved):

pino.d.ts
kibertoad: Thank you for clarification!

Issue #1106, opened 2021-08-28
Deprecate prettyPrint: true


mcollina:
We had too many folks complaining about how we load pino-pretty with yarn 2/3 and similar. I think we should either

1) require the user to pass in a reference to pino-pretty
2) make pino-pretty a full transport and let folks use it from there.

jsumners:
I'd prefer to fully deprecate it from Pino. Let people pass in a transport or whatever that does it. I'd also like to keep `pino-pretty` narrowed to its current focus: parsing `stdin`. If necessary, and possible, I'd rather see us create `pino-pretty-transport` to provide the transport.

leorossi:
Hi guys, I am willing to work on this.

Just to be clear, the intention of this is 
- remove `prettyPrint` option from constructor
- remove all tests that relies on this feature (i.e: checking the output has been "prettified" and things like this)
- remove all documentation related to that option, and redirect to `pino-pretty` 

is that correct?

Do we want to show a warning when passing `{ prettyPrint: true }` in the constructor?

mcollina:
I have some code prepared for this, wait until I upload it (might take a few days).

zhex900:
Hi,

I am using streams. What should I do to remove the deprecation message?

```

const loggerStream = () => {
  const logPath = path.resolve(
    process.platform === "win32"
      ? `${WIN_LOG_PATH}/${LOG_FILE}`
      : `${getAppDataPath()}/${LOG_FILE}`
  );
  return multistream(
    [
      {
        level: "error",
        stream: createStream(logPath, LOG_ROTATION_SETTINGS),
      },
      {
        level: "debug",
        stream: createStream(logPath, LOG_ROTATION_SETTINGS),
      },
      {
        stream: createStream(logPath, LOG_ROTATION_SETTINGS),
      },
    ],
    { dedupe: true }
  );
};

const logger = pino(
  {
    prettyPrint: {
      colorize: true,
      errorLikeObjectKeys: ["err", "error", "details"],
      levelFirst: true,
      translateTime: "SYS:hh:MM:ss TT dd-mm-yyyy",
      ignore: "pid,hostname,service",
      messageFormat: (log, messageKey) => {
        return applyServiceStyles(log, messageKey);
      },
    },
    level: process.env.LOG_LEVEL || "debug",
    serializers: {
      error: pino.stdSerializers.err,
    },
  },
  loggerStream()
);
````

PR #1249, opened 2021-11-30
Remove prettyPrint option.

mcollina:
100% code coverage back.

This is the start of a v8 planning. We might want to do this sooner than expected and roll it out with Fastify v4.


jsumners:
🤣 https://github.com/pinojs/pino/issues/1106#issuecomment-907614629

mcollina:
@kibertoad could you take a look why TS is failing badly?

davidmarkclements:
should we create a v8 branch and merge to that?

lib/deprecations.js
```diff
@@ -3,8 +3,6 @@
 const warning = require('fastify-warning')()
 module.exports = warning
 
-const warnName = 'PinoWarning'
+// const warnName = 'PinoWarning'
 
-warning.create(warnName, 'PINODEP008', 'prettyPrint is deprecated, look at https://github.com/pinojs/pino-pretty for alternatives.')
-
-warning.create(warnName, 'PINODEP009', 'The use of pino.final is discouraged in Node.js v14+ and not required. It will be removed in the next major version')
+// warning.create(warnName, 'PINODEP010', 'A new deprecation')
```
davidmarkclements, line 8: stray comment?

lib/deprecations.js
```diff
@@ -3,8 +3,6 @@
 const warning = require('fastify-warning')()
 module.exports = warning
 
-const warnName = 'PinoWarning'
```
davidmarkclements, line 6: is this file required anywhere else now? if not, remove?

mcollina:
> should we create a v8 branch and merge to that?

It should be a `next` branch. I'll get that started asap.

lib/tools.js
```diff
@@ -386,6 +256,8 @@ function autoEnd (stream, eventName) {
       stream.end()
     })
   } else {
+    // For some reason istanbul is not detecting this, but it's there
+    /* istanbul ignore next */
```
davidmarkclements, line 260: might be solvable with brittle+c8

lib/deprecations.js
mcollina: This is on purpose to avoid deleting this file.

lib/deprecations.js
mcollina: We'll definitely need it at some point in the future, better to keep.

lib/tools.js
```diff
@@ -510,16 +316,6 @@ function buildFormatters (level, bindings, log) {
   }
 }
 
-function setMetadataProps (dest, that) {
```
davidmarkclements, line 513: love to see this go

test/basic.test.js
```diff
@@ -730,3 +730,11 @@ test('nestedKey should not be used for non-objects', async ({ strictSame }) => {
     msg: message
   })
 })
+
+test('throws if prettyPrint is passed in as an option', async (t) => {
+  t.throws(() => {
+    pino({
+      prettyPrint: true
+    })
+  }, new Error('prettyPrint option is no longer supported, see the pino-pretty package.'))
```
davidmarkclements, line 739: include a link?

test/multistream.test.js
```diff
@@ -503,21 +503,17 @@ test('flushSync', function (t) {
     '_' + Math.random().toString(36).substr(2, 9)
   )
   const destination = pino.destination({ dest: tmp, sync: false, minLength: 4096 })
-  const log = pino({ level: 'info' }, multistream([{ level: 'info', stream: destination }]))
+  const stream = multistream([{ level: 'info', stream: destination }])
```
davidmarkclements, line 506: pino v8 side note: can we merge multistream and transport APIs in a way where the multistream same-thread logic is used based on available threads (or a force option)

lib/deprecations.js
davidmarkclements: is the file needed? if not we should remove it - if we ever need it again it's in the commit history right?

lib/deprecations.js
davidmarkclements: for me it's a loose end - maybe this whole concept should be a separate module or something

lib/deprecations.js
jsumners: It's easier to keep it and update it as necessary.

kibertoad:
> @kibertoad could you take a look why TS is failing badly?

I'm on it.

kibertoad:
@mcollina Please don't laugh, but I think this one was actually a Jest issue. Fixes in a non-elegant way for now because Jest requires it, but shouldn't matter because we only use tsconfig for local testing anyway. Not sure why it only triggered now.

They promise to do better in v28: https://github.com/facebook/jest/issues/12098

test/multistream.test.js
kibertoad: I suggest to park this till v9, it's better to ship what we have and improve in the next iteration.

kibertoad:
@mcollina Is there anything else that you want to see in this PR, or we can merge after CI is green?

mcollina:
Can you add Node v18 and drop v17?

mcollina:
@jsumners you good with pino@8 being a tiny update just to drop node v12?

jsumners:
> @jsumners you good with pino@8 being a tiny update just to drop node v12?

Yes.

jsumners (approved):

PR #1254, opened 2021-12-07
compatibility fix

### LINKED ISSUE #1237: Pino with transports not compatible with pkg
I have included all the worker scripts in scripts list in pkg.
But I still can't run the app when packaged due to the incompatibility of the real-require which is part of the pinojs modules

This is the error I get:

TypeError: Invalid host defined options
    at eval (eval at <anonymous> (C:\snapshot\test\node_modules\real-require\src\index.js:4:20), <anonymous>:3:1)
    at start (C:\snapshot\test\node_modules\thread-stream\lib\worker.js:21:17)
    at Object.<anonymous> (C:\snapshot\test\node_modules\thread-stream\lib\worker.js:67:1)

rokku-x:
This will fix some incompatibility issues with some node.js use-cases by handling over the module loading to require. This patch is dependent also to the patches I made with the thread-stream and real-require.

resolves #1237

### DIFF
--- a/lib/worker.js
+++ b/lib/worker.js
@@ -13,6 +13,12 @@ module.exports = async function ({ targets }) {
   targets = await Promise.all(targets.map(async (t) => {
     let fn
     try {
+      if(typeof process !== "undefined"){
+        //reuse the error code.
+        const err = new Error()
+        err.code = 'ERR_MODULE_NOT_FOUND';
+        throw err;
+      }
       const toLoad = 'file://' + t.target
       fn = (await realImport(toLoad)).default
     } catch (error) {


lib/worker.js
```diff
@@ -13,6 +13,12 @@ module.exports = async function ({ targets }) {
   targets = await Promise.all(targets.map(async (t) => {
     let fn
     try {
+      if(typeof process !== "undefined"){
```
jsumners, line 16: Please follow the linting rules of this project.

lib/worker.js
```diff
@@ -13,6 +13,12 @@ module.exports = async function ({ targets }) {
...
+        //reuse the error code.
```
jsumners, line 17: What is this comment referencing?

lib/worker.js
```diff
@@ -13,6 +13,12 @@ module.exports = async function ({ targets }) {
...
+        const err = new Error()
```
jsumners, line 18:
```suggestion
        const err = Error()
```

Also, an appropriate message should be passed to the constructor.

lib/worker.js
```diff
@@ -13,6 +13,12 @@ module.exports = async function ({ targets }) {
...
+        err.code = 'ERR_MODULE_NOT_FOUND';
```
jsumners, line 19: Is this code appropriate for the condition?

jsumners (changes requested):
Please provide unit test to cover new and changed code.

lib/worker.js
rokku-x: I am sorry. I meant that I reused the error code for the condition to match in the if statement catch clause.

lib/worker.js
mcollina: We should not emit Node core error codes

lib/worker.js
mcollina: How can this condition happen?

mcollina (changes requested):
Tests are needed

rokku-x:
I was long away from from the project where I used this. I remembered it as someone proposed a better fix. I tried it and it was working. I should close this now.

PR #1276, opened 2021-12-20
fix(caller): use calling file name for libraries calling Pino

FredericEspiau:
Excluding all the files in `node_modules` didn't work with `nestjs-pino`

I guess because there are too many files before reaching the user's files

This will return the file in second position in the stacktrace as `get-caller-file` did if we can't find a file outside `node_modules`

Stacktrace:
````
/Users/my-repo/node_modules/one-dependency/packages/file1.js
/Users/my-repo/node_modules/one-dependency/packages/file2.hs
/Users/my-repo/node_modules/pino-http/logger.js
/Users/my-repo/node_modules/pino-http/logger.js
/Users/my-repo/node_modules/nestjs-pino/LoggerModule.js
/Users/my-repo/node_modules/nestjs-pino/LoggerModule.js
/Users/my-repo/node_modules/@nestjs/core/middleware/middleware-module.js
/Users/my-repo/node_modules/@nestjs/core/middleware/middleware-module.js
````

Other example:
````
/Users/my-repo/node_modules/nestjs-pino/PinoLogger.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
/Users/my-repo/node_modules/@nestjs/core/injector/injector.js
````

I wasn't able to create a test for this case

This fixes https://github.com/pinojs/pino/issues/1277

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,3 +55,5 @@ profile-*
 # lock files
 yarn.lock
 package-lock.json
+
+!test/fixtures/eval/node_modules
--- a/lib/caller.js
+++ b/lib/caller.js
@@ -1,7 +1,17 @@
 'use strict'
 
+const fs = require('fs')
+
 function noOpPrepareStackTrace (_, stack) {
   return stack
+}
+
+function isFileName (input) {
+  return typeof input === 'string' && fs.existsSync(input)
+}
+
+function isOutsideNodeModules (file) {
+  return isFileName(file) && file.indexOf('node_modules') === -1
 }
 
 module.exports = function getCaller () {
@@ -14,13 +24,15 @@ module.exports = function getCaller () {
     return undefined
   }
 
-  for (const entry of stack.slice(2)) {
+  const entries = stack.slice(2)
+
+  for (const entry of entries) {
     const file = entry ? entry.getFileName() : undefined
 
-    if (file && file.indexOf('node_modules') === -1) {
+    if (isOutsideNodeModules(file)) {
       return file
     }
   }
 
-  return stack[2]?.getFileName()
+  return entries[0] ? entries[0].getFileName() : undefined
 }
--- a/lib/caller.js
+++ b/lib/caller.js
@@ -21,4 +21,6 @@ module.exports = function getCaller () {
       return file
     }
   }
+
+  return stack[2]?.getFileName()
 }
--- a/lib/caller.js
+++ b/lib/caller.js
@@ -6,12 +6,8 @@ function noOpPrepareStackTrace (_, stack) {
   return stack
 }
 
-function isFileName (input) {
-  return typeof input === 'string' && fs.existsSync(input)
-}
-
-function isOutsideNodeModules (file) {
-  return isFileName(file) && file.indexOf('node_modules') === -1
+function isOutsideNodeModules (input) {
+  return typeof input === 'string' && fs.existsSync(input) && input.indexOf('node_modules') === -1
 }
 
 module.exports = function getCaller () {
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

**Review by mcollina — COMMENTED** [review]
Could you add a test?

**Comment by FredericEspiau** [comment]
I will try, I just had an idea on how to do it

**Comment by mkaufmaner** [comment]
Oh this seems to be related; https://github.com/pinojs/pino/issues/1278

**Comment by FredericEspiau** [comment]
Yup its the same issue but I won't update my PR before tomorrow (2021-12-21T08:00:00+01:00)

File: lib/caller.js [hunk 1]
```diff
@@ -21,4 +21,6 @@ module.exports = function getCaller () {
       return file
     }
   }
+
+  return stack[2]?.getFileName()
```
mcollina, line 25: Don't use the `?` as it's not in node v12.

jsumners: Also, a comment explaining what this is would be very helpful.

FredericEspiau:
done, this also handles https://github.com/pinojs/pino/issues/1278

mcollina (approved):
lgtm

mcollina:
Thanks!

mcollina:
@ShogunPanda could you take a look?

lib/caller.js
```diff
@@ -1,9 +1,19 @@
...
+
 function noOpPrepareStackTrace (_, stack) {
   return stack
 }
 
+function isFileName (input) {
+  return typeof input === 'string' && fs.existsSync(input)
+}
+
+function isOutsideNodeModules (file) {
```
ShogunPanda, line 13: Little nit: I would join `isFileName` and `isOutsideNodeModules` as follow to avoid extra context switch:

```js
function isOutsideNodeModules (file) {
  return typeof input === 'string' && fs.existsSync(input) && file.indexOf('node_modules') === -1
}}
```

ShogunPanda (approved):
Single nit, but in general LGTM!

mcollina:
CI is failing.

FredericEspiau:
I can't reproduce the issue on my computer, I don't know what I can do to this the CI failing 🤔

ShogunPanda:
@FredericEspiau I found the problem: you have several files (like `fixtures/eval/node_modules/2-files.js`) which are not included in the commit since `node_modules` is in `.gitignore`.

You have to explicitly add to the PR or, even better, provide them as an archive which you can unarchive during test time in a temporary folder.

FredericEspiau:
Excellent finding, indeed while doing my dev I kept on thinking that I should remove the this folder from the gitignore and then I forgot! Thanks I'll do it asap

lib/caller.js
FredericEspiau: done

.gitignore
```diff
@@ -55,3 +55,5 @@ profile-*
 # lock files
 yarn.lock
 package-lock.json
+
+!test/fixtures/eval/node_modules
```
jsumners, line 59: This is the correct solution to getting the fixture into the repo. Adding a binary archive would not be acceptable in my opinion.

FredericEspiau:
Awesome, the CI passed !

PR #1279, opened 2021-12-21
feat: dynamic stream destination

### LINKED ISSUE #1271: [question] How to create custom destination based on log level
Hi, I would like to log errors to stderr and everything else to stdout. I know I could archieve this in multiple ways (using transports or multi stream) BTW that would cause logs to be duplicated, I would like to customize the destination of pino pretty and redirect logs based on the level. Could anyone suggest me how? I think this could also be a useful option to add, something like `errorsDestination` that by default is set to `destination`

robertsLando:
Fixes #1271

### DIFF
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -50,7 +50,7 @@ function multistream (streamsArray, opts) {
     for (let i = 0; i < streams.length; i++) {
       dest = streams[i]
       if (dest.level <= level) {
-        stream = dest.stream
+        stream = typeof dest.stream === 'function' ? dest.stream(level) : dest.stream
         if (stream[metadata]) {
           const { lastTime, lastMsg, lastObj, lastLogger } = this
           stream.lastLevel = level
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
I don't understand this. The following pass the same assertions:

```js
test('dynamic stream destination', function (t) {
  let messageCount = 0
  const stream1 = writeStream(function (data, enc, cb) {
    messageCount += 1
    cb()
  })

  const stream2 = writeStream(function (data, enc, cb) {
    messageCount -= 1
    cb()
  })

  const streams = [
    {
      stream: stream1,
      level: 'info'
    },
    {
      stream: stream2,
      level: 'error'
    }
  ]

  const log = pino({
    level: 'trace'
  }, multistream(streams, { dedupe : true }))
  log.info('info stream')
  log.error('error stream')
  t.equal(messageCount, 0)
  t.end()
})
```

robertsLando:
@mcollina Ok by looking at it again I misunderstood that... I think I need to patch pino pretty instad of pino

I actually have a pretty stream transport and a file stream transport (that is only working in production). I would like the pretty stream to use the stdout for all levels except errors, I cannot use dedupe as this would prevent logs to be logged on fileStream in production

robertsLando:
@mcollina Closing this for now as it doesn't add anything new, if you have any tips about the comment above let me know. Thanks

mcollina:
> I actually have a pretty stream transport and a file stream transport (that is only working in production). I would like the pretty stream to use the stdout for all levels except errors, I cannot use dedupe as this would prevent logs to be logged on fileStream in production

Just remove dedupe.

PR #1284, opened 2021-12-28
feat: support for `maxLevel` in multistream

### LINKED ISSUE #1271: [question] How to create custom destination based on log level

robertsLando:
Fixes #1271

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1101,7 +1101,7 @@ finalLogger.info('exiting...')
 
 <a id="pino-multistream"></a>
 
-### `pino.multistream(options) => Stream`
+### `pino.multistream(streams, options) => Stream`
 
 Create a stream composed by multiple destination streams:
 
@@ -1111,7 +1111,8 @@ var pino = require('pino')
 var pretty = require('pino-pretty')
 var streams = [
   {stream: fs.createWriteStream('/tmp/info.stream.out')},
-  {stream: pretty() },
+  {stream: pretty(), dest: 1, maxLevel: 'error' }, // pretty print logs on stdout (excluded errors)
+  {stream: pretty(), dest: 2, level: 'error' }, // pretty print errors on stderr
   {level: 'debug', stream: fs.createWriteStream('/tmp/debug.stream.out')},
   {level: 'fatal', stream: fs.createWriteStream('/tmp/fatal.stream.out')}
 ]
@@ -1126,7 +1127,17 @@ log.info('this will be written to /tmp/debug.stream.out and /tmp/info.stream.out
 log.fatal('this will be written to /tmp/debug.stream.out, /tmp/info.stream.out and /tmp/fatal.stream.out')
 ```
 
-In order for `multistream` to work, the log level __must__ be set to the lowest level used in the streams array.
+In order for `multistream` to work, the log `level` __must__ be set to the lowest `level` used in the streams array.
+
+#### Streams
+
+Stream or array of streams. Each stream may be a `stream` (`DestinationStream`) or `object` (`StreamEntry`) with the following properties:
+
++ `stream`: A Node.js stream (Default to `process.stdout`)
++ `level`: The log level of the stream (Default to `'info'`)
++ `maxLevel`: The max log level of the stream (Optional)
+
+`level` and `maxLevel` are used to filter the logs using the following rule: `stream.level <= log.level < stream.maxLevel`. If `maxLevel` is not specified it will not be used.
 
 #### Options
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -1111,8 +1111,8 @@ var pino = require('pino')
 var pretty = require('pino-pretty')
 var streams = [
   {stream: fs.createWriteStream('/tmp/info.stream.out')},
-  {stream: pretty(), dest: 1, maxLevel: 'error' }, // pretty print logs on stdout (excluded errors)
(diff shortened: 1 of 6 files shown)


jsumners (changes requested):
Documentation is missing. Also, I'm wondering if `allowedLevels: []` is a better API.

pino.d.ts
```diff
@@ -249,6 +249,7 @@ declare namespace pino {
     interface StreamEntry {
         stream: DestinationStream
         level?: Level
+        maxLevel?: Level
```
mcollina, line 252: Can you add a test for `maxLevel` property.

robertsLando: Done

robertsLando:
@jsumners Documentation added. Would like to know @mcollina opinion about `allowedLevels`, IMO it's not necessary

mcollina (approved):
lgtm

I don't think allowedLevels would work as with custom levels these arrays could get really big and complicate the logic.

I would recommend we rename level to minLevel (aliasing it) to clarify it further.

robertsLando:
> I would recommend we rename level to minLevel (aliasing it) to clarify it further.

Do you mean in StreamEntry or in internal stream array? If the first one do you plan to keep a back compatibility or a breaking change?

mcollina:
I'm not sure what StreamEntry is. I mean the same object where you added maxLevel. We would need to maintain backward compatibility.

robertsLando:
@mcollina StreamEntry it's the type of the streams passed as object in pino: https://github.com/pinojs/pino/blob/master/pino.d.ts#L249

The type could be that one or DestinationStream

BTW I can do that now

mcollina:
I'd prefer to have it in this PR.

docs/api.md
```diff
@@ -1127,17 +1127,17 @@ log.info('this will be written to /tmp/debug.stream.out and /tmp/info.stream.out
...
 
 Stream or array of streams. Each stream may be a `stream` (`DestinationStream`) or `object` (`StreamEntry`) with the following properties:
 
 + `stream`: A Node.js stream (Default to `process.stdout`)
-+ `level`: The log level of the stream (Default to `'info'`)
++ `minLevel`: The log level of the stream (Default to `'info'`)
 + `maxLevel`: The max log level of the stream (Optional)
 
-`level` and `maxLevel` are used to filter the logs using the following rule: `stream.level <= log.level < stream.maxLevel`. If `maxLevel` is not specified it will not be used.
+`minLevel` and `maxLevel` are used to filter the logs using the following rule: `stream.minLevel <= log.level < stream.maxLevel`. If `maxLevel` is not specified it will not be used.
```
jsumners, line 1140: This is what confuses me about `maxLevel`. Setting `minLevel` is inclusive, but `maxLevel` is exclusive. This is inconsistent and confusing. But if `maxLevel` is _inclusive_ then the feature does not accomplish the goal of allowing only a _single_ level to reach the stream. This is why I was thinking an allow list would be better (while recognizing that it has a few problems).

In order to get consistent behavior, I think the stream should accept `onlyLevel: 'info'` as an option.

I am desiring this feature myself, but I fear we are getting deep into "hack an API" land. What are your thoughts @mcollina @davidmarkclements ?

robertsLando: @jsumners How could I archieve the same result with `onlyLevel`? Check my original problem:

> > I actually have a pretty stream transport and a file stream transport (that is only working in production). I would like the pretty stream to use the stdout for all levels except errors, I cannot use dedupe as this would prevent logs to be logged on fileStream in production. If I remove dedupe I would have the logs duplicated on error and stdout, If i add dedupe, in production, I would miss the logs in file. I just would like to have an option in pretty stream to say log error levels to stderr and others to stdout, or create a custom stream to handle this

If I set onlyLevel to error stream I would have the same problem I have now that the logs are duplicated, and if I use the dup flag I would miss the logs in the file stream

I think this is the best solution to have more control of the stream levels

jsumners: > If I set onlyLevel to error stream I would have the same problem I have now that the logs are duplicated, and if I use the dup flag I would miss the logs in the file stream

I don't understand. This should allow for sending desired levels to specific and all things to others:

```js
const log = pino({ level: 'info' }, pino.multistream([
  { onlyLevel: 'error', stream: errorStream },
  { onlyLevel: 'debug', stream: debugStream },
  { level: 'info', stream: pino.destination() } // normal output stream
]))
```

robertsLando: Ok so in my case i would like to print `silly`, `debug`, `info` and `warn` to stdout and `error`, `fatal` to stderr, I should add 4 stream entries for the stdout and 2 for the stderr plus one for the file stream.

robertsLando:
@mcollina LGTM or should we change the implementation like @jsumners suggested?

mcollina:
I think @jsumners objection is valid, so let's implement that instead.

thetutlage:
In case we are looking for more feedback. How about using the existing `level` property, but allowing it to be an object as well with the following properties.

```ts
level: { min: 'info', max: 'warn' }
```

mcollina:
@thetutlage I think this PR stalled. Would you like to send a fresh one?

PR #1286, opened 2021-12-30
fix: passing stream to transport

### LINKED ISSUE #1285: Multiple transports doesn't work: The "path" argument must be of type string. Received undefined
I have a very simple transport setup as follows:

```js
transport = pino.transport({
    targets: [
      { target: 'pino-pretty' },
      {
        target: 'pino/file',
        options: {
          destination: `logs/${dayjs().format('YYYY-MM-DDTHH:mm:ss')}.log`
        }
      }
    ]
  })
```

However, pino shits the bed, complaining of the following:
```
The "path" argument must be of type string. Received undefined
TypeError: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:371:5)
    at validateString (node:internal/validators:119:11)
    at fixTarget (/home/node/node_modules/pino/lib/transport.js:116:9)
    at transport (/home/node/node_modules/pino/lib/transport.js:111:22)
    at normalizeArgs (/home/node/node_modules/pino/lib/tools.js:410:10)
    at pino (/home/node/node_modules/pino/pino.js:85:28)
    at Object.<anonymous> (/home/node/lib/logger.js:32:16)
    at Module._compile (node:internal/modules/cjs/loader:1101:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
```

Ultimately, I've tracked it down to the fact that despite the fact that I _specifically_ set `targets` property when initializing the transport, it's not there when the `ThreadStream` actually gets created (that's the transport "instance"). Here's the proof (I console.logged the actual transport instance that was being passed to the `function transport()` in `lib/transport.js`):
```
transport: ThreadStream {
    _events: [Object: null prototype] {
      ready: [Function: onReady],
      close: [Function (anonymous)]
    },
    _eventsCount: 2,
    _maxListeners: undefined,
    worker: Worker {
      _events: [Object: null prototype],
      _eventsCount: 4,
      _maxListeners: undefined,
      performance: [Object],
      stream: [FakeWeakRef],
      [Symbol(kCapture)]: false,
      [Symbol(kHandle)]: [Worker],
      [Symbol(kPort)]: [MessagePort [EventTarget]],
      [Symbol(kParentSideStdio)]: [Object],
      [Symbol(kPublicPort)]: [MessagePort [EventTarget]],
      [Symbol(kNewListener)]: [Function (anonymous)],
      [Symbol(kRemoveListener)]: [Function (anonymous)],
      [Symbol(kLoopStartTime)]: -1,
      [Symbol(kIsOnline)]: false
    },
    [Symbol(kCapture)]: false,
    [Symbol(kImpl)]: {
      stateBuf: [SharedArrayBuffer],
      state: [Int32Array],
      dataBuf: [SharedArrayBuffer],
      data: <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... 4194254 more bytes>,
      sync: false,
      ending: false,
      ended: false,
      needDrain: false,
      destroyed: false,
      flushing: false,
      ready: false,
      finished: false,
      errored: null,
      closed: false,
      buf: ''
    }
  }
}
```

Clearly, a whole lot of symbols and a whole not of `targets`. This is why when you get to the `lib/transport.js` and do this, the `targets` is empty and thus causes the "Received undefined" error:
```js
function transport (fullOptions) {
  const { pipeline, targets, options = {}, worker = {}, caller = getCaller() } = fullOptions
...
```

I think this should be a straightforward fix?

Thanks

climba03003:
throw if people pass a stream to `option.transport`.

Closes #1285

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -406,7 +406,7 @@ function createArgsNormalizer (defaultOptions) {
       stream = opts
       opts = {}
     } else if (opts.transport) {
-      if (opts instanceof SonicBoom || opts.transport.writable || opts.transport._writableState) {
+      if (opts.transport instanceof SonicBoom || opts.transport.writable || opts.transport._writableState) {
         throw Error('option.transport do not allow stream, please pass to option directly. e.g. pino(transport)')
       }
       stream = transport({ caller, ...opts.transport })
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -406,6 +406,9 @@ function createArgsNormalizer (defaultOptions) {
       stream = opts
       opts = {}
     } else if (opts.transport) {
+      if (opts instanceof SonicBoom || opts.transport.writable || opts.transport._writableState) {
+        throw Error('option.transport do not allow stream, please pass to option directly. e.g. pino(transport)')
+      }
       stream = transport({ caller, ...opts.transport })
     }
     opts = Object.assign({}, defaultOptions, opts)
--- a/test/transport/core.test.js
+++ b/test/transport/core.test.js
@@ -480,3 +480,20 @@ test('transport options with target and stream', async ({ fail, equal }) => {
     equal(err.message, 'only one of option.transport or stream can be specified')
   }
 })
+
+test('transport options with stream', async ({ fail, equal, teardown }) => {
+  try {
+    const dest1 = join(
+      os.tmpdir(),
+      '_' + Math.random().toString(36).substr(2, 9)
+    )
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm, thanks!

PR #1296, opened 2022-01-10
Call .end() on thread-stream before exiting the process.

mcollina:
Fixes https://github.com/pinojs/pino-pretty/issues/290


jsumners (approved):
Looks good to me.

PR #1297, opened 2022-01-11
fix: stricter type

### LINKED ISSUE #1293: TypeScript types do not provide expected safety
I recently migrated all my `console` logging statements to use a pino logger on a TypeScript backend codebase. Most of the changes worked out of the box with the matching default functions (e.g. `console.error(...)` mapping to `pino.error(...)`). 

The TS type checker let me happily move my `console.log(...)` statements to `pino.log(...)`, so I incorrectly assumed there was a default `log` method on the pino logger API. 

Obviously, there's no `log` function, so the service failed with error `TypeError: logger_1.logger.log is not a function`.

Totally my fault, but I expected the type system to help me here. Is there anything that can be done to get some help from TypeScript?

climba03003:
Fixes #1293

### Notable Changes:
- Remove `Record<string, any>` from `Logger`.
- Update `tsconfig.json`, I think the current `include` should be `exclude` because it should not compile the `test` file which may contain error.
- Reflect `customLevels` options and combine it with `Logger`.

I think it can provide even more stricter type on the `options` like `useOnlyCustomLevels`, `level`, etc. But it require a heavy refactor and times.


climba03003:
The CI failure is actually not related to my changes.

test/types/pino.test-d.ts
```diff
@@ -90,6 +91,7 @@ pino({
 });
 
 pino({ base: null });
+// @ts-expect-error
 if ("pino" in log) console.log(`pino version: ${log.pino}`);
```
climba03003, line 95: Not sure why it is inside the type check.
I cannot find the relevant code in js.

mcollina:
Should we update the doc?

mcollina (approved):
lgtm

climba03003:
> Should we update the doc?

I don't know which file is appropriate for the update. Currently, `pino` do not provide a `TypeScript` specific document.

mcollina:
Ah don't worry then. Let's just wait for @kibertoad.

kibertoad (approved):
This looks great!

MaikuMori:
I believe that this broke something.

Basically, `Logger<{customLevels}>` is not carried over after `logger.child()` call and I get type error on the child instance when using a custom level function.

```ts
const logger = pino({
  customLevels: {
    warning: 400,
  },
  useOnlyCustomLevels: true,
});
const child = logger.child({whatever:true});
// Type error.
child.warning("Hello there");
```

This was not the case before.

PR #1300, opened 2022-01-14
fix: allow transport target options to be optional and support custom log levels

techmunk:


mcollina (changes requested):
Can you please add a unit test? We use tsd to test our types.

Issue #1303, opened 2022-01-19
Pnpm and Pino Transports


moshie:
Hi,

I have created a private package called `@zopauk/nextjs-logger` the idea of this package is nextjs applications can install it and have access to the logger however, the global configuration (transports and such) remain under the package it's self (to keep applications aligned with logging config).

The issue we have with this is that when installing `@zopauk/nextjs-logger` with **pnpm** on to the nextjs application the app errors with the following:

```
Error: Cannot find module 'pino-pretty'
Require stack:
- /Users/david.hewitt/code/loans-dashboard/.next/server/pages/_document.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.mod._resolveFilename (/Users/david.hewitt/code/loans-dashboard/node_modules/.pnpm/next@12.0.8_dfc0b93d99fba272c9564d3139a2581a/node_modules/next/dist/build/webpack/require-hook.js:179:28)
    at Function.resolve (node:internal/modules/cjs/helpers:108:19)
    at fixTarget (/Users/david.hewitt/code/loans-dashboard/node_modules/.pnpm/pino@7.6.3/node_modules/pino/lib/transport.js:129:30)
    at transport (/Users/david.hewitt/code/loans-dashboard/node_modules/.pnpm/pino@7.6.3/node_modules/pino/lib/transport.js:112:22)
    at normalizeArgs (/Users/david.hewitt/code/loans-dashboard/node_modules/.pnpm/pino@7.6.3/node_modules/pino/lib/tools.js:412:16)
    at pino (/Users/david.hewitt/code/loans-dashboard/node_modules/.pnpm/pino@7.6.3/node_modules/pino/pino.js:84:28)
    at PinoLogger (/Users/david.hewitt/code/loans-dashboard/node_modules/.pnpm/@zopauk+nextjs-logger@1.1.6_next@12.0.8/node_modules/@zopauk/nextjs-logger/dist/nextjs-logger.cjs.development.js:885:13)
    at eval (webpack-internal:///./src/utils/logger.ts:9:99)
    at Object../src/utils/logger.ts (/Users/david.hewitt/code/loans-dashboard/.next/server/pages/_document.js:220:1) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/Users/david.hewitt/code/loans-dashboard/.next/server/pages/_document.js'
  ]
}
```

[...]

PR #1304, opened 2022-01-19
fix: test for multiple caller entries

moshie:
Fixes: #1303

### DIFF
--- a/lib/caller.js
+++ b/lib/caller.js
@@ -1,16 +1,10 @@
 'use strict'
 
-const fs = require('fs')
-
 function noOpPrepareStackTrace (_, stack) {
   return stack
 }
 
-function isOutsideNodeModules (input) {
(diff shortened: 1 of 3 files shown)


mcollina:

mcollina:
Does it fix your problem?

moshie:
yes it does 👍

moshie:
I was just looking at the documentation and noticed that the docs are specifying that caller is an option:
https://getpino.io/#/docs/api?id=pinotransportoptions-gt-threadstream

Here I have changed it to callers this may end up breaking some people's implementations.

I can update the docs and the implementation to accept caller and callers then wrap caller in an array?

moshie:
@mcollina This should be ready for a review whenever you're ready 👍

mcollina (approved):
lgtm

PR #1306, opened 2022-01-21
fix(types): child is not respect customLevels

climba03003:
Fixes: child is not respect parent `customLevels` options and nested child is not inherit the `customLevels` options.


climba03003:
Please check if it fix your issue.
cc @MaikuMori

mcollina (approved):
lgtm

MaikuMori:
The fix looks good, but I didn't test it locally.

PR #1316, opened 2022-02-02
Add thread locking action

jsumners:
https://github.com/dessant/lock-threads

https://github.com/dessant/lock-threads#why-are-only-some-issues-and-pull-requests-processed


tubbo:
shouldn't have been a release fellas :/

PR #1329, opened 2022-02-11
Add `flush` function to types

kblcuk:
After 5e31678 `Logger` doesn't allow arbitrary properties anymore, which
revealed that types are missing `.flush` function definition.


mcollina (approved):
lgtm

PR #1328, opened 2022-02-10
Add option `mixinMergeStrategy`

alexpts:
Added new option `mixinMergeStrategy` for define custom strategy merge `mergeObject` from log method argument and `mixin`

Example:
```js
// Custom muttable strategy, `mixin` has priority
const logger = pino({
    mixin() {
        return { tag: 'k8s' }
    },
    mixinMergeStrategy(mergeObject, mixinObject) {
        return Object.assign(mergeObject, mixinObject)
    }
})
logger.info({
    tag: 'local'
}, 'Message')
// {"level":30,"time":1591195061437,"pid":16012,"hostname":"x","tag":"k8s","msg":"Message"}
```

PR #1333, opened 2022-02-15
add option `mixinMergeStrategy` (#1328)

mcollina:
Forward port of #1328 to v7.


mcollina:
Closing as this does not work as I would think it should.

PR #1335, opened 2022-02-15
Refactor the mixing logic

mcollina:


jsumners (approved):
I like this a lot better. It was very unclear to me in the other PR if the input was being overwritten.

Issue #1327, opened 2022-02-08
How do you add custom stream to existing pino logger?


lxbc135:
How do you add custom stream to existing logger, like the older Probot's Bunyan logger `app.log.target.addStream(myStream)`?

Would like to upgrade to Probot v11+, which uses `pino` instead of `bunyan`.  So need pino equivalent of `addStream`.

Reference: [probot/issues/776](https://github.com/probot/probot/issues/776)

dbacarel:
I think you need to create your custom ```transport``` - see https://getpino.io/#/docs/transports

mcollina:
you will need to pass in a `pino.multistream()`, it has an `add` method that is not documented and it should be https://github.com/pinojs/pino/blob/ef5e2e64164af01967b033c186910a6eb6f4cd7a/lib/multistream.js#L124.

dbacarel:
I guess that this issue can be closed now

mcollina:
Thanks!

PR #1342, opened 2022-02-16
Added documentation about MultiStreamRes, StreamEntry, DestinationStr…

dbacarel:
Should solve #1327

This PR introduces 2 top-level sections in the documentation: `Interfaces` and `Types`
Each describes respectively the interfaces and types involved in the object returned by `pino.multistream()` which is otherwise undocumented.

Also:
- Added UTs for `pino.multistream().add` in `test/multistream.test.js`
- Added type test for `pino.multistream().add` in `test/multistream.test.js`
- `pino.multistream` and `pino.multistream.add` now throw Error if provided with an invalid stream object
- Minor related type alignments

### DIFF
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -6,9 +6,10 @@ const { levels } = require('./levels')
 const defaultLevels = Object.create(levels)
 defaultLevels.silent = Infinity
 
+const DEFAULT_INFO_LEVEL = levels.info
+
 function multistream (streamsArray, opts) {
   let counter = 0
-
   streamsArray = streamsArray || []
   opts = opts || { dedupe: false }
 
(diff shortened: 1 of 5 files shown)


lib/multistream.js
```diff
@@ -77,22 +78,41 @@ function multistream (streamsArray, opts) {
...
+      typeof dest.levelVal === 'number'
+        ? dest.levelVal
+        // Otherwise, if level property is provided as string
+        : typeof dest.level === 'string'
+          ? levels[dest.level]
+          // Otherwise, use it as is if a number
+          : typeof dest.level === 'number'
+            ? dest.level
+            // Default fallback level
+            : DEFAULT_INFO_LEVEL
```
mcollina, line 105: Would you mind to restore the previous if/else block?

mcollina (changes requested):

mcollina:
@jsumners  your action did not work :(

lib/multistream.js
dbacarel: of course, I restored the if/else block instead of the conditional operator - unless you meant reverting the whole change?

mcollina (approved):
lgtm

PR #1346, opened 2022-02-22
Fix workflow target main branch workflow

jsumners:


marcbachmann (approved):

.github/workflows/target-main.yml
```diff
@@ -1,10 +1,12 @@
+name: PR Target Check
+
 on:
   pull_request_target:
     types: [opened]
 
 jobs:
   comment:
-    if: ${{ github.base_ref }} != "master"
+    if: ${{ github.base_ref != "master" }}
```
marcbachmann, line 9: https://docs.github.com/en/actions/learn-github-actions/contexts#determining-when-to-use-contexts
that condition should be correct like that

PR #1352, opened 2022-02-26
Add LevelWithSilent as an exported type

JorgenPhi:
Causes issues when using pino-multistream in TypeScript


mcollina (commented):
Could you add a test for this? We use tsd.

JorgenPhi:
Tried it for a while, but I'm not familiar enough with tsd to do this properly. Sorry.

Fdawgs:
Superseded by #1399

PR #1359, opened 2022-03-04
Fix hasOwnProperty usage

joshkel:
Under certain circumstances, the user's mergingObject may be passed through as is to `write`. If the mergingObject was created with no prototype (`Object.create(null)`), then calling hasOwnProperty will fail.

Sample problematic code:

```js
const PinoMS = require('pino-multi-stream');
const pino = require('pino');
const logger = pino({}, pino.multistream([
  { stream: PinoMS.prettyStream({ prettyPrint: { colorize: true, translateTime: 'SYS:standard' } }) }
]));

const example = Object.create(null);
Object.assign(example, { hello: 'world' });
logger.info(example, 'hello world');
```

(This worked prior to Pino 7.8.0.)

For consistency, I also updated another usage of hasOwnProperty to match the new style.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -290,7 +290,7 @@ function prettifierMetaWrapper (pretty, dest, opts) {
       const formattedObj = formatters.log ? formatters.log(lastObj) : lastObj
 
       const messageKey = lastLogger[messageKeySym]
-      if (lastMsg && formattedObj && !Object.prototype.hasOwnProperty(formattedObj, messageKey)) {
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Good spot! Can you please add a unit test?

joshkel:
@mcollina Done. I apologize for the delay.

jsumners (approved):
Looks good to me.

mcollina (approved):
lgtm

Issue #1353, opened 2022-03-02
formatters level set label cant transport


zhongshanxian:
why set level => label, cant record the log, test.log is empty?

"pino": "^7.8.0"

```
const logger = pino({
    formatters: {
        level(label) {
            // if set level => label, cant record the log, test.log is empty
            // if set level => number, Everything is all right
            return { level: label };
        }
    },
    transport: {
        targets: [
            {
                target: 'pino/file',
                options: { destination: './test.log' }
            }
        ]
    }
});

export default logger
```

jsumners:
It is not clear what you are asking. Please review https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code and provide a minimal reproduction of your problem.

zhongshanxian:
https://codesandbox.io/s/cool-sun-t395f0?file=/README.md

mcollina:
`pino.multistream()` which underpins our transport system is incompatible with customizing the level. It would be better for pino to throw in that case.

guilhermelimak:
Hey @mcollina, I've opened a pull request to solve this issue and added some info about it to the docs, could you review it and let me know if there's something wrong?

techmunk:
The issue is actually just that level is not the numeric label. I was actually running with multiple transports, with a level formatter of
```
formatters: {
   level: (label, number) => ({ level: number, 'level-label': label })
},
```
And it was working fine. All transports got logged as expected.

Is there some other way to achieve an outcome such as this on the latest release of pino?

mcollina:
I don't think so, the system is not designed in that way. A PR that improves on it would be awesome.

techmunk:
@mcollina is there a reason the default formatter does not include the label, or an option to this effect? Would a change such as this be acceptable?

All I really want is the friendly level label to get to the logs so people don't need to know in logstash for example, that 30 is 'info'.

I'm happy to make a PR, but what kind of change would be acceptable here? Now that the code throws an error, I cannot do what I was doing before that worked just fine, which in my mind is a regression, even if it was not by design.

As I said, I'm happy to make a PR, but would appreciate some guidance on how such an outcome could be achieved.

techmunk:
@mcollina That could work, and I'd be happy to do that. I was just going through the code, and I have two other options that would solve my issue that might be more appropriate.

[...]

PR #1362, opened 2022-03-08
fix: Throw when using custom level formatters with multiple transports (#1353)

guilhermelimak:
This PR aim to fix an [issue](https://github.com/pinojs/pino/issues/1353) where if the user is using a custom level formatter and multiple transports at the same time logs will not be written and the user will not be notified.

Instead of failing silently the new behavior is to throw an error when both options are used at the same time (as mentioned [here](https://github.com/pinojs/pino/issues/1353#issuecomment-1059203373))


mcollina (approved):
lgtm

PR #1364, opened 2022-03-09
feat: add level number to mixin callback

techmunk:


techmunk:
@mcollina Those failed checks don't seem related. Can they be ignored? It doesn't seem like this patch could cause the reported problem.

mcollina (approved):
lgtm, this is great.

techmunk:
Force push to change commit author details

PR #1366, opened 2022-03-13
Fix: TypeScript: Make TransportOptions optional

glensc:
Options is not optional currently:
- https://github.com/pinojs/pino/blob/5d8eb719761003bce52d12ccc9e0a5b69ef9e185/pino.d.ts#L227

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -224,7 +224,7 @@ declare namespace pino {
 
     interface TransportTargetOptions<TransportOptions = Record<string, any>> {
         target: string
-        options: TransportOptions
+        options?: TransportOptions
         level: LevelWithSilent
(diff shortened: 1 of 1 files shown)


mcollina:
Could you fix the TS definitions instead? That's not mandatory.

glensc:
Dropped f3b439c8f19ad657ebce0b658416d2e2c6e501b6, added 931e06d.

glensc:
CI failing is not related to the changes. Retry shows random jobs failing.

mcollina (changes requested):
Could you add a test for the type? We use tsd.

mcollina:
Closing for no activity

Issue #1212, opened 2021-11-05
`formatters.bindings` are not called when creating a child


devonhumes:
I'm probably misunderstanding something, but the docs for `formatters.bindings` [state](https://getpino.io/#/docs/api?id=bindings)
```
Changes the shape of the bindings. The default shape is { pid, hostname }. 
The function takes a single argument, the bindings object. 
It will be called every time a child logger is created.
```
I understand this to mean that when `logger.child` is called the option for `binding` on the parent is called and the result is then used as `bindings` in the child. However, the code will only run a binding function when passed into the call `logger.child`

Example

```javascript
const logger = pino({
  formatters: {
    bindings(bindings) {  // NEVER RUN
      console.log('parent');
      return bindings
    }
  }
});

const child = logger.child({test: 'test'}); // the bindings from the parent is not run
const secondChild = logger.child({test: 'test'}, {
  bindings(bindings) {
    console.log('child');  // this is run, NOT the parent (which is expected, I think)
    return child;
  }
});
```

After testing and reviewing the relevant code I feel that the issue is with this code in `proto.js/child` [starting on line 85](https://github.com/pinojs/pino/blob/master/lib/proto.js#L85)

```javascript
  // correctly gets the parent formatters. At this point formatters.bindings is set to the value on the parent
  const formatters = this[formattersSym]; 
  ...
  // This code NEVER uses the `bindings` property from the parent. It ALWAYS overrides it.
  if (options.hasOwnProperty('formatters')) { 
    const { level, bindings: chindings, log } = options.formatters
    instance[formattersSym] = buildFormatters(
      level || formatters.level,
      chindings || resetChildingsFormatter,
      log || formatters.log
    )
  } else {
    instance[formattersSym] = buildFormatters(
      formatters.level,
      resetChildingsFormatter,
      formatters.log
    )
  }
  ...
  // This will use the newly created child which does not inherit the parent bindings and so they are never run.
  instance[chindingsSym] = asChindings(instance, bindings)
...
```

mcollina:
Good spot! What do you think should happen if a formatter is passed both to the child and the parent?

Anyhow.. 
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

devonhumes:
I would say that the parent formatter should be run for the child and the child formatter to be run if anyone calls `.child` on the child logger.

mcollina:
A PR would be nice

ZYSzys:
[...]

PR #1367, opened 2022-03-15
fix: Run parent bindings formatter when creating child logger

guilhermelimak:
Should fix pinojs/pino#1212.

I'm not entirely sure about my approach here, specially the `JSON.parse` part, I ran the benchmarks and didn't notice any regressions, but also got some random variation so I'm not 100% sure, but I didn't found another way to pass the right `bindings` argument to the parent formatter.

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -117,7 +117,12 @@ function child (bindings, options) {
   } else {
     instance[formattersSym] = buildFormatters(
       formatters.level,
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
Unfortunately this caused a regression (see https://github.com/pinojs/pino/issues/1378), I'll revert.

PR #1370, opened 2022-03-17
use pino/file instead of pino-pretty for caller transport tests

### LINKED ISSUE #1368: Caller test failure on CITGM
Pino is failing on CITGM in RHEL and Ubuntu systems

https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2880/nodes=rhel8-s390x/testReport/junit/(root)/citgm/pino_v7_8_1/
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2880/nodes=rhel8-x64/testReport/junit/(root)/citgm/pino_v7_8_1/
https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2880/nodes=ubuntu1804-64/testReport/junit/(root)/citgm/pino_v7_8_1/

```
# Subtest: when using a custom transport where all files in stacktrace are in the node_modules, the first file inside node_modules should be used
         not ok 1 - should match pattern provided
           ---
           diff: |
             --- expected
             +++ actual
             @@ -1,1 +1,1 @@
             -/done!/
             +""
           pattern: !re /done!/
           at:
             line: 52
             column: 5
             file: test/transport/caller.test.js
             type: Test
           stack: |
             Test.<anonymous> (test/transport/caller.test.js:52:5)
           source: |2
           
               t.match(actual, /done!/)
             ----^
             })
           ...
         
         1..1
         # failed 1 test
     not ok 3 - when using a custom transport where all files in stacktrace are in the node_modules, the first file inside node_modules should be used # time=618.84ms
```

However that test passes 100% of the time locally and on our CI. I think there is a bug lurking somewhere.

@mhdawson @targos could you replicate this problem on your machines (a `citgm pino` would do)? Do you think it would be possible to get access to one of those systems so that I can diagnose the problem?

mcollina:
This is due to https://github.com/pinojs/pino-pretty/issues/304 which
might trigger under severe load and short-lived scripts. The fix
on pino-pretty will happen on its own time but there is no need
to use pino-pretty in these tests.

Fixes #1368

### DIFF
--- a/test/fixtures/eval/index.js
+++ b/test/fixtures/eval/index.js
@@ -5,7 +5,7 @@ const pino = require('../../../')
 
 const logger = pino(
   pino.transport({
-    target: 'pino-pretty'
+    target: 'pino/file'
   })
 )
 
--- a/test/fixtures/eval/node_modules/file14.js
+++ b/test/fixtures/eval/node_modules/file14.js
@@ -3,7 +3,7 @@ const pino = require("../../../../");
 module.exports = function() {
     const logger = pino(
         pino.transport({
-            target: 'pino-pretty'
+            target: 'pino/file'
         })
     )
 
--- a/test/transport/caller.test.js
+++ b/test/transport/caller.test.js
@@ -1,53 +1,23 @@
 'use strict'
 
-const writer = require('flush-write-stream')
 const { join } = require('path')
 const { test } = require('tap')
 const execa = require('execa')
 
-const { once } = require('../helper')
-
 test('when using a custom transport outside node_modules, the first file outside node_modules should be used', async function (t) {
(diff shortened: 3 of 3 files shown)


richardlau:
@mcollina Is this the cause of this failure (in CITGM for the Node.js 14.19.1 proposal): https://ci.nodejs.org/job/citgm-smoker/2886
e.g. https://ci.nodejs.org/job/citgm-smoker/2886/nodes=rhel8-x64/testReport/junit/(root)/citgm/pino_v7_9_0/

mcollina:
No, for some unknown reasons a few tests have become flaky specifically on CITGM. This fixed the _other_ failure that was identified.

PR #1373, opened 2022-03-17
Fix pino.multristream test failure.

mcollina:
Using proxyquire is really a bad pattern but it's needed sometimes. However in this cases it has bitten us :/


mcollina:
cc @richardlau. this is the problem you saw on CITGM.

PR #1381, opened 2022-03-22
Ensure first-class support for transpiled & ts code in transports

### LINKED ISSUE #1243: V7 transports and Typescript
I've tried creating a super-simple wrapper for using pino-seq with the new worker thread transports introduced in Pino v7, but it looks like this won't work right out of the box, at least not with the setup we're running.

Pino logger transport config:
`transport: { target: path.join(__dirname, 'pinoSeqStream.js') }`

And the transport itself:
```
// pinoSeqStream.ts
import { createStream, SeqConfig } from 'pino-seq';

export default (config: SeqConfig) => {
  return createStream(config);
};
```
This transpiles into the following:
```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
const pino_seq_1 = require("pino-seq");
exports.default = (config) => {
    return (0, pino_seq_1.createStream)(config);
};
//# sourceMappingURL=pinoSeqStream.js.map
```

When running the node application, I just get the following error:
```
[...]\node_modules\thread-stream\lib\worker.js:38
  destination = await fn(workerData.workerData)
                      ^
TypeError: fn is not a function
```

I'm guessing this is down to how tsc transpiles this to JS? Any pointers on how one can create these new transports with Typescript? Are there any specific requirements for tsc/tsconfig.json that needs to be set?

I briefly looked at Typescript 4.5 and its ES modules support, but that was scrapped by the TS team for now, and is only available as an experimental feature in typescript@next.

castarco:
This PR upgrades the `thread-stream` dependency, applies a couple of changes on how transports are created, and adds new tests related to typescript & transpiled code.

- [X] Changes to improve support for typescript in transports (both transpiled and untranspiled)
- [X] Tests checking support for untranspiled code
- [x] Tests checking support for transpiled code

Fixes #1243 .


lib/worker-pipeline.js
```diff
@@ -23,6 +35,13 @@ module.exports = async function ({ targets }) {
         throw error
       }
     }
+
+    // Depending on how the default export is performed, and on how the code is
+    // transpiled, we may find cases of two nested "default" objects.
+    // See https://github.com/pinojs/pino/issues/1243#issuecomment-982774762
+    if (typeof fn === 'object') fn = fn.default
+    if (typeof fn === 'object') fn = fn.default
```
ramonmulia, line 43: duplicated code by mistake?

lib/worker.js
```diff
@@ -23,6 +35,13 @@ module.exports = async function ({ targets }) {
         throw error
       }
     }
+
+    // Depending on how the default export is performed, and on how the code is
+    // transpiled, we may find cases of two nested "default" objects.
+    // See https://github.com/pinojs/pino/issues/1243#issuecomment-982774762
+    if (typeof fn === 'object') fn = fn.default
+    if (typeof fn === 'object') fn = fn.default
```
ramonmulia, line 43: duplicated code?

lib/worker-pipeline.js
castarco: No, it's very strange and ugly, but it covers a real problem with some transpiled files 😢 .

lib/worker.js
castarco: same as in the previous case.

mcollina:
Could you add some docs?

castarco:
> Could you add some docs?
Sure. I'll check the broken builds as well.

mcollina:
That'd be awesome, feel free to add it. We can still rely on the timer if that's not available.

package.json
```diff
@@ -23,12 +23,13 @@
...
-    "test-ci-pnpm": "pnpm run lint && tap --no-coverage --no-check-coverage test/*test.js test/*/*test.js && pnpm run test-types",
-    "test-ci-yarn-pnp": "yarn run lint && tap --no-check-coverage test/*test.js test/*/*test.js --coverage-report=lcovonly",
+    "test": "npm run lint && npm run transpile && tap --ts test/*test.js test/*/*test.js test/**/*.test.ts && jest test/jest && npm run test-types",
+    "test-ci": "npm run lint && npm run transpile && tap --ts --no-check-coverage test/*test.js test/*/*test.js test/**/*.test.ts --coverage-report=lcovonly && npm run test-types",
+    "test-ci-pnpm": "pnpm run lint && npm run transpile && tap --ts --no-coverage --no-check-coverage test/*test.js test/*/*test.js test/**/*.test.ts && pnpm run test-types",
+    "test-ci-yarn-pnp": "yarn run lint && npm run transpile && tap --ts --no-check-coverage test/*test.js test/*/*test.js test/**/*.test.ts --coverage-report=lcovonly",
     "test-types": "tsc && tsd && ts-node test/types/pino.ts",
-    "cov-ui": "tap --coverage-report=html test/*test.js test/*/*test.js",
+    "transpile": "sh ./test/fixtures/ts/transpile.sh",
```
mcollina, line 31: Please convert this to a cross platform script. We support development on Windows.

jsumners (dismissed):

package.json
castarco: Done

lib/worker-pipeline.js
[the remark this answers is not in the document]
castarco: The  case where a functions is exported directly is already handled, as `fn` is directly obtained from that import, and we only try this re-assignment in case it's an object.

I'll try to move this into the try block as suggested.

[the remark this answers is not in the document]
castarco: I took another look to the code, and I'm not sure it brings any benefit to move these assignments inside the `try` block. If there's anything to be thrown (like `throw Error('exported worker is not a function')`) it would be automatically catch and re-thrown (making the error handling a bit slower), and, besides that, we should duplicate the assignments, because we have a case inside the `catch` block where we use "require" instead of import.

lib/worker-pipeline.js
```diff
@@ -14,7 +14,19 @@ module.exports = async function ({ targets }) {
     let fn
     try {
       const toLoad = 'file://' + t.target
-      fn = (await realImport(toLoad)).default
+
+      if (toLoad.endsWith('.ts') || toLoad.endsWith('.cts')) {
```
jsumners, line 18: Given the complexity added to this map function by this PR, can we extract it to a utility function that can be used by both `worker-pipeline.js` and `pipeline.js`.

castarco: Done

docs/transports.md
```diff
@@ -223,6 +223,22 @@ logger.info('hello world')
 __NOTE: there is no "default" destination for a pipeline but
 a terminating target, i.e. a `Writable` stream.__
 
+### TypeScript compatibility
+
+Pino provides basic support for transports written in TypeScript.
+
+Ideally, they should be transpiled to ensure maximum compatibility, but some
+times you might want to use tools such as TS-Node or TSM, to execute your
```
jsumners, line 231:
```suggestion
times you might want to use tools such as TS-Node or TSM, to execute your
```

Why mention "or TSM" if it isn't actually supported?

lib/worker-fn.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+
+const { realImport, realRequire } = require('real-require')
+
+async function getWorkerFunction (t) {
```
jsumners, line 5: I'd like to see a jsdoc block that describes the purpose of this function, its parameters, and what it returns/throws.

lib/worker-fn.js
```diff
@@ -0,0 +1,41 @@
...
+  // transpiled, we may find cases of two nested "default" objects.
+  // See https://github.com/pinojs/pino/issues/1243#issuecomment-982774762
+  if (typeof fn === 'object') fn = fn.default
+  if (typeof fn === 'object') fn = fn.default
+  if (typeof fn !== 'function') throw Error('exported worker is not a function')
+
+  return fn
+}
+
+module.exports = getWorkerFunction
```
jsumners, line 41: Move this to line 3.

test/transport/core.test.js
```diff
@@ -84,6 +84,32 @@ test('pino.transport errors if file does not exists', ({ plan, pass }) => {
   })
 })
 
+test('pino.transport errors if transport worker module does not export a function', ({ plan, equal }) => {
+  // TODO: add case for non-pipelined single target (needs changes in thread-stream)
```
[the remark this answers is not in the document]
castarco: There is not an explicit throw statement like in the code I wrote for pinojs, so when the loaded module does not export a function as expected, the error is different (similar, but not as clear & explicit as what we have here)

jsumners (commented):
I have two concerns with this PR:

1. I don't know who is going to maintain this new code as neither @mcollina or myself are TypeScript developers.
2. There are many TODOs littered throughout this new code. I think they should be resolved before this is actually ready to be merged. But I have removed my "request for changes", regardless.

castarco:
@jsumners I sympathise with your two points, they're valid concerns.

1. Regarding the first point, I see myself maintaining it, I'm a "TypeScript developer" and also a user of Pinojs for my own personal projects. Also, although not formally under Matteo's command, I work at NearForm too. NearForm usually allocates some resources to maintain certain packages of interest.
2. Regarding the second point, while I agree, I didn't want to fall into scope-creeping. The PR is already quite big as it is, and some of the highlighted problems in the TODO comments deserve experiments and research of their own. Maybe I should remove the comments and write proper issues instead.

What really concerns me about this PR... is how I still didn't manage to fix the flaky behaviour of the `big.test.js` test file 😞 .

docs/transports.md
castarco: dropped the reference.

mcollina:
@kibertoad could you review this? This might a bit fall on you to help with!

simoneb:
I know that @castarco cannot work on this any longer, is anybody else looking into this? Any input needed?

mcollina (approved):
lgtm

kibertoad:
will take a look shortly, can resolve conflicts as well

mcollina:
Thanks!

castarco:
@kibertoad I rebased to resolve the conflicts.

kibertoad (approved):

PR #1384, opened 2022-03-24
Update tap config

jsumners:
Moves the glob to the rc file so it doesn't have to be copied and pasted around the `package.json`.


mcollina (approved):
lgtm

PR #1389, opened 2022-03-31
fix incorrect type definition of censor function

mksmtn:
`censor` takes two arguments: value and path


mcollina (commented):
Could you add a unit test/assertion? We use tsd.

mksmtn:
never used tsd, hope these assertions are fine

mcollina (approved):
lgtm

PR #1395, opened 2022-04-06
Do not leak files when running tests

mcollina:
Signed-off-by: Matteo Collina <hello@matteocollina.com>


test/helper.js
```diff
@@ -97,4 +100,28 @@ function watchForWrite (filename, testString) {
...
 
-module.exports = { getPathToNull, sink, check, once, sleep, watchFileCreated, watchForWrite, isWin, isYarnPnp }
+let files = []
+let count = 0
+
+function file () {
+  const file = join(tmpdir(), `pino-${pid}-${count++}`)
+  files.push(file)
+  return file
+}
```
jsumners, line 110:
```suggestion
const files = []

function file () {
  const file = join(tmpdir(), `pino-${pid}-${crypto.randomUUID()}`)
  files.push(file)
  return file
}
```

PR #1398, opened 2022-04-07
Add option to pass custom levels down to `multistream()` from transport

### LINKED ISSUE #1189: Need both multistream and transport to accomplish a task
But they are mutually exclusive. Situation: I'm working on adding `pino` to a CLI tool where error logs will be written to `stderr` as straight messages (not ndjson), and program output will be written to `stdout` through a custom level (value: `100`), e.g. `log.output(csv)`.

First attempt is logging configuration like:

**logger.js**:
```js
'use strict';

const path = require('path');
const pino = require('pino');

module.exports = pino({
  level: 'output',
  customLevels: {
    output: 100
  },
  tansport: {
    targets: [
      // { target: 'pino/file', level: 'debug', options: { destination: 2 } },
      { target: path.resolve(path.join(__dirname, 'stderr.transport.mjs')), level: 'debug' },
      { target: 'pino/file', level: 'output', options: { destination: 1 } }
    ]
  }
});
```

**stderr.transport.mjs**:
```js
import { Writable } from 'stream';

export default (options) => {
  const myTransportStream = new Writable({
    autoDestroy: true,
    write(chunk, enc, cb) {
      const str = chunk.toString();
      let json = str;
      try {
        json = JSON.parse(str);
      } catch {}
      process.stderr.write(json.msg || json);
      cb();
    }
  });
  return myTransportStream;
};
```

But we know that won't work because setting the base logger to `level: 100` will prevent anything from ever being sent to the `stderr` transport. So we would consider using the multistream feature instead. Except, that won't work because we need our stream to be a transport.

What do we think should be done to support such use cases?

savardc:
Closes #1189

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -8,7 +8,7 @@ const time = require('./lib/time')
 const proto = require('./lib/proto')
 const symbols = require('./lib/symbols')
 const { configure } = require('safe-stable-stringify')
-const { assertDefaultLevelFound, mappings, genLsCache } = require('./lib/levels')
+const { assertDefaultLevelFound, mappings, genLsCache, levels } = require('./lib/levels')
 const {
   createArgsNormalizer,
   asChindings,
@@ -48,6 +48,7 @@ const hostname = os.hostname()
 const defaultErrorSerializer = stdSerializers.err
 const defaultOptions = {
   level: 'info',
+  levels,
   messageKey: 'msg',
   nestedKey: null,
   enabled: true,
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -96,6 +96,19 @@ const transport = pino.transport({
 pino(transport)
 ```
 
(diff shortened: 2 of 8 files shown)


mcollina (commented):
Could you add a test also for the JS part? Good work!

pino.js
```diff
@@ -48,6 +48,7 @@ const hostname = os.hostname()
 const defaultErrorSerializer = stdSerializers.err
 const defaultOptions = {
   level: 'info',
+  levels,
```
savardc, line 51: To avoid a circular dependency, I couldn't require the levels in tools.js. I added the default level here instead of adding a copy. This object is not exported and is only used by `createArgsNormalizer`. I'm not sure if there's a better solution?

mcollina (approved):
lgtm

mcollina:
Good work!

PR #1399, opened 2022-04-07
Add LevelWithSilent as an exported type

deongroenewald:
As per #1352, the community supported `@types/pino-multi-stream` package has a dependency on `LevelWithSilent`.

However, that PR is lacking any tests.

I've added a simple types test but open to feedback if there are bettter/different ways to test this.


mcollina (approved):
lgtm

Issue #1375, opened 2022-03-17
Flakyness in transport/core test


mcollina:
In https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2888/nodes=rhel8-s390x/testReport/junit/(root)/citgm/pino_v7_9_1/, we can see that the transport/core test is flaky:

```
not ok 34 - test/transport/core.test.js # time=7447.435ms
   ---
   env: {}
   file: test/transport/core.test.js
   timeout: 480000
   command: /home/iojs/build/workspace/citgm-smoker/smoker/bin/node
   args:
     - test/transport/core.test.js
   stdio:
     - 0
     - pipe
     - 2
   cwd: /home/iojs/tmp/citgm_tmp/50771e87-870e-4ae0-a6ae-3dd39d8c2ab4/pino
   exitCode: 1
   ...
 {
     # Subtest: pino.transport with file
         ok 1 - should be equivalent
         1..1
     ok 1 - pino.transport with file # time=337.506ms
     
     # Subtest: pino.transport with file (no options + error handling)
         ok 1 - should be equal
         1..1
     ok 2 - pino.transport with file (no options + error handling) # time=205.117ms
     
     # Subtest: pino.transport with file URL
         ok 1 - should be equivalent
         1..1
     ok 3 - pino.transport with file URL # time=202.347ms
     
     # Subtest: pino.transport errors if file does not exists
         1..1
         ok 1 - error received
     ok 4 - pino.transport errors if file does not exists # time=172.223ms
     
     # Subtest: pino.transport with esm
         ok 1 - should be equivalent
         1..1
     ok 5 - pino.transport with esm # time=207.977ms
     
     # Subtest: pino.transport with two files
         ok 1 - should be equivalent
         ok 2 - should be equivalent
         1..2
     ok 6 - pino.transport with two files # time=327.201ms
     
     # Subtest: pino.transport with an array including a pino-pretty destination
         ok 1 - should be equivalent
         ok 2 - should match pattern provided
         1..2
     ok 7 - pino.transport with an array including a pino-pretty destination # time=514.598ms
     
     # Subtest: no transport.end()
         ok 1 - should be equivalent
         1..1
     ok 8 - no transport.end() # time=208.295ms
     
     # Subtest: autoEnd = false
         ok 1 - should be equal
         ok 2 - should be equivalent
         1..2
     ok 9 - autoEnd = false # time=228.057ms
     
     # Subtest: pino.transport with target and targets
[...]

PR #1401, opened 2022-04-08
Add synchronous sleep in case of synchronous exit

mcollina:
Fixes https://github.com/pinojs/pino/issues/1375


jsumners (approved):
Ship it.
giphy

PR #1402, opened 2022-04-09
replace extreme destination with async destination

mksmtn:
`extreme` destination does not exist on `pino` anymore


mcollina (changes requested):
I would just drop all that v4/v5 section. It's probably  for the best, those are really old!

mksmtn:
You mean like that?

mcollina:
You'd need to delete it from https://github.com/pinojs/pino/blob/master/docsify/sidebar.md and the README. Thanks!

mksmtn:
Oh, didn't read the notice "Sorry, forked repositories are not currently searchable." and just thought that nothing is found for "legacy" query. Fixed

mcollina (approved):
lgtm

jsumners (approved):

PR #1408, opened 2022-04-19
fix(type): Allow stream arg to be undefined

kennethtxytqw:


mcollina (changes requested):
Can you please add a unit test? We use tsd.

climba03003:
Shouldn't it be allowed in the line above?

kennethtxytqw:
> Shouldn't it be allowed in the line above?

Odd... you are right. let me test it out again. Closing for now.

PR #1414, opened 2022-04-24
docs: add `pino-loki` transport

Julien-R44:
Hey ! Thanks for Pino, awesome logging library 👍

I just added my little `pino-loki` package that I recently rewrote in Typescript and v7+ compatible


mcollina (approved):
lgtm

PR #1424, opened 2022-04-30
chore: Enable codeql action

naveensrinivasan:
This action runs GitHub's industry-leading semantic code analysis engine, CodeQL, against a repository's source code to find security vulnerabilities.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql

https://github.com/ossf/scorecard/blob/main/docs/checks.md#sast

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,68 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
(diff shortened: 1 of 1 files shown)


mcollina:
Codeqql is likely more useful for apps than libraries. I highly suspect it will flag more false positives then needed here. Thanks, but we'll pass.

PR #1444, opened 2022-05-26
docs: add note about nestedKey to mixin section

### LINKED ISSUE #1441: nestedKey + mixin behavior change as of V7
We've been using nestedKey + mixin setting to add standard keys to the root level of the logger, but put caller data under a nested key.  We have a shared library with a factory func that creates a standardized logger, something like this:

```js
const createLogger = () => {
  return pino({
     ...
     nestedKey: 'data',
     mixin: () => {
       return { "ctx": { 'caller': 'some_func' } } // ignore the "static" nature here - this is just a sample
     },
  })
}
```

As of the v6 release, this used to let us segment where the metadata is coming from in our logs -- if it's at the root of the log, it's coming from pino or our standard lib mixin.  If it's in the 'data' field, it's from the caller.

```typescript
// on v6
logger.info({ "key": 1 }, "some log")
=> { "data": { "key": 1 }, "message": "some log", "ctx": { "caller": "some_func" }}
```

With the v7 release, our mixin is now mixed and nested:

```
// on v7+
logger.info({ "key": 1 }, "some log")
=> { "data": { "key": 1, "ctx": { "caller": "some_func" }}, "message": "some log" }
```

Is this the desired behavior? It was done properly with regards to semver, but the release notes don't really reflect that this was a breaking change.

If this is the desired behavior, I'd be happy to contribute some docs to mixin to make it a bit more clear how they work together.  But I'd also like to ask how we can get the original behavior back :)

mothershipper:
Fixes #1441 -- once I get a minute to play with the formatters, I'll try to submit a "recipe" or something similar to the "Help" docs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -178,6 +178,24 @@ logger.error('Message 2')
 If the `mixin` feature is being used merely to add static metadata to each log message,
 then a [child logger ⇗](/docs/child-loggers.md) should be used instead.
 
+As of pino 7.x, when the `mixin` is used with the [`nestedKey` option](#opt-nestedkey), 
+the object returned from the `mixin` method will also be nested. Prior versions would mix 
+this object into the root. 
+
+```js
+const logger = pino({
+    nestedKey: 'payload', 
+    mixin() {
+        return { requestId: requestId.currentId() }
+    }
+})
+
+logger.info({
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #1453, opened 2022-06-02
fix(publishing): add preval pino version

budarin:
Preval pino version before publishing the package in the prepare package stage

### DIFF
--- a/lib/utils/preval-pino-version.js
+++ b/lib/utils/preval-pino-version.js
@@ -1,0 +1,15 @@
+const fs = require('fs');
+const path = require('path');
+
+const pkgPath = path.resolve('./package.json');
+const metaPath = path.resolve('./lib/meta.js');
+
+const pkg = fs.readFileSync(pkgPath, { encoding: 'utf-8'});
+const {version} = JSON.parse(pkg);
+
+const metaContent = `'use strict';
+
+module.exports = { version: '${version}' };
+`;
+
+fs.writeFileSync(metaPath, metaContent, { encoding: 'utf-8'});
--- a/lib/utils/preval-pino-version.js
+++ b/lib/utils/preval-pino-version.js
@@ -1,10 +1,10 @@
-const fs = require('fs');
-const path = require('path');
-const { version } = require('../../package.json');
+const fs = require('fs')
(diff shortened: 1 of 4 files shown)


mcollina:
the approach we follow in Fastify is slightly different

There is https://github.com/fastify/fastify/blob/c37888c281087089c37b3e24f2d2fa61a3304ce4/package.json#L19 that verifies that the version in meta is the same of package.json.

We manually bump the version by running the update script (no prepare step).

budarin:
ок,I'll refactor the code

lib/utils/preval-pino-version.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs')
```
mcollina, line 1: move this file to the build/ folder

mcollina (changes requested):
Please take a look at how it's done in Fastify. This is still far from it.

There is no package.json script but a test to verify if the meta version number has been bumped.

mcollina (changes requested):
Could you add https://github.com/fastify/fastify/blob/1152a45f433b05b56a9c4206855d1322cecb6e83/package.json#L19

and 

https://github.com/fastify/fastify/blob/main/test/internals/version.test.js

mcollina (approved):
lgtm, thanks

test/internals/version.test.js
```diff
@@ -0,0 +1,15 @@
+'use strict'
+
+const fs = require('fs')
+const path = require('path')
+const t = require('tap')
+const test = t.test
+const fastify = require('../..')()
```
climba03003, line 7:
```suggestion
const pino = require('../..')()
```

test/internals/version.test.js
```diff
@@ -0,0 +1,15 @@
...
+
+test('should be the same as package.json', t => {
+  t.plan(1)
+
+  const json = JSON.parse(fs.readFileSync(path.join(__dirname, '..', '..', 'package.json')).toString('utf8'))
+
+  t.equal(fastify.version, json.version)
```
climba03003, line 14:
```suggestion
  t.equal(pino.version, json.version)
```

PR #1471, opened 2022-06-19
ci: update workflows

Fdawgs:
This PR:

-   Removes Git credentials/SSH keys after checkout as a security precaution by setting `persist-credentials` to false, they are not used after the initial checkout
-   Adds `--ignore-scripts` option to `npm install` commands to protect against malicious scripts in dependencies
-   Moves the `actions/dependency-review-action` into the main `ci.yml` workflow, and deletes the old `dependency-review.yml` workflow (brings it more into line with Fastify's CI style and makes it easier to maintain)
-   Declares the minimum permissions for CI workflows to run at either the workflow or job level, following principle of least privilege; see [related GitHub security post](https://securitylab.github.com/research/github-actions-building-blocks/)
-   Enables `concurrency` in `ci.yml`; see [related docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency), this allows a subsequently queued workflow run to interrupt previous runs in PRs
-   Removes the Snyk badge from the readme, as the `actions/dependency-review-action` does the same thing as Snyk, so it is no longer needed
-   Uses the latest GitHub `.gitignore` template, whilst also adding the pnpm lockfile
-	Sets the Node version used in `bench.yml` to `lts/*`, so it will always test with the current LTS, and it doesn't need to be updated when the LTS changes
-	Adds a conditional `if` to the automerge job in `ci.yml`; this stops the job from running if the user is not Dependabot, saving a few seconds CI run time


Fdawgs:
@mcollina as discussed, this is part of improving Pino's repo CI. If happy with this PR i'll make equivalents in the other repos.

mcollina (approved):
lgtm

mcollina:
go for it! This is amazing work!

PR #1481, opened 2022-07-01
Introducing Aibrake transport

enricodeleo:
Hi there,

I had the chance to build an Aibrake.io transport for one of my projects, I thought it could be handy for other people out there :)


mcollina (approved):
lgtm

PR #1488, opened 2022-07-07
docs(API): add description to `level` constructor option

KoltesDigital:
This option is described in [`logger.levels`](https://github.com/pinojs/pino/blob/master/docs/api.md#loggerlevels-object) but not in its own section.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -57,6 +57,8 @@ The name of the logger. When set adds a `name` field to every JSON line logged.
 
 Default: `'info'`
 
+The minimum level to log: Pino will not log messages with a lower level. Setting this option reduces the load, as typically, debug and trace logs are only valid for development, and not needed in production.
+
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

docs/api.md
```diff
@@ -57,6 +57,8 @@ The name of the logger. When set adds a `name` field to every JSON line logged.
 
 Default: `'info'`
 
+The minimum level to log, i.e. lower levels won't be outputted. This reduces the load, as typically, debug and trace messages are only useful for development, quite verbose, and not needed in production.
```
mcollina, line 60:
```suggestion
The minimum level to log: Pino will not log messages with a lower level. Setting this option reduces the load, as typically, debug and trace logs are only valid for development, and not needed in production.
```

mcollina: cc @jsumners

KoltesDigital: Updated to your suggestion.

PR #1496, opened 2022-07-17
resolves #1489 add test

### LINKED ISSUE #1489: Multistream should catch Error thrown by thread-stream
The `write` function provided by `thread-stream` can throw an Error of the worker has exited or is ending:

https://github.com/pinojs/thread-stream/blob/9bbbcd196234971c4e546590a496199863447e44/index.js#L213-L219

The multistream is calling `stream.write(data)` without a `try/catch` meaning that the Error will propagate:

https://github.com/pinojs/pino/blob/74893ae3ee4c0271459b909e39da1d1258ba0a4e/lib/multistream.js#L64

As a result, an error will be thrown by the logger function, for instance `logger.info()`.
In my opinion, the logger function should never throw an Error, and even more, when using multi streams where only one stream is "failing".

Alternatively, I think `thread-stream` should not throw an Error but instead emit an `error` to notify that the log cannot be written on this stream.

Mogztter:
resolves #1489

### DIFF
--- a/test/fixtures/noop-transport.js
+++ b/test/fixtures/noop-transport.js
@@ -1,0 +1,10 @@
+const { Writable } = require('stream')
+
+module.exports = () => {
+  return new Writable({
+    autoDestroy: true,
+    write (chunk, enc, cb) {
+      cb()
+    }
+  })
+}
--- a/test/multistream.test.js
+++ b/test/multistream.test.js
@@ -2,6 +2,7 @@
 
 const writeStream = require('flush-write-stream')
 const { readFileSync } = require('fs')
+const { join } = require('path')
 const test = require('tap').test
 const pino = require('../')
 const multistream = pino.multistream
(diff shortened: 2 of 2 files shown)


lib/multistream.js
```diff
@@ -103,11 +103,12 @@ function multistream (streamsArray, opts) {
       level = DEFAULT_INFO_LEVEL
     }
 
+    const id = counter++
     const dest_ = {
       stream: stream_,
       level,
       levelVal: undefined,
-      id: counter++
+      id
```
mcollina, line 111: Can you explain why this change fixes it?

mcollina (changes requested):

lib/multistream.js
Mogztter: Sorry, it's an unrelated change. At first, I thought it would be useful to listen to the `error` event and automatically remove the stream from the stream list (using the  stream id).

But, as you've mentioned, I think users should listen to the `error` event and decide what to do.

I can revert this change if needed.

The real fix is the `thread-stream` upgrade to 2.0.0.
If you revert this change (i.e., revert thread-stream to 1.x) the added test will fail because it will throw an error.

Does it make sense?

mcollina: > I can revert this change if needed.

Thanks, that'd be awesome.

> The real fix is the thread-stream upgrade to 2.0.0.
If you revert this change (i.e., revert thread-stream to 1.x) the added test will fail because it will throw an error.

Awesome confirmation!

mcollina (approved):
lgtm

lib/multistream.js
Mogztter: Done, thanks for your review!

PR #1507, opened 2022-07-26
feat: add errorKey option

qwelias:
We have custom `messageKey` and `serializers`, makes sense to have custom `errorKey` as well.
Not sure if that should affect `browser` part


mcollina:
There seem to be a significant decrease of throughput in our benchmarks.

Current:

```
benchPino*10000: 148.753ms
```

this branch:

```
benchPino*10000: 166.715ms
```

mcollina:
CI is failing too.

qwelias:
Changes have minimal logic, so I don't think there's any room for perf improvement.
They don't affect any transport functionality either, I assume tests may be flaky, may need a rerun

jsumners:
> @mcollina How did you run your benchmarks? 1st few runs are usually faster because CPU is cold. Make sure to do a few consecutive runs on each branch to avoid CPU throttling bias.
> 
> For me both versions run `bench-basic` with `Pino average` in 96-104 (master actually peaked at 107 once, so even slower, did about dozen runs each)

See https://github.com/pinojs/pino/runs/7528251758?check_suite_focus=true

qwelias:
@jsumners 
I can see no significant difference:
- `benchmark branch` BASIC `benchPino*10000: 183.329ms`
- `benchmark current` BASIC `benchPino*10000: 188.812ms`

So idk where @mcollina took the numbers from

mcollina:
we run the benchmarks for every CI run

mcollina:
the first in this PR

qwelias:
@mcollina 
I'm sorry, but I see `benchPino*10000: 188.812ms` [here's link](https://github.com/pinojs/pino/runs/7528251758?check_suite_focus=true#step:5:16).

Can you link yours? I literally have no idea where you see those numbers, search doesn't find them either

mcollina:
current: https://github.com/pinojs/pino/runs/7528251758?check_suite_focus=true#step:5:24
this branch: https://github.com/pinojs/pino/runs/7528251630?check_suite_focus=true#step:5:24

qwelias:
Thanks!
But why are they called the same!?

jsumners:
@mcollina check the averages:

+ current: https://github.com/pinojs/pino/runs/7528251758?check_suite_focus=true#step:5:154
+ branch: https://github.com/pinojs/pino/runs/7528251630?check_suite_focus=true#step:5:154

mcollina:
There is still a difference.

qwelias:
What's average margin of error? Is the difference more than average?
Is it significant to block the feature?

mcollina:
I tested on a dedicated workstation, and there is no perceptible slowdown. We can land.

mcollina:
Don't worry about the flaky test.

mcollina (approved):
lgtm

jsumners (approved):

Issue #1506, opened 2022-07-25
Multistream dedupe example in documentation does not work


celebro:
The following example from documentation about `dedupe` flag in `multistream` (https://github.com/pinojs/pino/blob/master/docs/api.md#pino-multistream) does not work as intended, it only logs `info` and `error`, while `debug` and `fatal` are ignored.

```javascript
var pino = require('pino')
var multistream = pino.multistream
var streams = [
  {stream: process.stdout},
  {level: 'error', stream: process.stderr},
]

var opts = {
    levels: {
        silent: Infinity,
        fatal: 60,
        error: 50,
        warn: 50,
        info: 30,
        debug: 20,
        trace: 10
    },
    dedupe: true,
}

var log = pino({
  level: 'debug' // this MUST be set at the lowest level of the
                // destinations
}, multistream(streams, opts))

log.debug('this will be written ONLY to process.stdout')
log.info('this will be written ONLY to process.stdout')
log.error('this will be written ONLY to process.stderr')
log.fatal('this will be written ONLY to process.stderr')
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

celebro:
I'm afraid I can't dedicate the time for implementing a fix, I can only add what I'm currently using instead is specifying a stream for every log level:

```javascript
var streams = [
    { level: 'debug', stream: process.stdout },
    { level: 'info', stream: process.stdout },
    { level: 'warn', stream: process.stdout },
    { level: 'error', stream: process.stderr },
    { level: 'fatal', stream: process.stderr }
];
```

bl00dhound:
Hi, there!
There are two parts of problem:
* if`dedup` is set to`true` log is sent to the stream with[ the same level only](https://github.com/pinojs/pino/blob/master/lib/multistream.js#L63). That's the reason why we can't find log with `fatal` level in `process.stderr` stream. I've fixed it in [my PR](https://github.com/bl00dhound/pino/commit/f631737f821cd36b00919b7c17c47a4fcd14cf40). 
[...]

PR #1508, opened 2022-08-03
Fix: dedup works with multistream (refs: #1506)

bl00dhound:
If `dedupe` option is set to `true` we need to iterate by streams array from last to first items.


mcollina (approved):
lgtm

PR #1516, opened 2022-08-11
Simplify Google logging-specific configuration example

andrew-demb:
Reuse the `messageKey` config key instead of custom formatter


mcollina (approved):
lgtm

PR #1530, opened 2022-08-25
Fix transports doc pipeline example

AceFire6:
[This thread](https://github.com/pinojs/pino-syslog/issues/15) gave me the clue I needed to get this example working. It seems like `pino-pretty` requires `NDJSON` input. In the example the newline is omitted which means the logs will never be emitted.

This PR updates the example to include a newline after the `JSON.stringify` to fix that.


mcollina (approved):
lgtm, good spot!

PR #1533, opened 2022-08-27
docs(help): use formatters since useLevelLabels is deprecated

arkerone:
Since `useLevelLabels` option is deprecated, we can use the `formatters `option with a `level` function to format the level.

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -154,9 +154,29 @@ for information on this is handled.
 <a id="level-string"></a>
 ## Log levels as labels instead of numbers
 Pino log lines are meant to be parseable. Thus, Pino's default mode of operation
-is to print the level value instead of the string name. However, while it is
-possible to set the `useLevelLabels` option, we recommend using one of these
-options instead if you are able:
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

HengCC:
```js

import pino from 'pino';
import { logURL, nodeName } from './config.js';

const logger = pino({
  level: "info",
  messageKey: "message",
  errorKey: "error",
  formatters: {
    level: (label) => {
      return { level: label }
    }
  },
  base: {
    APP_NAME: "xxxx,
    NODE_NAME: nodeName,
  },
  transport: {
    targets: [
      {
        target: 'pino-pretty',
        options: {
          colorize: true,
          singleLine: true,
        }
      },
      {
        target: 'pino-socket',
        options: {
          address: logURL,
          port: 5000,
          mode: 'tcp'
        },
      }
    ]
  }

});

export default logger;
```

this will cause an error  , `Error: option.transport.targets do not allow custom level formatters`,

Issue #1534, opened 2022-08-29
Get access to children instances from parent


Diabl0269:
Hey,
I am using `pino` in a big project containing multiple children. I want to be able to change the log level on all loggers at once.
I saw there is another issue (https://github.com/pinojs/pino/issues/1330) about a similar scenario in which the answer was to create an array of loggers.
I would like to suggest to implement an array of children on a parent myself if you'd accept a PR regarding the subject.
The PR will:
1. Add `children` array to the `prototype` object [here](https://github.com/pinojs/pino/blob/master/lib/proto.js#L53).
2. Push the newly created child at the end of its creation function [here](https://github.com/pinojs/pino/blob/master/lib/proto.js#L144).

Please let me know your thoughts and in hope you agree I will create a PR :)

mcollina:
We want the children to be lightweight and be able to create a new logger for each incoming and http request. Therefore tracking each children from the parent will be a memory leak.

What you'd like to achieve could be done in an external module. I _think_ https://github.com/pinojs/pino-arborsculpture does what you need.

Diabl0269:
@mcollina Firstly, thanks for the quick response.
Secondly, what do you think about allowing a to pass a callback function on creation of an instance (which will be also passed to the children, same as the other properties).
Then the users of `pino` (e.g me), could pass such a callback in a single place to do a certain operation on the newly created instance.

About your point regarding memory leak, I understand the issue in implementing the original request as for some use cases it could actually be dangerous, but for my use case there is no risk of this. (All loggers are born and die on the service startup and shutdown).

Diabl0269:
Issue accidentally closed. Please let me know what do you think of my last comment.

mcollina:
yeah, I'll be ok with an `onChild` callback

jonathansamines:
@mcollina Would it be feasible for pino to track all child instances internally using an iterable weakset implementation and transparently change the log level when the parent level is updated?

mcollina:
That technique you are mentioning adds quite a bit of overhead in a hot path.

Diabl0269:
@mcollina I have a PR I'm ready to push but missing access rights to push my branch, could you grant the needed permission?

mcollina:
You should fork the repository (https://www.freecodecamp.org/news/how-to-fork-a-github-repository/)

Diabl0269:
I am new to contributing to open source projects, so thank you for the explanation!
[...]

PR #1541, opened 2022-08-31
Added `onChild` callback with tests and documentation

Diabl0269:
As discussed in this issue https://github.com/pinojs/pino/issues/1534, i've added an `onChild` callback that uses the child instance as its argument.
I also included tests and documentation.

Please let me know if anything should be change :)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -509,7 +509,8 @@ when using the `transport` option. In this case an `Error` will be thrown.
 
 #### `onChild` (Function)
 
-The `onChild` function is a synchronous callback that will be called on each creation of a new child, passing the child instance as its first argument.
+The `onChild` function is a synchronous callback that will be called on each creation of a new child, passing the child instance as its first argument.  
+Any error thrown inside the callback will be uncaught and should be handled inside the callback.
 ```js
 const parent = require('pino')({ onChild: (instance) => {
   // Exceute call back code for each newly created child.
--- a/docs/api.md
+++ b/docs/api.md
@@ -509,7 +509,7 @@ when using the `transport` option. In this case an `Error` will be thrown.
 
 #### `onChild` (Function)
 
-The `onChild` function is a callback that will be called on each creation of a new child, passing the child instance as its first argument.
+The `onChild` function is a synchronous callback that will be called on each creation of a new child, passing the child instance as its first argument.
 ```js
 const parent = require('pino')({ onChild: (instance) => {
   // Exceute call back code for each newly created child.
--- a/docs/api.md
+++ b/docs/api.md
@@ -508,6 +508,18 @@ when using the `transport` option. In this case an `Error` will be thrown.
 
 * See [pino.transport()](#pino-transport)
 
+#### `onChild` (Function)
+
+The `onChild` function is a callback that will be called on each creation of a new child and use the child instance as it's first argument.
+```js
+const parent = require('pino')({ onChild: (instance) => {
+  // Exceute call back code for each newly created child.
+}})
+// `onChild` will now be executed with the new child.
+parent.child(bindings)
+```
+
+
 <a id="destination"></a>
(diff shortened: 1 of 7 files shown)


mcollina (commented):
Good work! Can you please add a test for the type in https://github.com/pinojs/pino/blob/master/test/types/pino.test-d.ts?

Diabl0269:
@mcollina - I see some tests are failing on the CI.
The benchmark seems like it just hangs, on my machine it passes fairly quick.
I also checked macOS with version 18 and it passes, about windows I can't currently check locally.
What should I do about these?

Diabl0269:
Thanks you!
First time I use `tsd` so I hope I did it right.
Over then that I updated the `onChild` type to recognize it as an initialization option as well as a property on the logger. (The JS actually works for both but the type was missing).

mcollina (approved):
lgtm

Diabl0269:
Thanks again!
I don't have permission to merge the PR, will you give me the permissions/merge it?

mcollina:
Let's wait a few more reviews, then I'll merge and ship this.

docs/api.md
```diff
@@ -508,6 +508,18 @@ when using the `transport` option. In this case an `Error` will be thrown.
 
 * See [pino.transport()](#pino-transport)
 
+#### `onChild` (Function)
+
+The `onChild` function is a callback that will be called on each creation of a new child and use the child instance as it's first argument.
```
jsumners, line 513:
```suggestion
The `onChild` function is a callback that will be called on each creation of a new child, passing the child instance as its first argument.
```

jsumners, line 513: Does the callback block child creation? Do we handle any errors that occur within the callback?

test/basic.test.js
```diff
@@ -725,3 +725,13 @@ test('throws if prettyPrint is passed in as an option', async (t) => {
...
 })
+
+test('Should invoke `onChild` with the newly created child', async ({ equal }) => {
+  let innerChild
+  const child = pino({
+    onChild: (instance) => {
+      innerChild = instance
+    }
+  }).child({ foo: 'bar' })
+  equal(child, innerChild)
```
jsumners, line 736: How do we know that `innerChild` has been set when this assertion is invoked? Shouldn't the assertion be within the callback and a `t.plan` added to the start of the subtest?

jsumners (changes requested):

docs/api.md
Diabl0269: Currently no. Should we wrap the callback in a `try-catch` block?

test/basic.test.js
Diabl0269: The `onChild` function runs [here](https://github.com/pinojs/pino/pull/1541/files#diff-78c851c106059931b9042837eb4d521a7650b32535f306d9cb85bbddef22f877R144), right before the `child` instance is retuned by the `child` function, so that's how we could know that it's been set.
Over then that we couldn't make the assertion inside the callback since it will only have access to the `instance` and not the returned value `child`.
Please let me know if you still think any changes should be made.

jsumners: Are we asserting that the callback must be synchronous? If so, the documentation should be updated to make that clear.

docs/api.md
jsumners: I think the way the callback is handled should be made clear in the documentation. Adding a `try/catch` during child creation will have a performance penalty. So we would want to alter the documentation to make it clear that any error in the callback will be uncaught.

test/basic.test.js
Diabl0269: You are right, I've updated the documentation.

docs/api.md
Diabl0269: Got it. I've updated the documentation to mention the callback doesn't handle errors.

mcollina:
@jsumners PTAL

jsumners (approved):
Looks good to me.

PR #1542, opened 2022-08-31
Add missing type browser.disabled

hironytic:
Add missing `disabled` as an exported type for TypeScript.

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -543,6 +543,11 @@ declare namespace pino {
                  */
                 send: (level: Level, logEvent: LogEvent) => void;
             };
+
+            /**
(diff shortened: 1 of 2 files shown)


mcollina:
I'm not convinced about landing this. The main use case for pino is on the server and the `disable` property is not present there.
Possibly we should reassess how we support browsers to provide a better typing exerience.

mcollina:
wdyt @kibertoad ?

hironytic:
FYI: Related PR is #1480

moshie:
There is an enabled flag?
https://github.com/pinojs/pino/blob/master/docs/api.md#enabled-boolean

`disabled` is also very vague imo as it could be misinterpreted as disabling the whole browser functionality including the transmit function where as this should just control logging to the console.

PR #1547, opened 2022-09-01
Revert "build(deps): bump jsumners/lock-threads from b27edac0ac998d42b2815e122b6c24b32b568321 to 3"

jsumners:
Reverts pinojs/pino#1545


kibertoad:
@jsumners why play with commits? why not use numbered versions?

jsumners:
There isn't a version. It's a fork.

PR #1559, opened 2022-09-15
fix: add `this` to `hooks.logMethod`

SimenB:
This way, one can access e.g. `this.levels`


mcollina (approved):
lgtm

SimenB:
I don't _think_ the failing test is on me, seeing as I've only changed type definitions

PR #1561, opened 2022-09-16
fix: type `formatters.log` with a Record type

SimenB:
We have a `massageLogData` which takes an object with some custom types - `object` is not assignable to it. `object` just means non-primitive - it doesn't mean `Object` (which is discouraged, see https://github.com/typescript-eslint/typescript-eslint/blob/739479692bedb4aeca8c3d0c657f88d9c60f1fe0/packages/eslint-plugin/src/rules/ban-types.ts#L86-L88)


mcollina (approved):
lgtm

mcollina (approved):
lgtm

PR #1569, opened 2022-10-06
fix: remove pretty print in types, docs

susmithagudapati:
Screenshot 2022-09-27 at 5 15 43 PM

`pretty-print` option was deprecated in older releases and is meant to throw an error in `normalizeArgs` method in latest releases. Hence removing it from types as typescript gives a wrong sense that the `pretty-print` option is allowed while the js script throws an error. 

This PR ensures ts throws an error if `pretty-print` is used in the application and corrects the `pretty` part of the documentation.

Test cases are already written ensuring `the Pino` instantiation throws an error for `pretty-print`.


mcollina (approved):
lgtm

PR #1576, opened 2022-10-20
Pass logger instance as an argument in `level-change` event

### LINKED ISSUE #1572: level-change workaround doesn't work when `once` is used instead of `on`
There is a workaround for the `level-change` event described [here](https://github.com/pinojs/pino/blob/master/docs/api.md#level-change), which allows users to ignore `level-change` events created by child loggers. It appears this doesn't work when using `once` instead of `on`.

### Working code (Using `on`)
The given code sample using `on` works fine:
```
logger.on('level-change', function myHandler(lvl, val, prevLvl, prevVal) {
  if (logger !== this) {
    return
  }
  console.log('%s (%d) was changed to %s (%d)', prevLvl, prevVal, lvl, val)
})
logger.child({}); // trigger an event by creating a child instance, notice no console.log
logger.level = 'trace' // trigger event using actual value change, notice console.log
```
This correctly logs:
```
info (30) was changed to trace (10)
```

### Not working code (Using `once`)
The same thing does **NOT** work when replacing `on` with `once`:
```
logger.once('level-change', function myHandler(lvl, val, prevLvl, prevVal) {
  if (logger !== this) {
    return
  }
  console.log('%s (%d) was changed to %s (%d)', prevLvl, prevVal, lvl, val)
  logger.once('level-change', myHandler)
})
logger.child({});
logger.level = 'trace'
```

This incorrectly logs:
```
info (30) was changed to info (30)
info (30) was changed to trace (10)
```

bienzaaron:
closes #1572

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1003,12 +1003,13 @@ The logger instance is also an [`EventEmitter ⇗`](https://nodejs.org/dist/late
 
 A listener function can be attached to a logger via the `level-change` event
 
-The listener is passed four arguments:
+The listener is passed five arguments:
 
 * `levelLabel` – the new level string, e.g `trace`
 * `levelValue` – the new level number, e.g `10`
 * `previousLevelLabel` – the prior level string, e.g `info`
 * `previousLevelValue` – the prior level number, e.g `30`
+* `logger` – the logger instance from which the event originated
 
 ```js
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

PR #1581, opened 2022-10-30
fix: Redact mutates the original object

### LINKED ISSUE #1575: Redact mutates the original object
We noticed that redact is mutating the original object, and not bringing back its properties.

Snippet to reproduce:

```ts
import pino from 'pino'

const config: pino.LoggerOptions = {
  level: 'debug',
  redact: [ // Eliminate these fields from the logs!
    'context.password',
    'password',
    'context.*.password',
    'context[*].password'
  ]
}

const logger = pino(config)

const test = {
  whatever: { password: '1234' }
}

logger.info({ context: test }, 'info')

console.log(test) //  { whatever: { password: '[Redacted]' } }

```

Versions used:
```json
"pino": "7.11.0",
"pino-pretty": "7.6.1"
```

debadutta98:
Hi @mcollina,

This PR Closes #1575 

## Work 

- Change `lib/tools.js`
- Update test case in `test/redact.test.js`
- Use an external library [clonedeep](https://www.npmjs.com/package/lodash.clonedeep) to deep clone an object

Thank you!! waiting for your response

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -6,7 +6,6 @@ const format = require('quick-format-unescaped')
 const { mapHttpRequest, mapHttpResponse } = require('pino-std-serializers')
 const SonicBoom = require('sonic-boom')
 const onExit = require('on-exit-leak-free')
-const cloneDeep = require('lodash.clonedeep')
 const {
   lsCacheSym,
   chindingsSym,
@@ -26,7 +25,7 @@ const {
 const { isMainThread } = require('worker_threads')
 const transport = require('./transport')
 
-function noop () {}
+function noop () { }
 
 function genLog (level, hook) {
   if (!hook) return LOG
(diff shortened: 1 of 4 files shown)


mcollina (changes requested):
The whole goal of the redact feature is to not incur in the cost of a full depth clone.

lib/tools.js
```diff
@@ -6,6 +6,7 @@ const format = require('quick-format-unescaped')
 const { mapHttpRequest, mapHttpResponse } = require('pino-std-serializers')
 const SonicBoom = require('sonic-boom')
 const onExit = require('on-exit-leak-free')
+const cloneDeep = require('lodash.clonedeep')
```
jsumners, line 9: In addition to Matteo's comment, we would not pull _any_ lodash dependencies into this project. In particular, the `lodash.*` packages are no longer maintained (last I knew). You should look at https://www.npmjs.com/package/rfdc

kibertoad:
I'm -1 on this. If redact docs don't mention the mutation, they should, but for perf reasons user should be the one responsible for cloning if they need it, as this is not a common need.

jsumners:
> I'm -1 on this. If redact docs don't mention the mutation, they should, but for perf reasons user should be the one responsible for cloning if they need it, as this is not a common need.

They do note that the mutation is the core of the optimization. But they also state that the original values should be restored after redaction.

debadutta98:
> The whole goal of the redact feature is to not incur in the cost of a full depth clone.

Hi @mcollina, Thanks for your time!

The changes are done successfully and I think this [commit](https://github.com/pinojs/pino/pull/1581/commits/2a50d3e96013f732fcdc460353398a7bacabddfe) meets your expectations

mcollina:
This is significantly better however it's still cloning everything before redacting. The generic approach of `fast-redact` is to clone in place. It's possible to fix those bugs there.

cc @davidmarkclements

debadutta98:
> This is significantly better however it's still cloning everything before redacting. The generic approach of `fast-redact` is to clone in place. It's possible to fix those bugs there.
> 
> cc @davidmarkclements

Is there any alternative idea that you want to recommend here?

davidmarkclements:
@debadutta98 the bug will need chasing down in fast-redact, PR's definitely welcome as I'm quite time constrained atm but I will get to it at some (potentially distant) point if not

debadutta98:
> @debadutta98 the bug will need chasing down in fast-redact, PR's definitely welcome as I'm quite time constrained atm but I will get to it at some (potentially distant) point if not

sure, i will

PR #1583, opened 2022-11-03
[Enhancement] Update to named parameters

### LINKED ISSUE #1577: Move to named parameters
We should move to named parameter functions in the next release. For example, with #1576 we tacked on another parameter to the `level-change` event:

https://github.com/pinojs/pino/blob/085401b659dd491051fd1342e7d5fd5c6d70e2db/docs/api.md?
plain=1#L1027-L1032

While the function signature of an event listener is easier to understand than one someone invokes themselves, I still think named parameters is a better experience.

debadutta98:
This PR Closes #1577 

Update simple parameters to named parameters of LevelChangeEventListener so the users can easily identify the type of parameter

## Work

- Update `pino.d.ts`
- Update `doc/api.md`

Thank you!

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1023,15 +1023,15 @@ Please note that due to a [known bug](https://github.com/pinojs/pino/issues/1006
 fire a `level-change` event. These events can be ignored by writing an event handler like:
 
 ```js
-const logger = require('pino')()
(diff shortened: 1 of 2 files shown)


jsumners:
That issue still needs discussion. We don't want to do this one method at a time. It would be a major change.

Issue #1491, opened 2022-07-10
Recommended setup for production


david-szabo97:
The documentation contains a mixed use of `pino.transport`, `pino.destination`, `pino.multistream`, and Pino created without arguments. Legacy vs Pino v7+ transports is also confusing.

What's the recommended setup for production? Production = least overhead, best performance, logging to stdout

My first guess would be

```js
const logger = pino(pino.destination());
```

After reading the documentation, I've found a paragraph:
> The difference between using the pino/file transport builtin and using pino.destination is that pino.destination runs in the main thread, whereas pino/file sets up pino.destination in a worker thread.

This makes `pino.destination` a wrong choice because it runs in the main thread. Therefore to get the best performance out of Pino we must use v7+ transports.

In our case we want to log to stdout, therefore we create a `pino/file` transport:

```js
const transport = pino.transport({
  target: "pino/file",
});

const logger = pino(transport);
```

Having a section in the documentation about running Pino in production would be great. Unfortunately, there are too many options at the moment, which can confuse new users of Pino (or even existing users like myself).

mcollina:
The default settings are defined to work well in your simple production setup. You don't need to tweak anything.

All the above are required if you have more sophisticated requirements.

Would you like to send a PR to clarify this further in our docs?

david-szabo97:
### By default settings you mean this, right?

```js
const logger = pino();
```

According to the documentation, the default `destination` is `pino.destination(1)`.
So the above is equivalent to the following:

```js
const logger = pino(pino.destination(1));
```

As far as I understand, `pino.destination` runs on the main thread in this case. So the following _should_ perform better because the logging is delegated to a worker thread:
```js
const transport = pino.transport({
  target: "pino/file",
});

const logger = pino(transport);
```

### Benchmark - based on the `basic-bench.js` - seems to agree with my assumptions:
```
benchPino*10000: 372.211ms
benchPinoMinLength*10000: 132.718ms
benchPinoNodeStream*10000: 417.469ms
benchPDestinationDevNull*10000: 346.205ms
benchPfile*10000: 77.384ms
```

`benchPDestinationDevNull` = `pino(pino.destination("/dev/null"))`
`benchPfile` = `pino(pino.transport({ target: "pino/file", options: { destination: "/dev/null" } }))`
Rest is the same as in `basic-bench.js`, all goes to `/dev/null`

### /dev/null is messing up the benchmark?
[...]

PR #1584, opened 2022-11-06
docs: add 'best performance for basic stdout logging' to help.md

janeklb:
My attempt to capture @mcollina's reply: https://github.com/pinojs/pino/issues/1491#issuecomment-1180129193 in some doc form

### DIFF
--- a/docs/help.md
+++ b/docs/help.md
@@ -300,5 +300,5 @@ default configuration:
 const log = require('pino')();
 ```
 
-You should only have to configure custom transports if you have broader logging
-requirements.
+You should only have to configure custom transports or other settings
+if you have broader logging requirements.
--- a/docs/help.md
+++ b/docs/help.md
@@ -12,7 +12,7 @@
 * [Unicode and Windows terminal](#windows)
 * [Mapping Pino Log Levels to Google Cloud Logging (Stackdriver) Severity Levels](#stackdriver)
 * [Avoid Message Conflict](#avoid-message-conflict)
-* [Best performance for basic stdout logging](#best-performance-for-stdout)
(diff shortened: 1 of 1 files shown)


docs/help.md
```diff
@@ -288,3 +289,16 @@ log.info({ msg: 'mapped to originalMsg' }, 'a message')
 // {"level":30,"time":1596313323106,"pid":63739,"hostname":"foo","msg":"no original message"}
 // {"level":30,"time":1596313323107,"pid":63739,"hostname":"foo","msg":"a message","originalMsg":"mapped to originalMsg"}
 ```
+
+<a id="best-performance-for-stdout"></a>
+## Best performance for logging to `stdout`
+
+The best performance for logging directly to stdout is achieved by using the
```
mcollina, line 296:
```suggestion
The best performance for logging directly to stdout is _usually_ achieved by using the
```

docs/help.md
```diff
@@ -288,3 +289,16 @@ log.info({ msg: 'mapped to originalMsg' }, 'a message')
...
+default configuration:
+
+```js
+const log = require('pino')();
+```
+
+You should only have to configure custom transports if you have broader logging
+requirements.
```
mcollina, line 304:
```suggestion
You should only have to configure custom transports or other settings
if you have broader logging requirements.
```

docs/help.md
janeklb: @mcollina just for the sake of curiosity could you elaborate on when the default configuration would not yield the best performance for logging directly to stdout?

mcollina (approved):
lgtm

PR #1588, opened 2022-11-10
chore: remove test directory & non-existent files from dist

ydcjeff:
Users don't need test in the final package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,11 +12,7 @@
     "pino.d.ts",
     "bin.js",
     "browser.js",
-    "pretty.js",
-    "usage.txt",
-    "test",
     "docs",
-    "example.js",
     "lib"
   ],
   "scripts": {


jsumners:
I know we had a long discussion in https://github.com/pinojs/pino/issues/136 and https://github.com/pinojs/pino/pull/137 about this topic, but I think we should readdress it. My statements in https://github.com/fastify/skeleton/issues/42#issuecomment-1208090783 are being proved here. We routinely ignore `(package.json).files` and it clearly gets out-of-date. I'd rather we:

1. have a simple `.gitignore`
2. _remove_ `(package.json).files`
3. _do not_ add a `.npmignore` file

@mcollina @davidmarkclements what are your thoughts?

ydcjeff:
If we are going to have only `.gitignore`, test directory will still be included in the final distribution.

jsumners:
> If we are going to have only `.gitignore`, test directory will still be included in the final distribution.

Correct. We are fine with that. When we are working on projects and need to dig into why a dependency isn't working as expected, or just need to understand the dependency better, we will read the included test files.

ydcjeff:
Yes, I agree with that, but I usually read the source on GitHub, so test files are unnecessary at least for me.

mcollina:
Those files are there because we want them there.

PR #1590, opened 2022-11-13
Add missing type for disabled in Logger options

moshie:
Add missing type for disabled under the browser logger options.


moshie:
Closing because #1590 exists

PR #1630, opened 2023-01-23
fix: badges

xtexChooser:
github.com/badges/shields#8671


mcollina (approved):
lgtm

jsumners (approved):
Thank you. We were unaware of this change.

Issue #544, opened 2018-11-15
pino prefix string to each log line


eran10:
Hi,
is there any option to add some const string to pino logger so for every log line we use "info" for example so i may add some prefix string that will be added ?

e.g
pino.info(`${my prefix} some log line`);
pino.info(`${my prefix} some other log line`);

mcollina:
Thanks for reporting! This is currently not possible.

davidmarkclements:
I'd suggest creating a little library for this, you could call it `pino-prefix` or some such,
the following should get you started: 

```js
function pinoPrefix(prefix, pinoInstance) {
  const { levels } = pinoInstance
  const { values } = levels
  Object.keys(values).forEach((level) => {
     const LOG = pinoInstance[level].bind(pinoInstance)
     pinoInstance[level] = (msg, ...args) => { 
      return LOG(typeof msg === 'string' ? prefix + msg : msg, ...args)
     }
  })
  return pinoInstance
}
```

In terms of putting it in core, we're happy to review PR's but it would need to have the following properties

1. zero overhead when not in use
2. very low performance impact when in use
3. clearly defined parameters of what would happen in the case of an object logging
4. documentation
5. tests (100%)
6. benchmarks

PR #1635, opened 2023-01-31
Allow message prefix - #544

rluvaton:
Allowing message prefix

fixing:
- #544

This PR contains:

- [x] Fix
- [x] Tests
- [x] Updated docs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -388,7 +388,7 @@ See `errorKey` option to change `err` namespace.
 
 Default: `undefined`
 
-The `msgPrefix` property allows you to specify a prefix for every message of the logger and it's children.
+The `msgPrefix` property allows you to specify a prefix for every message of the logger and its children.
 
 ```js
 const logger = pino({
--- a/docs/api.md
+++ b/docs/api.md
@@ -384,6 +384,24 @@ See `errorKey` option to change `err` namespace.
 
 * See [pino.stdSerializers](#pino-stdserializers)
 
+#### `msgPrefix` (String)
+
+Default: `undefined`
+
+The `msgPrefix` property allows you to specify a prefix for every message of the logger and it's children.
+
+```js
+const logger = pino({
+  msgPrefix: '[HTTP] '
+})
+logger.info('got new request!')
+// >  [HTTP] got new request!
+
+const child = logger.child({})
+child.info('User authenticated!')
+// >  [HTTP] User authenticated!
+```
+
 #### `base` (Object)
 
 Default: `{pid: process.pid, hostname: os.hostname}`
@@ -844,18 +862,21 @@ child.debug('debug!') // will log as the `level` property set the level to debug
 
 ##### `options.msgPrefix` (String)
 
+Default: `undefined`
+
 The `msgPrefix` property allows you to specify a prefix for every message of the child logger.
 By default, the parent prefix is inherited.
 If the parent already has a prefix, the prefix of the parent and then the child will be displayed.
 
 ```js
-const logger = pino()
-const child = logger.child({avengers: 'assemble'}, {msgPrefix: '[HTTP] '})
-child.info('got new request!')
+const logger = pino({
+  msgPrefix: '[HTTP] '
+})
+logger.info('got new request!')
 // >  [HTTP] got new request!
 
-const grandchild = child.child({aim: 'for the head'}, {msgPrefix: '[Proxy] '})
-child.debug('message proxied!')
+const child = logger.child({avengers: 'assemble'}, {msgPrefix: '[Proxy] '})
+child.info('message proxied!')
 // >  [HTTP] [Proxy] message proxied!
 ```
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -842,6 +842,23 @@ const child = logger.child({foo: 'bar'}, {level: 'debug'})
 child.debug('debug!') // will log as the `level` property set the level to debug
 ```
 
+##### `options.msgPrefix` (String)
+
+The `msgPrefix` property allows you to specify a prefix for every message of the child logger.
+By default, the parent prefix is inherited.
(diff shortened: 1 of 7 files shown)


rluvaton:
Could someone please approve running the workflows so I can see the performance change :)

rluvaton:
Could someone please stop this job:
https://github.com/pinojs/pino/actions/runs/4049806965/jobs/6977390174

rluvaton:
@Fdawgs Could you please me out and allow running the workflows 😬

Fdawgs:
Have reran it @rluvaton

mcollina:
The benchmarks looks ok to me. Could you add some tests for the fix you are making?

rluvaton:
yeah sure, to make sure this fix is ok for you, I'll try to add it + docs later this day

lib/tools.js
```diff
@@ -52,9 +53,16 @@ function genLog (level, hook) {
         msg = n.shift()
         formatParams = n
       }
+      if (typeof this[msgPrefixSym] === 'string' && msg !== undefined && msg !== null) {
+        msg = this[msgPrefixSym] + msg
```
jsumners, line 57: I think this is one of the cases where a coercive check is preferred:

```suggestion
      if (typeof this[msgPrefixSym] === 'string' && msg) {
        msg = this[msgPrefixSym] + msg
```

lib/tools.js
```diff
@@ -52,9 +53,16 @@ function genLog (level, hook) {
...
+      }
       this[writeSym](o, format(msg, formatParams, this[formatOptsSym]), level)
     } else {
-      this[writeSym](null, format(o === undefined ? n.shift() : o, n, this[formatOptsSym]), level)
+      let msg = o === undefined ? n.shift() : o
+      if (typeof this[msgPrefixSym] === 'string' && msg !== undefined && msg !== null) {
+        msg = this[msgPrefixSym] + msg
```
jsumners, line 63:
```suggestion
      if (typeof this[msgPrefixSym] === 'string' && msg) {
        msg = this[msgPrefixSym] + msg
```

lib/tools.js
rluvaton: thanks, done

lib/tools.js
rluvaton: thanks, done

rluvaton:
This PR is now ready for review @jsumners

mcollina (approved):
lgtm

rluvaton:
let's merge it?

mcollina:
let's wait for @jsumners

docs/api.md
```diff
@@ -842,6 +842,23 @@ const child = logger.child({foo: 'bar'}, {level: 'debug'})
 child.debug('debug!') // will log as the `level` property set the level to debug
 ```
 
+##### `options.msgPrefix` (String)
+
+The `msgPrefix` property allows you to specify a prefix for every message of the child logger.
+By default, the parent prefix is inherited.
+If the parent already has a prefix, the prefix of the parent and then the child will be displayed.
+
```
jsumners, line 850: I think this is also a root logger option, correct? We should have documentation that explains the option for the root logger. I think we are also missing a test that covers the root logger.

rluvaton: no, it's child logger option

rluvaton: @jsumners Added support for the root logger now

docs/api.md
```diff
@@ -384,6 +384,24 @@ See `errorKey` option to change `err` namespace.
 
 * See [pino.stdSerializers](#pino-stdserializers)
 
+#### `msgPrefix` (String)
+
+Default: `undefined`
+
+The `msgPrefix` property allows you to specify a prefix for every message of the logger and it's children.
```
jsumners, line 391:
```suggestion
The `msgPrefix` property allows you to specify a prefix for every message of the logger and its children.
```

jsumners (approved):
Looks good to me.

rluvaton:
I think the failed test is flaky, had it before... in the CI but couldn't reproduce on ubuntu, would love a rerun

jsumners:
I don't see any errors with the CI run. But the benchmarks may not be acceptable. @mcollina can you run them on your dedicated system?

rluvaton:
They changed a lot between:
504d1f1bdc60ba492245d6553a160cd84ed4e9db

to

f8067843fdc2b11bfcd3ad6576ae8c88d94425a1

and I only changed this:
[`a94a9da` (#1635)](https://github.com/pinojs/pino/pull/1635/commits/a94a9da3166672d2857650c7d2dc0f1d4153f12f#diff-7f88797c7db38a6af7a3283e3af30f2cdd3c2489bd9c85a7a25b32ec2485f724R56-R62)

jsumners:
> They changed a lot between: [504d1f1](https://github.com/pinojs/pino/commit/504d1f1bdc60ba492245d6553a160cd84ed4e9db)
> 
> to
> 
> [f806784](https://github.com/pinojs/pino/commit/f8067843fdc2b11bfcd3ad6576ae8c88d94425a1)
> 
> and I only changed this: [`a94a9da` (#1635)](https://github.com/pinojs/pino/pull/1635/commits/a94a9da3166672d2857650c7d2dc0f1d4153f12f#diff-7f88797c7db38a6af7a3283e3af30f2cdd3c2489bd9c85a7a25b32ec2485f724R56-R62)

Then the coercive check may not be acceptable. Revert the change and add a note like:

```js
// We do not use a coercive check for `msg` as it is
// measurably slower than the explicit checks.
```

rluvaton:
Reverted but it doesn't make any sense, this check has never met as we don't run benchmarks with log-prefix

rluvaton:
it's ready for merge :)

rluvaton:
@jsumners can this be merged?

PR #1636, opened 2023-01-31
docs(readme): replace shields with native badge

Fdawgs:


jsumners (commented):
If we are going to have multiple of these badges (and I find them rather useless altogether), then I would rather they all be consistent. Hence the usage of the shields variant for all badges. 

I am 👎 on this.

Fdawgs:
No worries, will just leave them as is then.

PR #1643, opened 2023-02-09
Add script to bump version

jsumners:
Since we are saddled with maintain the version number in code, I propose this script to make releases easier. All three of the following are valid:

```sh
$ ./inc-version.sh major
$ ./inc-version.sh minor
$ ./inc-version.sh patch
```

By default, the following will bump for a patch release:

```sh
$ ./inc-version.sh
```

### DIFF
--- a/inc-version.sh
+++ b/inc-version.sh
@@ -1,0 +1,42 @@
+#!/bin/bash 
+
+set -e
+
+PATH=./node_modules/.bin:${PATH}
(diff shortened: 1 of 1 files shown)


mcollina:
We have https://github.com/pinojs/pino/blob/master/build/sync-version.js to sync the version.

mcollina:
> I don't know how that happens. 

Manually. I launch that command. This script is good.

mcollina (approved):
lgtm

PR #1644, opened 2023-02-12
Make sure Pino is run on the right minimum required Node version

pH-7:
Lately, I had to install Pino on an old Node application. The Node version is 12. However, Pino couldn't be run as it uses `on-exit-leak-free` where `FinalizationRegistry` is being used which is only available since Node v14.

I was actually getting the following error coming from [on-exit-leak-free](https://github.com/pinojs/pino/blob/master/package.json#L114) since Pino is using it as a dependency.

```
ReferenceError: FinalizationRegistry is not defined
    at Object.<anonymous> (node_modules/on-exit-leak-free/index.js:11:18)
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,5 +123,8 @@
   },
   "tsd": {
     "directory": "test/types"
+  },
+  "engines": {
+    "node": ">=14.0.0"
   }
 }


jsumners (commented):
Thank you for the contribution. However, the engines block is not actually used by Node nor npm. Adding it will only add maintenance burden. 

I am 👎

kibertoad:
https://link.medium.com/Gl2Li0hfmxb
I am +1 on this change.

jsumners:
> https://itnext.io/node-engines-helping-developers-everywhere-avoid-phantom-bugs-2eef519604b2 I am +1 on this change.

package json | npm Docs - Vivaldi 2023-02-12 08-31-02

kibertoad:
@jsumners warnings are so much better than silently failing for unknown reason. jelps people to address the issue

jsumners:
> @jsumners warnings are so much better than silently failing for unknown reason. helps people to address the issue

In my experience, package installs generate so many useless warnings that they are completely ignored. Unless the package managers are going to enforce checking of this value with a failure by default, then it is pointless in my opinion.

The issue at hand is about adding Pino to an application. That is, Pino is a direct dependency of the application, and I am willing to be this is the most common scenario. We are very clear about what versions we support in our LTS doc (https://github.com/pinojs/pino/blob/ba5ab33bfa9738b6ccd96f2647e08edbcf9a6358/docs/lts.md). I think that is completely sufficient notice for people adding a direct dependency.

Thus, I stand by my argument that adding the engines block is nothing but a maintenance burden.

mcollina:
I think this field is mostly irrelevant and it does not cust us much to keep it up to date.

mcollina (approved):
lgtm

Fdawgs:
We did remove the `engines` fields from all @fastify repos to reduce maintenance burden as well prior to release of v4, and the node versions supported are [documented in Fastify's LTS schedule](https://github.com/fastify/fastify/blob/main/docs/Reference/LTS.md#schedule).

Pino has a similar [LTS schedule](https://github.com/pinojs/pino/blob/master/docs/lts.md#schedule), so this just seems like someone missing the docs?

mcollina:
Yes.

Issue #1617, opened 2023-01-04
Multistream resets logLevel to 30, logging trace and debug is not possible


TheAnimeGuru:
Hi,
I have encountered an issue with logging using a pino multistream.
Using a multistream, somehow disables the info and trace method, as the minimum value is set to 30 (corresponding to info). Any pino options and minimum Level Option in the Stream are discarded. Creating individual loggers without a Multistream, works fine.

Reference as code:
```ts
import pino from 'pino'
import pretty from 'pino-pretty'
import fs from 'fs'

const logDir = `logs/bot`

if (!fs.existsSync(logDir)) {
    fs.mkdirSync(logDir)
}

const logger = pino({ level: 'trace' }, pino.multistream([
    pretty({
        destination: pino.destination(`$logDir}/logger.log`),
        colorize: false,
        levelFirst: true,
        translateTime: 'dd.mm.yyyy h:MM:ss TT',
        ignore: 'pid,hostname',
    }),
    pretty({
        colorize: true,
        levelFirst: true,
        translateTime: 'dd.mm.yyyy h:MM:ss TT',
        ignore: 'pid,hostname',
    })
])
)

const message = 'This is a test'
//These will work
logger.info(message)
logger.warn(message)
logger.error(message)
logger.fatal(message)

// These will not work
logger.trace(message)
logger.debug(message)

const loggerWithoutMultistream = pino({ level: 'trace' }, pretty({
    destination: pino.destination(`${logDir}/logger.log`),
    colorize: false,
    levelFirst: true,
    translateTime: 'dd.mm.yyyy h:MM:ss TT',
    ignore: 'pid,hostname',
}))

// These will now work
loggerWithoutMultistream.trace(message)
loggerWithoutMultistream.debug(message)
```

When printing the object, the Multistream seems to edit minLevel in each individual Stream to 30, and set the pino.levelVal to 30 as well.

TheAnimeGuru:
Documentation says to set level on the lowest level, as done in the example

jsumners:
Have you tried it without the `pretty` stream?

TheAnimeGuru:
Yes and they also don't work. See:
```ts
import pino from 'pino'

const logger = pino({ level: 'trace' }, pino.multistream([
    { stream: process.stdout }
])
)

const message = 'TEST'
//These will work
logger.info(message)
logger.warn(message)
logger.error(message)
logger.fatal(message)

// These will not work
logger.trace(message)
logger.debug(message)

[...]

PR #1649, opened 2023-02-17
fix(docs): update multistream example and description

Cheprer:
Fix for clarifying the usage of multistream as I wasn't the only one https://github.com/pinojs/pino/issues/1617.


mcollina (approved):
lgtm thx!!!

Issue #1650, opened 2023-02-18
err errors are creating duplicated msg properties


leon-volq:
I'm using `"pino": "8.10.0"`

This script will output two logs with two "msg" properties each

```js
import pino from 'pino'

const log = pino()

log.info({ msg: 'hello', err: new Error('world') })

log.info({ err: new Error('world'), msg: 'hello' })
```
Log one
```json
{
    "level": 30,
    "time": 1676678480988,
    "pid": 17643,
    "hostname": "leons-matebook",
    "msg": "hello",
    "err": {
        "type": "Error",
        "message": "world",
        "stack": "Error: world\n    at file:///home/leon/Documents/martianbitcoins/game-servers/core-game/api/test-pino.js:5:31\n    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)\n    at async Promise.all (index 0)\n    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)\n    at async loadESM (node:internal/process/esm_loader:91:5)\n    at async handleMainPromise (node:internal/modules/run_main:65:12)"
    },
    "msg": "world"
}
```

Log two
```json
{
    "level": 30,
    "time": 1676678480989,
    "pid": 17643,
    "hostname": "leons-matebook",
    "err": {
        "type": "Error",
        "message": "world",
        "stack": "Error: world\n    at file:///home/leon/Documents/martianbitcoins/game-servers/core-game/api/test-pino.js:7:17\n    at ModuleJob.run (node:internal/modules/esm/module_job:193:25)\n    at async Promise.all (index 0)\n    at async ESMLoader.import (node:internal/modules/esm/loader:530:24)\n    at async loadESM (node:internal/process/esm_loader:91:5)\n    at async handleMainPromise (node:internal/modules/run_main:65:12)"
    },
    "msg": "hello",
    "msg": "world"
}
```

And then when parsing the JSON only the last "msg" value will remain.

Is a bit confusing since I had to start using "notice" key instead of "msg" key when logging errors, and I have to keep that in mind or my "msg" will be "eaten up"

Is this intended?
Is it documented and I missed it? 
If this is intended what is your recommended approach? 

Thank you for the great library! :heart:

mcollina:
I think this is an unwanted consequence of this line: https://github.com/pinojs/pino/blob/eafbc906a0cb065eb2da05edaebe144d838dcab2/lib/proto.js#L190-L193

I think we should also check if the object has a `msg` property.

[...]

PR #1654, opened 2023-02-22
fix msg precendence when sending a merging object with msg and err

leon-volq:
fix: https://github.com/pinojs/pino/issues/1650

### DIFF
--- a/lib/proto.js
+++ b/lib/proto.js
@@ -193,7 +193,7 @@ function write (_obj, msg, num) {
     }
   } else {
     obj = _obj
-    if (msg === undefined && _obj[errorKey]) {
+    if (msg === undefined && _obj.msg === undefined && _obj[errorKey]) {
       msg = _obj[errorKey].message
     }
   }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

lib/proto.js
```diff
@@ -193,7 +193,7 @@ function write (_obj, msg, num) {
     }
   } else {
     obj = _obj
-    if (msg === undefined && _obj[errorKey]) {
+    if (msg === undefined && _obj.msg === undefined && _obj[errorKey]) {
```
joelmukuthu, line 196: Hi all, should this be `_obj[messageKey]` (instead of `_obj.msg`) since that `msg` key is configurable?

joelmukuthu: Done: https://github.com/pinojs/pino/pull/1746. I did two separate `git push`es to try and trigger the failing test on CI but it didn't work as approval is needed. Hopefully you'll be able to trigger CI on the first commit!

Issue #1570, opened 2022-10-11
Add midnight-smoker to do some basic integration testing


mcollina:
Check out https://www.npmjs.com/package/midnight-smoker.

The goal is to catch configuration and dependency errors, e.g. a needed dependency wrongly classified as a devDepedency.

mcollina:
go for it!

Nabhag8848:
Hey @mcollina, while running midnight-smoker against pre-publish lifecycle script. i have found that `tap` module is install as <kbd>devDependency</kbd> instead of <kbd>dependency</kbd> which is giving following `Error: cannot find module tap`.

### When tap is install as devDependency: 
![Screenshot from 2023-01-06 01-21-52](https://user-images.githubusercontent.com/65061890/210872162-c6f1170f-4215-4fb0-9c9b-96abd8ebd13c.png)

### When tap is install as dependency: 
![Screenshot from 2023-01-06 01-19-41](https://user-images.githubusercontent.com/65061890/210872185-627b7b02-292c-4280-9e5e-a2d25e59626f.png)

- Thus, will need changes to install `tap` as dependency. currently tap has latest release of <kbd>v16.3.2</kbd> and we are currently having <kbd>v16.2.0.</kbd> should i go about installing the latest release or the version which already exist?

jsumners:
`tap` is and always should be a development dependency.

dancastillo:
hey @Nabhag8848 are you still working on this? If not can I take it up

Nabhag8848:
> hey @Nabhag8848 are you still working on this? If not can I take it up

Hey, Go ahead. I missed this and its been too long.

dancastillo:
@mcollina `midnight-smoker` needs [npm v7.0.0](https://github.com/boneskull/midnight-smoker#installation) in github actions CI check node 14 installs npm v6.14.18.
Would a github action be better? [resource](https://github.com/boneskull/midnight-smoker#resources)

jsumners:
I don't understand how they can claim Node 14 support when it doesn't ship with npm@7.

mcollina:
A separate github action is better.

PR #1663, opened 2023-03-03
feat: integrate midnight smoker in ci

couragecowardlydog:
Fix: https://github.com/pinojs/pino/issues/1570

@mcollina  @jsumners , please review and provide your comments 

Added a copy of example/basic.js in root as smoke.js


package.json
```diff
@@ -25,10 +23,11 @@
     "lint": "eslint .",
     "prepublishOnly": "tap --no-check-coverage test/internals/version.test.js",
     "test": "npm run lint && npm run transpile && tap --ts && jest test/jest && npm run test-types",
-    "test-ci": "npm run lint && npm run transpile && tap --ts --no-check-coverage --coverage-report=lcovonly && npm run test-types",
+    "test-ci": "npm run lint && npm run transpile && tap --ts --no-check-coverage --coverage-report=lcovonly && npm run test-types && npx midnight-smoker test-ci-smoke",
```
kibertoad, line 26: no additional dependency is needed?

package.json
```diff
@@ -12,11 +12,9 @@
     "pino.d.ts",
     "bin.js",
     "browser.js",
-    "pretty.js",
-    "usage.txt",
     "test",
     "docs",
-    "example.js",
+    "smoke.js",
```
kibertoad, line 17: does it need to be included in npm package?

couragecowardlydog: Yes , midnight-smoker execute the script inside the package

package.json
couragecowardlydog: Nope

mcollina: I would prefer to avoid `npx` and add `midnight-smoker` as a devdep.

couragecowardlydog: @mcollina, added `midnight-smoker` as dev-dep. Included separate script as well

mcollina (approved):
lgtm

kibertoad (approved):

couragecowardlydog:
> lgtm

@mcollina, the pipeline broke for some of the case,  npm version compatibility 


```
Minimum Node.js versions supported: ^14.17.0 || ^16.13.0 || >=18.0.0
Minimum npm version supported: v7.0.0
```

kibertoad:
@couragecowardlydog is it possible to run midnight smoker not through "exec" command?

couragecowardlydog:
> @couragecowardlydog is it possible to run midnight smoker not through "exec" command?

@kibertoad, can I change the script like this 
```
"smoker": "node ./node_modules/.bin/smoker test-ci-smoke",
```

is that fine ?

kibertoad:
I see no problem with that, if it ever stops working for whatever reason, CI will fail

couragecowardlydog:
> I see no problem with that, if it ever stops working for whatever reason, CI will fail

Fixed that, can you please approve the workflow to see if CI goes through

package.json
```diff
@@ -12,23 +12,23 @@
...
     "lib"
   ],
   "scripts": {
     "docs": "docsify serve",
     "browser-test": "airtap --local 8080 test/browser*test.js",
     "lint": "eslint .",
+    "smoker": "node ./node_modules/.bin/smoker test-ci-smoke",
```
mcollina, line 24:
```suggestion
    "smoker": "smoker test-ci-smoke",
```

this should work too

mcollina (changes requested):

Issue #1680, opened 2023-03-30
Type definitions indicate they are for an older version (7.0)


pushred:
The [lead comment](https://github.com/pinojs/pino/blob/2d5c522398731c1063a453295ed0f906b5ae54ff/pino.d.ts#LL1) in the TS definitions file currently indicates they are for v7.0. While inspecting the types in my project this prompted some sanity checking to verify my version.

Not sure if the comment is worth retaining if it's not being maintained with version updates?

mcollina:
We forgot to remove that line. Would you like to send a PR?

PR #1693, opened 2023-04-14
remove type definition for pino 7.0 comment

tarunrajput:
remove type definition for pino 7.0 comment as it's not being maintained with version updates
Related: #1680


mcollina (approved):
lgtm, thanks!

PR #1698, opened 2023-04-27
Docs: add favicon

WolfGangsta:
I like to set up a browser window for just my docs, all in pinned tabs on Chrome (i.e. just the icon showing). Since Pino didn't have one yet, I decided to see if I couldn't set one up myself.

Referencing the existing pino-tree.png is possible, but because the image isn't square, it looked squashed. I headed over to realfavicongenerator.net and put pino-tree.png through it to get some more standardized favicons.

These 3 files were generated (along with a lot of other special-case ones that I removed since I couldn't easily test them). The website also provided some code to add in, which only needed one tweak--the removal of a `/` before each `href`--to make GitHub pages support it.

How do they look? (For smaller icon sizes, there's also the option of using bilinear or some other scaling technique, which looks more pixelated. I went with the default, just because it was the default.)


jsumners (approved):
Looks good to me.

Issue #1690, opened 2023-04-11
`exposes faux stdSerializers` test fail on node v19.9.0


rluvaton:
The tests fail on node v19.9.0 on master:

```
 FAIL  test/browser.test.js
 ✖ pino.stdSerializers.errWithCause

  test: test/browser.test.js
  operator: ok
  expected: true
  actual: undefined
  at: Test.<anonymous> (/Users/rluvaton/dev/tmp/pino/test/browser.test.js:101:5)
  stack: >-
    Error: pino.stdSerializers.errWithCause
        at Test.assert [as _assert] (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:312:48)
        at Test.bound [as _assert] (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:95:17)
        at Test.assert (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:431:7)
        at bound (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:95:17)
        at Test.<anonymous> (/Users/rluvaton/dev/tmp/pino/test/browser.test.js:101:5)
        at Test.bound [as _cb] (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:95:17)
        at Test.run (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:115:28)
        at Test.bound [as run] (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/test.js:95:17)
        at Immediate.next [as _onImmediate] (/Users/rluvaton/dev/tmp/pino/node_modules/tape/lib/results.js:157:7)
        at process.processImmediate (node:internal/timers:478:21)
```

reproduce:

```bash
nvm install 19.9.0
nvm use 19.9.0
npm i
npm test
```

this cause the CITGM tests in node to fail.

related PR:
- https://github.com/nodejs/node/pull/47413

rluvaton:
So you say it's https://github.com/pinojs/pino-std-serializers issue?

jsumners:
> So you say it's https://github.com/pinojs/pino-std-serializers issue?

No. It's an issue here.

jsumners:
We need someone to dig into the timeouts happening in our test suite https://github.com/pinojs/pino/pull/1691#issuecomment-1504009718

PR #1691, opened 2023-04-11
Fix browser serializers

jsumners:
Resolves issue #1690

PR #1701, opened 2023-05-02
Expose errWithCause to browser API & add Node v20

mcollina:
Fixes https://github.com/pinojs/pino/issues/1690


Fdawgs:
Worth dropping Node 19 from the CI as well?

jsumners (approved):
I don't know why #1691 won't pass tests and this one will, but 🤷‍♂️.

Looks good to me.

PR #1707, opened 2023-05-03
Support custom levels on browser

### LINKED ISSUE #1706: CustomLevels not working on browser
The following example fails on browser

```js
pino({ customLevels: { ok: 31 } }).ok('foo')
```

Looking at the browser implementation, it doesn't add any of the `customLevels` to the logger 🤔
Looks like its a simple PR to make it work, I can work on that.

matheusvellone:
Closes #1706 

~Tests aren't passing locally yet. I'm looking at them~ Fixed

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -48,11 +48,13 @@ function pino (opts) {
     opts.browser.serialize.indexOf('!stdSerializers.err') > -1
   ) stdErrSerialize = false
 
-  const levels = ['error', 'fatal', 'warn', 'info', 'debug', 'trace']
+  const customLevels = Object.keys(opts.customLevels || {})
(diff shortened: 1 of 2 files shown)


browser.js
```diff
@@ -74,7 +76,7 @@ function pino (opts) {
     levels,
     timestamp: getTimeFunction(opts)
   }
-  logger.levels = pino.levels
+  logger.levels = getLevels(opts)
```
matheusvellone, line 79: I did it this way to avoid mutating `pino.levels`, which is public, to avoid a possible breaking change.

Or should I mutate `pino.levels` to add custom levels there anyway?

mcollina (approved):
lgtm, this is ok

browser.js
mcollina: this is correct

PR #1709, opened 2023-05-04
feat: pass logger/child logger as param to mixin

mmarchini:
Passing the logger or child logger as a parameter to `mixin` allows users to set logger-specific context in the logger object that can be used by `mixin` to enrich the context to be added to the resulting JSON.

One example use case for this is avoiding the "duplicate keys" caveat that comes with child loggers. If the user wants to make a known key "mergeable", they could add a custom function that concatenates values to the key and then use `mixin` to pass that merged value to the context. For example:

```js
instance = pino({
  mixin(obj, num, logger) {
    return {
      "tags": logger.tags
    }
  }
})
instance.tags = {}

instance.addTag = function (key, value) {
  child.tags[key] = value
}

function createChild (logger, ...context) {
  const child = logger.child(...context)
  child.tags = { ...logger.tags }
  child.addTag = function (key, value) {
    child.tags[key] = value
  }
  return child
}

instance.addTag('foo', 1)
child.addTag('bar', 2)
instance.info('this will only have `foo: 1`')
child.info('this will have both `foo: 1` and `bar: 2`')
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -121,6 +121,8 @@ Default: `undefined`
 
 If provided, the `mixin` function is called each time one of the active
 logging methods is called. The first parameter is the value `mergeObject` or an empty object. The second parameter is the log level number.
+The third parameter is the logger or child logger itself, which can be used to
+retrieve logger-specific context from within the `mixin` function.
 The function must synchronously return an object. The properties of the returned object will be added to the
 logged JSON.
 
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Good idea! Can you update the docs too ?

mmarchini:
done!

mcollina (approved):
lgtm

Issue #960, opened 2021-02-05
Childlogger Loglevel not respected in the browser


KeKs0r:
Hi,

I am using pino in the browser, since I have some parts of my application that are used in the browser as well as on node.

When using pino in the browser, it seems that different log levels for Child loggers are not working: 
Here is my root logger
```
const logger = pino({
    level: 'info',ser: {
        asObject: false,
        serialize: false,

        transmit: {
            level: 'error',
            send: function (level, logEvent) {
             //... Send error to sentry
            },
        },
    },
})

const childLogger = logger.child({name:'child', level:'debug'})

logger.debug("Root Logger Debug") // not logged
logger.info("Root Info") // logged
childLogger.debug("Child Debug") // Not logged, but I think it should
childLogger.info("Child Info") // logged
```

The output looks like this
![image](https://user-images.githubusercontent.com/2970814/107027132-015f7100-67ac-11eb-9473-8dd9ec9c56ca.png)

It seems the level for the child logger, is just taken as additional property, instead of changing the level of the child logger.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

jakobrosenberg:
I'm having a similar issue. 

**debug level has no effect**
```javascript
this.log = pino({ level: 'debug' })

this.log.debug('im a debug')
this.log.info('im info') //logged
this.log.warn('im a warning') //logged
this.log.error('im an error') //logged
this.log.fatal('im fatal') //logged
```

**other levels work**
```
this.log = pino({ level: 'error' })

this.log.debug('im a debug')
this.log.info('im info')
this.log.warn('im a warning')
this.log.error('im an error') // logged
this.log.fatal('im fatal') // logged
```

0x15e:
I'm having issues with this as well. Working with 7.0.0-rc2. No amount of passing parameters or setting properties ever gets the child logger to use the lower level.

E.g.,
```
const childLogger = parentLogger.child({'foo': 'bar'}, {level: 'debug'});
// childLogger.level == 'info' - same as parentLogger.level

childLogger.level = 'debug';
// childLogger.level == 'debug' but still doesn't log anything to the browser console
```

I have verified that the level filter in the browser is set to show debug.

NicoVogel:
I have the same problem in the browser.
Here a short reproducible setup: https://stackblitz.com/edit/typescript-5ka82t?devToolsHeight=33&file=index.ts

[...]

PR #1725, opened 2023-06-02
fix: browser respect child logger log level 

NicoVogel:
This PR addresses #960.

> not yet completed, only a draft

# Goal of this PR:

Child level determines which logs of the child are logged.

# Implementation before this PR

- the `prototype` of `logger` contains the actual log functions which should be used
- the `prototype` log functions are assigned to the `logger` whenever the `setLevel` function is called
- for log functions which level  is to low, `noop` is assigned instead

> problem: only the main `logger` actually changes, even if the level of a child is changed

# What this PR changes

- introduce a property which contains all the base logger functions (only the root logger has it) -> `Symbol('pino.logFuncs')`
  - the base log functions need to be callable independently by the child loggers and this property allows to access them all the time
- introduce a property which contains the hierarchy of the logger (pointing to the parent) -> `Symbol('pino.hierarchy')`
  - whenever a child logs, it needs all bindings as they are prepend to the log arguments.
    before this PR, it was handled via the prototype chain.
- move the responsibility of prepending the bindings into the `set` functions (for changing the log level)
  - as the `set` function is mainly responsible for defining the actual log functions that are usable, it felt like a logical step to also include the binding in it. 
- add a `bindings` property to each child that stores the bindings of the child.
  - this is needed to walk up the hierarchy, get all bindings and then use it to prepend the log arguments.
  - there are restrictions to this property: it is only considered once, when calling the `set` function. 
    so, reassigning it only takes affect if the `level` is changed and then it *only* is updated for this one logger.

# Insights I want to share

There is a specific order in which a log is processed:

1. the outer log function is called with whatever should be logged
2. the bindings are prepended to the arguments
3. the `LOG` function is used to transform the arguments to the desired structure (can be found in `createWrap`)
4. the actual logging function is called with the transformed arguments

Before this PR, the prototype chain was used the ensure that the bindings are prepended one after another.
Now they are collected once and then prepended at once.

Another improvement that comes with this PR as well is, that now in case of `noop` the logger actually does nothing.
Before this PR, it always prepended the bindings and transformed the arguments no matter if the `noop` function was used or not.

> there is a valid case as well and that is if a `transmit` exists, but this is properly handled

Finally, the base log functions are now only once defined in the `setupBaseLogFunctions` instead of in every `set` call.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -16,6 +16,38 @@ const stdSerializers = {
   err: asErrValue,
   errWithCause: asErrValue
 }
+function levelToValue (level, logger) {
+  return level === 'silent'
+    ? Infinity
+    : logger.levels.values[level]
+}
+const baseLogFunctionSymbol = Symbol('pino.logFuncs')
+const hierarchySymbol = Symbol('pino.hierarchy')
+
+const logFallbackMap = {
+  error: 'log',
+  fatal: 'error',
+  warn: 'error',
+  info: 'log',
+  debug: 'log',
+  trace: 'log'
+}
+
+function appendChildLogger (parentLogger, childLogger) {
+  const newEntry = {
+    logger: childLogger,
+    parent: parentLogger[hierarchySymbol]
+  }
+  childLogger[hierarchySymbol] = newEntry
+}
+
+function setupBaseLogFunctions (logger, levels, proto) {
+  const logFunctions = {}
+  levels.forEach(level => {
+    logFunctions[level] = proto[level] ? proto[level] : (_console[level] || _console[logFallbackMap[level] ?? 'log'] || noop)
+  })
+  logger[baseLogFunctionSymbol] = logFunctions
+}
 
 function shouldSerialize (serialize, serializers) {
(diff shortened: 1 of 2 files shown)


NicoVogel:
I am still not sure if the approach I take actually goes into the right direction.
My guess is that there should be a better way of solving this with the prototype chain, but I have a hard time getting into this mind set, because I never needed to work with prototypes before.

Anyone has a guess here?

mcollina:
Yes, I like this approach.

jsumners (approved):

mcollina (approved):
lgtm

NicoVogel:
@mcollina I noticed that I used `feat` in the commit message instead of `fix`. Should I fix this or do you use squash commits anyways?

Furthermore, I am unsure how to proceed with the failing checks.
- the benchmark for `current` (master) failed, but I can't fix that in this PR
- 14 windows failed the installation of dependencies, but I did not touch the dependencies

As for windows 16 and 18, I will try to figure out why this fails

PR #1769, opened 2023-08-04
Make options and level optional for TransportTargetOptions

KonradSwierczynski:
Hey,
I saw that TypeScript requires `options` and `level`, while it seems I don't really need to provide them.
I've changed types in `pino.d.ts` so I don't need to do `@ts-ignore`.
Please take a look


mcollina:
Would you mind adding a test for this type change in https://github.com/pinojs/pino/tree/master/test/types? We use `tsd`

PR #1779, opened 2023-08-08
adding asString to the key

tzviki:
This fixes issue-1767. It can maybe be added as a feature? not sure, but happy to discuss further


mcollina (approved):
This will slow us down a bit but I don't think we can help it.

jsumners (approved):

PR #1783, opened 2023-08-12
fix: typescript types

### LINKED ISSUE #1782: TypeScript interface LogFn bugged (e.g. logger.info, logger.debug, etc.)
The normal usage for Pino is to use one argument, like this:

```ts
logger.info("foo"); // "foo" is logged.
```

If you try to use 2 arguments, Pino will use the second argument as a string interpolation argument. However, Pino will silently drop the second argument it if you don't have any `%` signs in your first argument:

```ts
logger.info("foo", "bar"); // "foo" is logged, but "bar" is dropped.
```

[The TypeScript definitions for Pino](https://github.com/pinojs/pino/blob/master/pino.d.ts) are therefore bugged, as the TypeScript compiler does not throw an error when you supply a second argument.

In order to fix this problem, I believe that the type definitions for the function need to be more specific.

Zamiell:
Fixes #1782

The PR causes an error in one of the existing tests:

image

It's line 312 in "pino.test-d.ts". I don't understand the error or what it is supposed to be testing. Can anyone shed some light on this?


Zamiell:
Oh, I think I understand the error, it's actually catching the runtime error that the code is intentionally performing, LOL.

Zamiell:
There is a "breaking" change in this PR (depending on how exactly you want to define breaking) such that you can no longer use the first generic type parameter as a means to validate the object that you are passing. Users should be directed to use the `satisfies` operator instead in the patch notes for the next release.

(`satisfies` was released as part of TypeScript 4.9 on November 15th, 2022. Now that it exists, it makes sense to validate your objects with `satisfies` like you would with any other TypeScript code and to not provide library-specific hacks to do it.)

mcollina (approved):
lgtm, good work!

Zamiell:
Oh, looks like the tests are failing because the version of TypeScript is too old to have the `satisfies` operator. Do you want me to do a separate PR that upgrades TypeScript?

kibertoad:
@Zamiell yes, please!

Zamiell:
Actually, on second thought, [the version of TypeScript is really recent](https://github.com/pinojs/pino/blob/master/package.json#L100).

Can someone click the button to re-run CI for me (since I'm a first time contributor)?

Zamiell:
Possibly related: Is there a particular reason that this repository does not have a package manager lock file? (e.g. "package-lock.json" or "yarn.lock" or "pnpm-lock.yaml")
I'm pretty sure that all JavaScript/TypeScript projects are supposed to have one.
Not having one might lead to bugs like in CI where it is running older versions of the dependencies, which might be what is happening here.

Zamiell:
From the CI logs:
```
 ✖  278:2   Cannot find name satisfies.     
```
But according to the package.json file at the root of the repository, the version of TypeScript is 5.1.3, which should have the satisfies operator, which was introduced in TypeScript version 4.9.0. Thus, I conclude that the CI is running on an old version of the node_modules, so something in this repo is broken with respect to caching, which is outside of the scope of my types PR.

Can a maintainer take a look at this please?

Zamiell:
@mcollina Can we work to get this merged? It's been sitting for a while.

mcollina:
I don't have time to investigate at the moment Happy to land with a green CI.

Zamiell:
@mcollina Can we get another maintainer to investigate? I'm fairly certain that the errors in CI are not related to my PR.

mcollina:
CI is passing on master, please rebase

Zamiell:
> CI is passing on master, please rebase

I'm aware of that, I'm talking about the CI errors inside of this pull request. Can we get another maintainer to investigate? This PR has been sitting here for over a month.

jsumners:
You need to rebase this PR.

Zamiell:
@jsumners I rebased this PR onto master just now. Let me know if anything additional is required.

mcollina:
CI is failing

Zamiell:
@mcollina Sorry, I think we are looping. Yes, CI is failing for the same reason that I posted in my reply on October 4th here: https://github.com/pinojs/pino/pull/1783#issuecomment-1747669708

This reply explains why the errors are not part of my PR and why I think that a maintainer needs to investigate: https://github.com/pinojs/pino/pull/1783#issuecomment-1677553078

To recap, in response to asking for a maintainer to look at the logs, it was requested that I rebase the PR on main, which I have now done, but CI still gives the same error, so we are back to square 1, and I would ask again that a maintainer please take a look at these logs, as they do not seem related to my PR!

jsumners:
What I can tell you is that `npm run test-ci` works locally on the main branch, but it fails on this branch.

Zamiell:
@jsumners It fails on this branch because I am using TypeScript's `satisfies` operator in my PR, which is not present in the codebase otherwise. As I explained [in a previous comment in this thread](https://github.com/pinojs/pino/pull/1783#issuecomment-1675746986), the `satisfies` operator was added in TypeScript version 4.9.0, and the version of TypeScript that is used by this package is 5.1.3, thus it should have the satisfies operator, thus the error is spurious and/or indicates that your CI is drunk.

kibertoad:
I can take a look on Monday

jsumners:
There isn't a cache on my local system. The _only_ way TS is even on my system is through `npm i` on this specific project.

jsumners:
I don't know what the "`satisfies` operator" is. But if it is something added in this pull request, and you are asserting that it is the cause of the problem, then it stands to reason that it is the fault of this pull request.

Zamiell:
This PR doesn't update TypeScript: please see [this reply that I posted 105 days ago](https://github.com/pinojs/pino/pull/1783#issuecomment-1677553078) which discusses this.

Zamiell:
Thanks. I updated TSD, can you click the button to run CI?

package.json
```diff
@@ -115,5 +115,6 @@
   },
   "tsd": {
     "directory": "test/types"
-  }
+  },
+  "packageManager": "yarn@1.22.19"
```
kibertoad, line 119: please don't add this

Zamiell: this line is necessary, or else installation fails on computers with any modern version of yarn installed

kibertoad: Feel free to add it locally, but don't commit it to the repository. We are not using yarn

Zamiell: more specifically, run `corepack enable`, and then clone this repo and run `yarn install` and then watch what happens

Zamiell: oh, nevermind it's been too long and im remembering now that this repo doesn't prescribe a package manager for some reason

kibertoad: that's way beyond the scope of this PR.
to my knowledge, none of fastify or pino dependencies have this field set. If there are good reasons to set it - please open an issue describing the problem and giving some references about this field.

Zamiell: sorry, i got confused because this repo doesn't use a package manager, so installation defaulted to yarn, which caused the issue. as an aside, why isn't there a lock file committed to git? that's non-idiomatic afaict

kibertoad: it's a library, not an application. package-lock is not applied transitively

Zamiell: From your message, you seem to be implying that it is not correct for JavaScript/TypeScript libraries to commit a package lock file. But is that true? [Sindre Sorhus seems to echo that opinion here](https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514). He specifically gives the justification of "Lockfiles are great for apps where you want a controlled reproducible environment, but for packages this doesn't make much sense."

But consider this very issue with the `tsd` dependency - we very much want exact reproducible builds when troubleshooting CI issues for our libraries! Furthermore, as far as I can tell, the ecosystem disagrees with this, and it is idiomatic to commit a lockfile for a library. For example:

- [TypeScript has a package-lock.json file](https://github.com/microsoft/TypeScript/blob/main/package-lock.json)
- [React has a yarn.lock file](https://github.com/facebook/react/blob/main/yarn.lock)
- [Angular has a yarn.lock file](https://github.com/angular/angular/blob/main/yarn.lock)
- [Redux has a yarn.lock file](https://github.com/reduxjs/redux/blob/master/yarn.lock)
- [Next.js has a pnpm-lock.yaml file](https://github.com/vercel/next.js/blob/canary/pnpm-lock.yaml)
- [Vue has a pnpm-lock.yaml file](https://github.com/vuejs/core/blob/main/pnpm-lock.yaml)
- [Moment.js has a package-lock.json file](https://github.com/moment/moment/blob/develop/package-lock.json)
- [JQuery has a package-lock.json file](https://github.com/jquery/jquery/blob/main/package-lock.json)
- [Nest.js has a package-lock.json file](https://github.com/nestjs/nest/blob/master/package-lock.json)
- [Express does not have a lock file](https://github.com/expressjs/express)
- [Astro has a pnpm-lock.yaml file](https://github.com/withastro/astro/blob/main/pnpm-lock.yaml)
- [Quik has a pnpm-lock.yaml file](https://github.com/BuilderIO/qwik/blob/main/pnpm-lock.yaml)
- [Remix has a yarn.lock file](https://github.com/remix-run/remix/blob/main/yarn.lock)
- [Axios has a package-lock.json file](https://github.com/axios/axios/blob/v1.x/package-lock.json)

But you mostly answered the original question, so I'm satisfied for now, thanks.

Zamiell:
> There are still two failures remaining which do look like broken types, though, could you please take a look?

those should be fixed now in my latest commit

Zamiell:
Looks like CI is green now, finally.

The fact that `tsd` uses its own version of TypeScript seems really weird. Naively I would expect it to be a peer dep and let the user bring their own version. But I'm seeing that it supports testing against multiple versions of TypeScript, so I guess that is the justification.

test/types/pino.test-d.ts
```diff
@@ -264,9 +265,9 @@ interface StrictShape {
     err?: unknown;
 }
 
-info<StrictShape>({
+info({
     activity: "Required property",
-});
+} satisfies StrictShape);
```
mcollina, line 270: Can you explain this change?

docs/typescript.md
```diff
@@ -0,0 +1,74 @@
+# Usage With TypeScript
```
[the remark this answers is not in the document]
Zamiell: done

beemdvp:
fyi these typing changes really should've been a major bump even if they're a "fix". This has broken an app in pipeline for production use. Guess we shouldnt really trust any package to respect semantic versioning but still

kibertoad:
@mcollina I think we should revert.

@beemdvp can you share the code example that got broken? we need to expand our testsuite, it seems.

beemdvp:
@kibertoad it was a minor thing but I'd rather deal with updating types in a separate story I'm working on. Previously we were allowed to log `unknown` (default type in catch block) types as well but as you can imagine, I had to find that out from this release and its a pain to change every single line in the same PR im working on at work

mcollina:
Given that typescript or DefinitelyTyped does not support semantic versioning, how could anybody expect types to stay stable?

kibertoad:
@mcollina Fair point.

Not being able to log unknown anymore sounds like a regression to me, as it's a very common and very useful use-case.

@beemdvp would you be open to send a fix PR for it?

mcollina:
> Not being able to log unknown anymore sounds like a regression to me, as it's a very common and very useful use-case.

Generically this seems like a bad idea. Basically you want to log something that you know nothing about, and expect everything to be safe. This seems a good case for a warning, because... who knows what's in that `unknown`?

arijnr:
Hi everyone.

I wanted to let you know that I also had problems during a deployment to production due to changes in this PR that are now available in version 8.17.0.

For now I will have to use the previous version.

kibertoad:
@mcollina realistically in 99.9% cases catch block will return an error. in 0.1% it'll be a string. pino can log both of them just fine.

obviosly people can log unknown in other cases too, but I'd say that it shouldn't be on pino to hand-hold against it

mcollina:
Agreed. However I don't think it's possible to support unknown while validating the interpolation arguments.

Zamiell:
> However I don't think it's possible to support unknown while validating the interpolation arguments.

To give more context here, currently the LogFn is typed as a bunch of overloads. Adding an overload to handle `unknown` at the top would match everything, making all the overloads below it useless (at least for one arg). And adding an overload for `unknown` at the bottom would match everything, resulting in 0 type errors. So in either case, the following code would not error, because it would be caught by the `unknown` overload:

```ts
logger.info("hello, %s"); // We want this to be a type error, but it matches: `(arg: unknown) => void`
```

> in 99.9% cases catch block will return an error. in 0.1% it'll be a string. pino can log both of them just fine.

You're right. In these cases, I recommend changing the following code:

```ts
try {
  dangerousFunc();
} catch (err) {
  logger.error(err)
}
```

To this:

```ts
try {
  dangerousFunc();
} catch (err) {
  logger.error(err as never)
}
```

Zamiell:
Building on my previous comment, remember that the whole point of `unknown` is that you are supposed to type-narrow it. So if you want to be absolutely safe, you can also do this:

```ts
/**
 * Helper function to narrow an unknown value to an object (i.e. a TypeScript record).
 *
 * Under the hood, this checks for `typeof variable === "object"`, `variable !== null`, and
 * `!Array.isArray(variable)`.
 */
export function isObject(
  variable: unknown,
): variable is Record<string, unknown> {
  return (
    typeof variable === "object" &&
    variable !== null &&
    !Array.isArray(variable)
  );
}

/** Helper function to type-narrow `unknown` to a loggable value. */
function logError(err: unknown) {
  if (typeof err === "string" || isObject(err) {
    logger.error(err);
  }
}

try {
  dangerousFunc();
} catch (err) {
  logError(err);
}
```

PR #1800, opened 2023-09-08
Skip writing a big file test in CI and CITGM

mcollina:
Apparently [CITGM](https://github.com/nodejs/citgm) has some problems in dealing with a big file on the temporary disk, so let's avoid it. Seems a good idea to do that also on our CI. 

Ref https://github.com/nodejs/citgm/pull/959


jsumners (approved):
Not sure how we guard against regressions if our CI isn't running tests, but 🤷‍♂️

PR #1813, opened 2023-09-21
export the LoggerExtras interface

ethanniser:
I have been using elysia-logger, which uses pino and have had some type issues: bogeychan/elysia-logger#4

the most recent issue is this:
image

which I found was solved by going into node modules and exporting the type in question

so this pr:

- exports the `LoggerExtras` interface

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -45,7 +45,7 @@ interface redactOptions {
     remove?: boolean;
 }
 
-interface LoggerExtras<Options = LoggerOptions> extends EventEmitter {
+export interface LoggerExtras<Options = LoggerOptions> extends EventEmitter {
     /**
      * Exposes the Pino package version. Also available on the exported pino function.
      */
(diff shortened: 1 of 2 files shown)


mcollina (commented):
Could you add a couple of assertions in our types tests using tsd? So we don't regress.

ethanniser:
sorry im a bit confused
what exactly do you want me to test for?

I was gonna add this test:

```ts
const log = pino();
expectAssignable<LoggerExtras>(log);
```

but this test already exists
```ts
const log = pino();
expectType<Logger>(log);
```

The only use of `LoggerExtras` in the entire codebase is as an intersection in the definition for `Logger`

So this existing test covers the same thing

---

I might just be missing something, happy to write some tests if you can explain a bit more what you are looking for
thanks

mcollina:
> So this existing test covers the same thing

If the behavior you want to change is _already tested_ and said test is passing, then the change is not needed and you need to implement something different. You are adding an `export`, so I would guess the test would need to `import` that new exported value.

ethanniser:
Got it, I have added the test I outlined earlier and all is passing.

Let me know if anything else is required, thank you.

mcollina (approved):

PR #1819, opened 2023-09-30
Disable FinalizationRegistry if node code coverage is active

mcollina:
See https://github.com/nodejs/node/issues/49344#issuecomment-1741776308 for more details, this has been the source of some of the flakyness we have been seeing.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -231,11 +231,15 @@ function hasBeenTampered (stream) {
   return stream.write !== stream.constructor.prototype.write
 }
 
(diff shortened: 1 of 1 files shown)


jsumners:
CI does not seem happy about it. Also, we need to detect code coverage in the actual library code? 🤢

mcollina:
There is literally no way around it until Node.js and V8 fixes this bug. If node code coverage is used together with a finalizationregistry... there is a high chance of a race condition that would make the a process not terminate.

Those tests that fails are not effected by this change, as the patched problem does not affect our test harness (we don't use C8).

jsumners (approved):

PR #1824, opened 2023-10-04
update flush function type

rluvaton:


mcollina:
@jsumners I did a git bisect and found this commit as a culprit: 
 [0d67bfa3d98d2d2076a3b1a9d9f6445a569c59a9](https://github.com/pinojs/pino-pretty/commit/0d67bfa3d98d2d2076a3b1a9d9f6445a569c59a9)

Does anything spot immediately?

mcollina:
Taking a look now!

mcollina:
https://github.com/pinojs/pino/pull/1825 fixes it

pino.d.ts
```diff
@@ -123,7 +123,7 @@ export interface LoggerExtras<Options = LoggerOptions> extends EventEmitter {
     /**
      * Flushes the content of the buffer when using pino.destination({ sync: false }).
      */
-    flush(): () => void;
+    flush(): void;
```
mcollina, line 126: can you add a test for this type change?

jsumners:
> @jsumners I did a git bisect and found this commit as a culprit: [0d67bfa3d98d2d2076a3b1a9d9f6445a569c59a9](https://github.com/pinojs/pino-pretty/commit/0d67bfa3d98d2d2076a3b1a9d9f6445a569c59a9)
> 
> Does anything spot immediately?

Looks like you've figured it out. We shouldn't rely on internals of other modules like that. If we need them, we should export them from the other module in some way.

jsumners:
Rebase and tests should pass.

rluvaton:
Would appreciate adding the`HACKTOBERFEST-ACCEPTED` label once merged :)

pino.d.ts
mcollina, line 126: Can you add a test for this type? We use tsd.

rluvaton: Oh, I did understand you meant that, np

mcollina (approved):
lgtm

PR #1831, opened 2023-10-17
fix: make transports.options and transports.level optional

krosenk729:
According to https://github.com/pinojs/pino/blob/HEAD/docs/transports.md#v7-transports, the options and level are non-required properties for a transport

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -244,8 +244,8 @@ declare namespace pino {
 
     interface TransportTargetOptions<TransportOptions = Record<string, any>> {
         target: string
-        options: TransportOptions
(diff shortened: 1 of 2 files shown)


mcollina (commented):
Thanks for opening a PR! Can you please add a unit test? We use tsd for types.

krosenk729:
thanks for the review @mcollina  - I added a test in there. let me know how it looks

krosenk729:
hey @mcollina + @kibertoad -- would appreciate a review on this (or help if I'm missing anything). many thanks!

mcollina (approved):
lgtm

PR #1834, opened 2023-10-23
Support Node.js v21

mcollina:


jsumners:
Why is 16 failing on both macOS and Windows?

Issue #1835, opened 2023-10-24
Logger options type


UndefinedBehaviour:
This code makes the TS compiler scream because because "LoggerOptions" has no "hello" key (the exact error is `Argument of type '{ hello: number; }' is not assignable to parameter of type 'LoggerOptions | DestinationStream | undefined'`). I think this is the correct behavior.
```
import pino from 'pino';

const log = pino({
 hello: 100
});
```

On the other hand this code doesn't seem to raise any error, i.e. if the object passed as a parameter contains at least one legit key no exception is raised. Is this intended behavior?
```
import pino from 'pino';

const log = pino({
 safe: true,
 hello: 100
});
```

UndefinedBehaviour:
fixed by https://github.com/pinojs/pino/pull/1858

PR #1843, opened 2023-11-01
fix: make types stricter

UndefinedBehaviour:
A solution to https://github.com/pinojs/pino/issues/1835.
Makes types stricter so that you can't create loggers with unknown options
```
pino({
    name: 'logger'
    hello: 'world'
});
```

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -31,13 +31,13 @@ type TimeFn = () => string;
 type MixinFn = (mergeObject: object, level: number) => object;
 type MixinMergeStrategyFn = (mergeObject: object, mixinObject: object) => object;
 
-type CustomLevelLogger<Options> = Options extends { customLevels: Record<string, number> } ? Record<keyof Options["customLevels"], LogFn> : Record<never, LogFn>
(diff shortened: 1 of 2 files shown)


UndefinedBehaviour:
The basic idea is to make `LoggerOptions` a [generic](https://github.com/pinojs/pino/pull/1843/files#diff-bb1e88b0612556147dae1eff5d068f3260d32f523c81d54efc3a4a4750a9b182R323) type.
```
const logger = pino({
    name: 'logger',
    level: 'warn',
    customLevels: {
        myLevel: 25
    }
});
```
This way, given a call such as this one, tsc is still able to infer the type of customLevels (so that we can call `logger.myLevel('blabla')`) but the type of the options object itself is no longer inferred so if we mistype the name of an option (or if we use an option tht simply doesn't exist), tsc is able to notify us.

mcollina (approved):
lgtm

mcollina:
I'm sorry but I'll need to rever this as it breaks Fastify types badly and I don't have time to fix them right now.

mcollina:
Can you open a fresh PR again that does not break Fastify types, or provide a matching PR to Fastify?

Issue #1741, opened 2023-07-05
Question: Is it possible to send to default STDOUT _and_ to another transport?


muya:
Hello. I'm trying to configure pino as follows:

- All logs should go out to STDOUT (i.e. the default behaviour)
- ERROR and above should _also_ be sent to Sentry (via pino-sentry-transport)

However, it seems that enabling pino-sentry-transport disables the STDOUT logging.

How can one configure so that both work?

_Note: this is in a production setting (inside a Docker container), so I'd rather not use pino-pretty for STDOUT_

I think the approach would be to have some sort of "stdout transport", but I don't know how to configure this together with the pino-sentry-transport.

muya:
The solution is to use `pino/file`, with `destination` set to 1:

_Example below is based on `Fastify`, sending logs to both stdout & Sentry (for error and above)_

```ts
const transportTargets: TransportTargetOptions[] = [
    {
      target: 'pino/file',
      options: {
        destination: 1
      },
      level: 'info'
    },
    {
      target: 'pino-sentry-transport',
      options: {
        sentry: {
          dsn: 'https://some/sentry/url'
        },
        minLevel: 40
      },
      level: 'info'
    }
  ]

const logger = {
    level: config.logLevel,
    transport: {
      targets: transportTargets
    }
  }

// Pass this logger to the fastify instance

const server: FastifyInstance = Fastify({logger: logger})

// continue with Fastify setup.
```

PR #1860, opened 2023-12-06
docs: Clarify how to log both to a custom transport and to stdout

kibertoad:
As evidenced by https://github.com/pinojs/pino/issues/1741, and also by myself failing to understand how to do this despite trying to read documentation, this (fairly common) use-case might benefit from a more explicit section.

Yes, it is also mentioned in documentation for destination of "pino/file", but I've skipped the whole section because it's called "file" and its description says "The `pino/file` transport routes logs to a file (or file descriptor).". Not everyone knows that stdout is a file descriptor - maybe we should make that part more clear as well, e. g. saying "(or file descriptor, such as 1 for stdout)".

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -195,13 +195,13 @@ callback is called or the returned promise resolves. Otherwise, log lines will b
 
 ### Combining transport with writing to an stdout
 
-In case you want to both use a custom transport, and output the unprocessed log entries to stdout, you can use 'pino/file' transport configured with `destination: 1`:
+In case you want to both use a custom transport, and output the log entries with default processing to STDOUT, you can use 'pino/file' transport configured with `destination: 1`:
 
 ```js
     const transports = [
       {
         target: 'pino/file',
-        options: { destination: 1 } // this writes to stdout
+        options: { destination: 1 } // this writes to STDOUT
       },
       {
         target: 'my-custom-transport',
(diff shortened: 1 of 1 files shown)


docs/transports.md
```diff
@@ -238,7 +257,7 @@ const logger = pino({
     pipeline: [{
       target: './my-transform.js'
     }, {
-      // Use target: 'pino/file' to write to stdout
+      // Use target: 'pino/file' with 'options: { destination: 1 }' to write to STDOUT
```
jsumners, line 260: The added fragment to this comment does not seem to correspond to the example. What is it trying to explain?

docs/transports.md
```diff
@@ -193,6 +193,25 @@ To consume async iterators in batches, consider using the [hwp](https://github.c
 The `close()` function is needed to make sure that the stream is closed and flushed when its
 callback is called or the returned promise resolves. Otherwise, log lines will be lost.
 
+### Combining transport with writing to an stdout
```
jsumners, line 196:
```suggestion
### Writing to a custom transport & stdout
```

docs/transports.md
kibertoad: It clarifies the comment that was already there. It states that in case you want to log as-is, instead of prettifying, one needs to use pino/file. Original comment made it unclear how to do so.

jsumners: I'd rather see this changed to have the second transport be the `pino/file` transport. The less we push `pino-pretty` in the docs, the better.

kibertoad: done

jsumners (approved):

Issue #1865, opened 2023-12-14
Pino 8.17.0 breaks logging from catch in TypeScript


castdin:
#1783  seems to have re-introduced #1115 as `LogFn` no longer accepts an object of type `unknown`, which is what TypeScript emits for caught errors.

<img width="861" alt="image" src="https://github.com/pinojs/pino/assets/7321311/fde14354-7d76-4d17-a5ba-7a737748ff8f">

castdin:
Ah - I see now from the discussion in that PR that this is intentional. This is unfortunate as it constitutes a breaking change for those who previously logged this way. 

From a DX perspective it's also somewhat surprising that Pino won't accept the error object as-is.

mcollina:
Thanks, will revert asap.

https://github.com/pinojs/pino/pull/1866

PR #1866, opened 2023-12-14
Support unknown, reverts 1783, add tests

mcollina:
This reverts #1783.
I as not aware that TypeScript matched errors as `unknown`.

Specifically, the following should be supported without any additional helpers:

```
try {
  dangerousFunc();
} catch (err) {
  logger.error(err)
}
```

As this is part of the Pino and should be supported on both JavaScript and TypeScript.

Fixes https://github.com/pinojs/pino/issues/1865

cc @Zamiell @kibertoad


mcollina:
I've also added a couple of tests.

kibertoad (approved):

PR #1871, opened 2023-12-23
fix: type of child logger with custom levels

UndefinedBehaviour:
fixes child logger types so that stuff like this doesn't error anymore
```
const parentLogger = pino({
  customLevels: { myLevel: 90 },
  onChild: (child) => { const a = child.myLevel; }
}, process.stdout);
parentLogger.onChild = (child) => { child.myLevel(''); };
```


mcollina (approved):
lgtm

PR #1874, opened 2024-01-03
Correct type of `pino`'s `stream` parameter

quisido:
The second parameter of the `pino` function can be set to `undefined`. This is relevant when dependency injecting a stream:

```ts
function create(stream: DestinationStream | undefined): Logger {
  return pino({ level: 'debug' }, stream); // <-- second parameter is incorrectly marked as invalid
}
```

This allows me to use `create(undefined)` for default behavior (stdio), but also `create({ write: jest.fn() })` for mocked behavior in unit tests, _without creating a new branch_.

With the definition prior to this PR, the following code is required:

```ts
function create(stream: DestinationStream | undefined): Logger {
  if (typeof stream === 'undefined') { // <-- this branch shouldn't be required, because it does nothing different
    return pino({ level: 'debug' });
  }
  return pino({ level: 'debug' }, stream);
}
```

This creates an unnecessary code branch, as it's the same runtime behavior.

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -799,7 +799,7 @@ declare function pino<CustomLevels extends string = never>(optionsOrStream?: Log
  * relative protocol is enabled. Default: process.stdout
  * @returns a new logger instance.
  */
-declare function pino<CustomLevels extends string = never>(options: LoggerOptions<CustomLevels>, stream: DestinationStream): Logger<CustomLevels>;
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
Can you add a test with `tsd`?

quisido:
Sorry for the delay. I've added it. 👍

mcollina (approved):
lgtm

quisido:
Is it possible to re-run the failed test? I think it timed out.

PR #1881, opened 2024-01-13
fix: destination docs and types

### LINKED ISSUE #1312: Confusing "destination" docs
The docs for destination are a bit confusing https://getpino.io/#/docs/api?id=destination 

It says it can accept an object with a `write` method or a Node.js stream but I found those did not work.

I dug into the code a bit and found that `destination` building a `SonicBoom` instance that is just looking for a string path or a file descriptor.

https://github.com/pinojs/pino/blob/54cd23b8ca1839fc38902f653327573585692ec0/pino.js#L186-L193
https://github.com/pinojs/pino/blob/54cd23b8ca1839fc38902f653327573585692ec0/lib/tools.js#L335-L336

codershiba:
- Resolves #1312 
- Updates documentation for `pino.destination()`
- Updates types for `pino.destination()`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -575,13 +575,15 @@ parent.child(bindings)
 
 
 <a id="destination"></a>
-### `destination` (SonicBoom | WritableStream | String | Object)
+### `destination` (Number | String | Object | DestinationStream | SonicBoomOpts | WritableStream)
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
Lgtm

jsumners (approved):

PR #1883, opened 2024-01-16
Add ability to override custom levels comparison

obrus-corcentric:


lib/levels.js
```diff
@@ -119,7 +120,19 @@ function getLevel (level) {
...
+}
+
+function genLevelComparison (levelComparison) {
+  if (levelComparison === 'ASC') {
+    return (current, expected) => current >= expected
+  }
+
+  if (levelComparison === 'DESC') {
+    return (current, expected) => current <= expected
+  }
```
mcollina, line 133: Can you please avoid closures and define top level functions?

mcollina (changes requested):
A few things are missing:

* [ ] test for the new type (we use `tsd`)
* [ ] docs for the new option

lib/levels.js
obrus-corcentric: Sure, done.

obrus-corcentric:
> A few things are missing:
> 
> * [ ]  test for the new type (we use `tsd`)
> * [ ]  docs for the new option

Updated doc and types tests

mcollina (approved):
lgtm

lib/levels.js
```diff
@@ -119,7 +120,23 @@ function getLevel (level) {
...
+
+function compareLevel (direction, current, expected) {
+  if (direction === 'DESC') {
+    return current <= expected
+  }
+
+  return current >= expected
+}
+
+function genLevelComparison (levelComparison) {
```
jsumners, line 134: Please add jsdoc blocks to these functions.

lib/levels.js
```diff
@@ -180,6 +197,14 @@ function assertNoLevelCollisions (levels, customLevels) {
   }
 }
 
+function assertLevelComparison (levelComparison) {
```
jsumners, line 200: jsdoc

lib/levels.js
```diff
@@ -119,7 +120,23 @@ function getLevel (level) {
 function isLevelEnabled (logLevel) {
   const { values } = this.levels
   const logLevelVal = values[logLevel]
-  return logLevelVal !== undefined && (logLevelVal >= this[levelValSym])
+  return logLevelVal !== undefined && this[levelCompSym](logLevelVal, this[levelValSym])
+}
+
+function compareLevel (direction, current, expected) {
+  if (direction === 'DESC') {
```
jsumners, line 127: Can we use constants for the strings so that we are not repeating them?

lib/levels.js
obrus-corcentric: Done

lib/levels.js
obrus-corcentric: Done

lib/levels.js
obrus-corcentric: Done

lib/constants.js
```diff
@@ -0,0 +1,24 @@
...
+  info: 30,
+  warn: 40,
+  error: 50,
+  fatal: 60
+}
+
+/**
+ * @enum {"ASC" | "DESC"}
+ */
+const SORTING_ORDER = {
```
jsumners, line 16: These jsdoc blocks are missing the documentation part.

lib/constants.js
```diff
@@ -0,0 +1,24 @@
...
+  trace: 10,
+  debug: 20,
+  info: 30,
+  warn: 40,
+  error: 50,
+  fatal: 60
+}
+
+/**
+ * @enum {"ASC" | "DESC"}
```
jsumners, line 14: I am unfamiliar with this syntax. Can you point me to the documentation in https://jsdoc.app where it describes using string values for the type like this?

lib/levels.js
```diff
@@ -119,7 +113,36 @@ function getLevel (level) {
...
+
+/**
+ *
+ * @description compare `current` level with `expected` to make sure `current` level is enabled or not
+ * @param {SORTING_ORDER} direction
+ * @param {number} current
+ * @param {number} expected
+ * @returns {boolean}
+ */
+function compareLevel (direction, current, expected) {
```
jsumners, line 127:
```suggestion
/**
 * Determine if the given `current` level is enabled by comparing it
 * against the current threshold (`expected`).
 *
 * @param {SORTING_ORDER} direction
 * @param {number} current
 * @param {number} expected
 * @returns {boolean}
 */
function compareLevel (direction, current, expected) {
```

lib/levels.js
```diff
@@ -119,7 +113,36 @@ function getLevel (level) {
...
-  return logLevelVal !== undefined && (logLevelVal >= this[levelValSym])
+  return logLevelVal !== undefined && this[levelCompSym](logLevelVal, this[levelValSym])
+}
+
+/**
+ *
+ * @description compare `current` level with `expected` to make sure `current` level is enabled or not
+ * @param {SORTING_ORDER} direction
+ * @param {number} current
+ * @param {number} expected
```
jsumners, line 124: These params should have descriptions to explain their purposes.

lib/levels.js
jsumners, line 127: Would this be a clearer signature?
```suggestion
function isLevelEnabled (levelValue, threshold, sortOrder = SORTING_ORDER.DESC) {
```

lib/levels.js
```diff
@@ -119,7 +113,36 @@ function getLevel (level) {
...
+function compareLevel (direction, current, expected) {
+  if (direction === SORTING_ORDER.DESC) {
+    return current <= expected
+  }
+
+  return current >= expected
+}
+
+/**
+ * @description create a level comparison function based on `levelComparison`
```
jsumners, line 136: This description does not help me understand the purpose of this function. What is the purpose of creating such a function? What use cases is the resulting function intended for?

Please also omit the explicit `@description` tag in favor of an opening text block.

lib/levels.js
```diff
@@ -188,7 +225,8 @@ module.exports = {
   setLevel,
   isLevelEnabled,
   mappings,
```
jsumners, line 227: Why has `levels` been removed? This is a breaking change.

lib/levels.js
```diff
@@ -180,6 +203,20 @@ function assertNoLevelCollisions (levels, customLevels) {
...
+/**
+ * @description Validates whether levelComparison is correct
+ * @throws Error
+ * @param {SORTING_ORDER | Function} levelComparison
+ * @returns
+ */
+function assertLevelComparison (levelComparison) {
+  if (Object.values(SORTING_ORDER).includes(levelComparison) || typeof levelComparison === 'function') {
+    return
+  }
```
jsumners, line 215: I think this might be more efficient:
```suggestion
  if (typeof levelComparison === 'function') {
  		return
  }
  if (typeof levelComparison === 'string' && Object.prototype.hasOwnProperty.call(SORTING_ORDER, levelComparison)) {
    return
  }
```

lib/levels.js
```diff
@@ -180,6 +203,20 @@ function assertNoLevelCollisions (levels, customLevels) {
...
+
+  throw new Error('Levels comparison should be one of "ASC", "DESC" or "function" type')
```
jsumners, line 217: @mcollina do we really want to throw here?

jsumners (changes requested):

lib/constants.js
obrus-corcentric: I'm sorry, there is no exact mention of that, but [https://jsdoc.app/tags-type](here) exists a note: _"Full support for Google Closure Compiler-style type expressions is available in JSDoc 3.2 and later."_

lib/levels.js
obrus-corcentric: They actually wasn't they moved to `lib/constants.js`

lib/levels.js
obrus-corcentric: The signature of this function starts from order to be able to create a single sorting function using `bind` with predefined first argument `direction`.

lib/constants.js
jsumners: The linked docs also do not show this sort of declaration.

obrus-corcentric: Changed

lib/constants.js
obrus-corcentric: Done

lib/levels.js
obrus-corcentric: Done

lib/levels.js
obrus-corcentric: Done

lib/levels.js
obrus-corcentric: Added suggested changes

jsumners (approved):

PR #1901, opened 2024-02-09
`setLevel` should respect level comparison option

obrus-corcentric:
@mcollina @jsumners 

Previous PR #1883 contained only a half of a change, i just realized that `setLevel` also should respect `levelComparison`, not only `isLevelEnabled`

### DIFF
--- a/lib/levels.js
+++ b/lib/levels.js
@@ -88,7 +88,7 @@ function setLevel (level) {
   const hook = this[hooksSym].logMethod
 
   for (const key in values) {
-    if (!levelComparison(values[key], levelVal)) {
+    if (levelComparison(values[key], levelVal) === false) {
       this[key] = noop
       continue
     }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

lib/levels.js
```diff
@@ -84,10 +84,11 @@ function setLevel (level) {
   const preLevelVal = this[levelValSym]
   const levelVal = this[levelValSym] = values[level]
   const useOnlyCustomLevelsVal = this[useOnlyCustomLevelsSym]
+  const levelComparison = this[levelCompSym]
   const hook = this[hooksSym].logMethod
 
   for (const key in values) {
-    if (levelVal > values[key]) {
+    if (!levelComparison(values[key], levelVal)) {
```
jsumners, line 91:
```suggestion
    if (levelComparison(values[key], levelVal) === false) {
```

Issue #1445, opened 2022-05-30
proposal: spy transport


simone-sanfratello:
I'd like to have spy capabilities on `pino`
~~since its design, we can't use `sinon` or so to add spy capabilities on a `pino` instance~~

My use case: I want to ensure logs output, since alarms and metrics are bases on them. Because of application design, I can't inject a dummy log instance to spy, I need to spy the real instance, and I believe it's the cleanest solution

I'm thinking to write a transport, does it make sense?

so I'd like to do something like

`logger.js`

```js
const destination = process.env.NODE_ENV == 'test' ? require('pino-spy') : productionDestination
const logger = pino(options, destination)
```

`test.js`

```js
await t.rejects(() => fetchS3(...))
t.ok(logger.error.calledWith('Not Found S3, URL: https://bucket.s3....'))
```

of course, more api on spy are required

Thoughts?

mcollina:
You can do it with https://getpino.io/#/docs/api?id=pino-multistream and split2.

Take a look at:
- https://github.com/pinojs/pino/blob/37337510704bf81cfb1015d7729439cd599ef5b2/test/helper.js#L31-L42 
- https://github.com/pinojs/pino/blob/0a56154fa0a24ac2b4015b597e6ae06468821699/test/timestamp.test.js#L21

I think a less convoluted API to do that would be nice indeed.

simone-sanfratello:
I'll go with that solution, thanks!

kibertoad:
Maybe issue should be kept open, as there is still work to be done?

simone-sanfratello:
just an update, can be done with `sinon` too

EcksDy:
For anyone coming here searching for an easy way to spy on pino output.   

I've ended up using a simpler version of the sink function suggested by @mcollina, that will call the spy function(in my case `jest.fn()`) on each new log line:
```typescript
...
  const spyOnPinoOutput = jest.fn();
  const spyStream: DestinationStream = sink(spyOnPinoOutput);
  afterEach(() => jest.restoreAllMocks());
  
  const pino = pino(config, spyStream);
  
  it('message', async () => {
      logger.log('Some string');
      expect(spyOnPinoOutput).toHaveBeenLastCalledWith(
        matchObject({ // Function with templated expect.anything() on some properties
          level: 'log',
          message,
        }),
      );
    });
    
   function sink(spyFunc: (...args: any[]) => void) {
    const result = split((data) => {
      try {
        return JSON.parse(data);
[...]

Issue #1274, opened 2021-12-20
[question] How to "record" logs for testing purposes 


giltayar:
We have tests in our app that verify that logs were created and have the correct information. 

To do that, we need access to the final logs output by pino. The way we do it now is to use the `prettyPrint` option in `pino` to get at these final log JSONs and store them in an array that the test can access later.

Unfortunately, the `prettyPrint` option is being deprecated, so we need another solution. I looked at transports, and would love to write a transport to do this, but transports run in a worker thread, and there is no way for that worker thread to communicate with the main thread in order for it to pass the logs back to the main thread.

So...

1. Is there a way for a transport to pass information back to the main thread?
2. Am I missing something and there is a simpler way to do this?

(And thanks for a wonderful library!)

giltayar:
I noticed that what is returned from `.transport()` is the `ThreadStream`, and that `ThreadStream` has a property called `worker`, so I can theoretically use that to listen to messages from the transport and receive the final JSON logs that way.

But I have no guarantees on this, and this could change from one version of pino to another.

mcollina:
You can keep using `pino-pretty` directly:

```js
const pino = require('pino')
const pretty = require('pino-pretty')
const logger = pino(pretty())

logger.info('hi')
```

giltayar:
@mcollina I believe you misunderstood. I do not want to do pinopretty. In pino@6 (and 7), I could do this:


```js
const logs = []

const logger = pino({
  prettyPrint: true,
  prettifier: log => logs.push(log)
})

foo(logger)

expect(logs).to.eql({event: 'dsfjshdf', level: 30, ...})
```

We used this for testing purposes to see that the function used the logger and output the correct logs (obviously the code below is a simplified version of what we use).

Unfortunately, the `prettyPrint` option is being deprecated, so we need another solution. I looked at transports, and would love to write a transport to do this, but transports run in a worker thread, and there is no way for that worker thread to communicate with the main thread in order for it to pass the logs back to the main thread.

So...

    Is there a way for a transport to pass information back to the main thread?
    Am I missing something and there is a simpler way to do this?

jsumners:
Pass a custom stream that writes lines to your array.

![image](https://user-images.githubusercontent.com/321201/146782215-7239307d-2fea-494b-a3cb-49e855c78649.png)

![image](https://user-images.githubusercontent.com/321201/146782270-fdf39840-33c6-40a1-8414-9d5fbad51428.png)

[...]

PR #1908, opened 2024-02-24
add a test module with some test API

ruddenchaux:
Added the help test paragraph with some example also. 

I'm available for any adjustments and suggestions.

Closes: #1445

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1394,7 +1394,7 @@ const { test } = require('pino')
 
 <a id="pino-test-sink"></a>
 ### `pino.test.sink({ destroyOnError = false, emitErrorEvent = false }) => Transform`
-Create a pino destination stream to easily spy on over the logs created by pino.
+Create a Pino destination stream to easily inspect the logs processed by Pino.
 
 ```js
 const pino = require('pino')
@@ -1455,7 +1455,7 @@ stream.on('error', (err) => {
 
 <a id="pino-test-once"></a>
 ### `pino.test.once(stream, expected, is) => Promise<void>`
-Assert that a single pino log is expected.
+Assert that a single log is expected.
 
 ```js
 const pino = require('pino')
@@ -1487,7 +1487,7 @@ await pino.test.once(stream, { msg: 'bye world', level: 30 }, is) // throw an er
 
 <a id="pino-test-consecutive"></a>
 ### `pino.test.consecutive(stream, expected, is) => Promise<void>`
-Assert that consecutive pino's logs are expected.
+Assert that consecutive logs are expected.
 
 ```js
 const pino = require('pino')
@@ -1525,9 +1525,11 @@ await pino.test.consecutive(stream, expected, is) // doesn't throw an error
 await pino.test.consecutive(stream, [{ msg: 'bye world', level: 30 }], is) // throw an error
 ```
 
-Both `pino.test.once` and `pino.test.consecutive` use the internal `check` function
-- that asserts the `chunk.time`, `chunk.pid` and `chunk.hostname`, so that you is avoided to do it
-- that use the default `deepStrictEqual` assert function of the `node:assert` module.
+Both `pino.test.once` and `pino.test.consecutive` internally
+- assert log message `time` is less than or equal to the current time
+- assert log message `pid` matches the current process id
+- assert log message `hostname` matches the current hostname
+- uses the default `deepStrictEqual` assert function of the `node:assert` module.
 
 See
 * [Transform](https://nodejs.org/api/stream.html#class-streamtransform)
--- a/docs/api.md
+++ b/docs/api.md
@@ -1457,6 +1457,12 @@ stream.on('error', (err) => {
 ### `pino.test.once(stream, expected, is) => Promise<void>`
 Assert that a single log is expected.
 
+The function internally
+- assert log message `time` is less than or equal to the current time
+- assert log message `pid` matches the current process id
+- assert log message `hostname` matches the current hostname
+- uses the default `deepStrictEqual` assert function of the `node:assert` module.
+
 ```js
 const pino = require('pino')
 
@@ -1488,6 +1494,12 @@ await pino.test.once(stream, { msg: 'bye world', level: 30 }, is) // throw an er
 <a id="pino-test-consecutive"></a>
 ### `pino.test.consecutive(stream, expected, is) => Promise<void>`
 Assert that consecutive logs are expected.
+
+The function internally
+- assert log message `time` is less than or equal to the current time
+- assert log message `pid` matches the current process id
+- assert log message `hostname` matches the current hostname
+- uses the default `deepStrictEqual` assert function of the `node:assert` module.
 
 ```js
 const pino = require('pino')
@@ -1524,16 +1536,6 @@ function is (received, expected, msg) {
 await pino.test.consecutive(stream, expected, is) // doesn't throw an error
 await pino.test.consecutive(stream, [{ msg: 'bye world', level: 30 }], is) // throw an error
 ```
-
-Both `pino.test.once` and `pino.test.consecutive` internally
-- assert log message `time` is less than or equal to the current time
-- assert log message `pid` matches the current process id
-- assert log message `hostname` matches the current hostname
-- uses the default `deepStrictEqual` assert function of the `node:assert` module.
-
-See
-* [Transform](https://nodejs.org/api/stream.html#class-streamtransform)
-* [split2](https://github.com/mcollina/split2#readme)
 
 ## Interfaces
 <a id="pino-multistreamres"></a>
--- a/docs/api.md
+++ b/docs/api.md
@@ -1394,7 +1394,7 @@ const { test } = require('pino')
 
 <a id="pino-test-sink"></a>
 ### `pino.test.sink({ destroyOnError = false, emitErrorEvent = false }) => Transform`
-Create a stream using `split2` so that each line is a chunk and each chunk is JSON parsed.
+Create a pino destination stream to easily spy on over the logs created by pino.
 
 ```js
 const pino = require('pino')
@@ -1455,7 +1455,7 @@ stream.on('error', (err) => {
 
 <a id="pino-test-once"></a>
 ### `pino.test.once(stream, expected, is) => Promise<void>`
-Assert chunk is expected in the stream using the `stream.once` event listener. 
+Assert that a single pino log is expected.
 
 ```js
 const pino = require('pino')
@@ -1487,7 +1487,7 @@ await pino.test.once(stream, { msg: 'bye world', level: 30 }, is) // throw an er
 
 <a id="pino-test-consecutive"></a>
 ### `pino.test.consecutive(stream, expected, is) => Promise<void>`
-Assert consecutive chunks is expected in the stream using the `for await...of` loop. 
+Assert that consecutive pino's logs are expected.
 
 ```js
 const pino = require('pino')
--- a/docs/api.md
+++ b/docs/api.md
@@ -1400,11 +1400,13 @@ Create a stream using `split2` so that each line is a chunk and each chunk is JS
 const pino = require('pino')
 
 const stream = pino.test.sink()
-stream.write('{"hello":"world"}\n')
-stream.end()
+const logger = pino(stream)
+
+logger.info('hello world')
 
 stream.once('data', (data) => {
-  console.log(data.hello) // "world"
+  console.log(chunk.msg) // 'hello world'
+  console.log(chunk.level) // 30
 })
 ```
 
@@ -1494,7 +1496,7 @@ const stream = pino.test.sink()
 const instance = pino(stream)
 
 instance.info('hello world')
-instance.info('test')
+instance.info('hi world')
 
 const expected = [
   { msg: 'hello world', level: 30 },
--- a/docs/api.md
+++ b/docs/api.md
@@ -29,6 +29,10 @@
   * [pino.stdTimeFunctions](#pino-stdtimefunctions)
   * [pino.symbols](#pino-symbols)
   * [pino.version](#pino-version)
+* [Test](#test)
+  * [pino.test.sink()](#pino-test-sink)
+  * [pino.test.once()](#pino-test-once)
+  * [pino.test.consecutive()](#pino-test-consecutive)
 * [Interfaces](#interfaces)
   * [MultiStreamRes](#multistreamres)
   * [StreamEntry](#streamentry)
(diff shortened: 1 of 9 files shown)


mcollina (changes requested):

package.json
```diff
@@ -7,6 +7,7 @@
   "types": "pino.d.ts",
   "browser": "./browser.js",
   "scripts": {
+    "t": "jest test/jest",
```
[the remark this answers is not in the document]
ruddenchaux: just a temporary script used to test the changes. removed

ruddenchaux:
@mcollina let me know if something other is need

mcollina (approved):
lgtm

jsumners:
I'll review soon.

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+Pino provides a test module with some useful APIs that you can use importing it as the following
+```js
+require('pino').test
+// OR
+const { test } = require('pino')
+```
+
+<a id="pino-test-sink"></a>
+### `pino.test.sink({ destroyOnError = false, emitErrorEvent = false }) => Transform`
+Create a stream using `split2` so that each line is a chunk and each chunk is JSON parsed.
```
jsumners, line 1397: The purpose of this method should be explained. How the method does what it does is not useful information in the documentation.

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+stream.end()
+
+stream.on('error', (err) => {
+  console.log(err) // Unexpected token h in JSON at position 0
+})
+```
+
+<a id="pino-test-once"></a>
+### `pino.test.once(stream, expected, is) => Promise<void>`
+Assert chunk is expected in the stream using the `stream.once` event listener.
```
jsumners, line 1458: What is "chunk"? What is the purpose of this method?

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+  }
+}
+
+await pino.test.once(stream, expected, is) // doesn't throw an error
+await pino.test.once(stream, { msg: 'bye world', level: 30 }, is) // throw an error
+```
+
+<a id="pino-test-consecutive"></a>
+### `pino.test.consecutive(stream, expected, is) => Promise<void>`
+Assert consecutive chunks is expected in the stream using the `for await...of` loop.
```
jsumners, line 1490: Again, what are "chunks"? Why would someone use this method?

docs/help.md
```diff
@@ -14,6 +15,55 @@
...
+
+  const expected = [
+    { msg: 'hello world', level: 30 },
+    { msg: 'hi world', level: 30 }
+  ]
+  await pino.test.consecutive(stream, expected, is)
+})
+```
+
+If a complete mock of the pino module is needed, a library as `proxyquire` should be used.
```
jsumners, line 65: Why would someone need to do this? What situations could lead to it? This sentence by itself seems very out-of-place.

jsumners (commented):
In general, I think most people just need an easy to find, clear bit of, documentation of the example shown in https://github.com/pinojs/pino/issues/1274#issuecomment-997971329

docs/help.md
ruddenchaux: It can be helpful when you only want to test how pino is instantiated without testing the log messages.

In the following example, I would mock the pino module and assert that it is instantiated with a dev/prod transport based on the env and to do it a library as `proxyquire` should be used. Am I missing something?
```js
const pino = require('pino')

const { LOG_LEVEL, ENV } = process.env;
let transport;

const devTransport = {
  target: 'pino-pretty',
  options: {
    colorize: true,
  },
};

const prodTransport = {
  target: 'pino-file',
  options: {
    destination: '/var/log/service.log',
  },
};

if (ENV === 'prod') {
  transport = prodTransport
} else {
  transport = devTransport;
}

exports.logger = pino({ transport, level: LOG_LEVEL }).child({ context: 'some-context' });
```

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+Pino provides a test module with some useful APIs that you can use importing it as the following
+```js
+require('pino').test
+// OR
+const { test } = require('pino')
+```
+
+<a id="pino-test-sink"></a>
+### `pino.test.sink({ destroyOnError = false, emitErrorEvent = false }) => Transform`
+Create a pino destination stream to easily spy on over the logs created by pino.
```
jsumners, line 1397:
```suggestion
Create a Pino destination stream to easily inspect the logs processed by Pino.
```

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+stream.end()
+
+stream.on('error', (err) => {
+  console.log(err) // Unexpected token h in JSON at position 0
+})
+```
+
+<a id="pino-test-once"></a>
+### `pino.test.once(stream, expected, is) => Promise<void>`
+Assert that a single pino log is expected.
```
jsumners, line 1458:
```suggestion
Assert that a single log is expected.
```

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+  }
+}
+
+await pino.test.once(stream, expected, is) // doesn't throw an error
+await pino.test.once(stream, { msg: 'bye world', level: 30 }, is) // throw an error
+```
+
+<a id="pino-test-consecutive"></a>
+### `pino.test.consecutive(stream, expected, is) => Promise<void>`
+Assert that consecutive pino's logs are expected.
```
jsumners, line 1490:
```suggestion
Assert that consecutive logs are expected.
```

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+  if (received.msg !== expected.msg) {
+    throw new Error(`expected msg ${expected.msg} doesn't match the received one ${received.msg}`)
+  }
+}
+
+await pino.test.consecutive(stream, expected, is) // doesn't throw an error
+await pino.test.consecutive(stream, [{ msg: 'bye world', level: 30 }], is) // throw an error
+```
+
+Both `pino.test.once` and `pino.test.consecutive` use the internal `check` function
```
jsumners, line 1528: What is "the internal `check` function"?

docs/api.md
```diff
@@ -1380,6 +1384,155 @@ Exposes the Pino package version. Also available on the logger instance.
...
+- that asserts the `chunk.time`, `chunk.pid` and `chunk.hostname`, so that you is avoided to do it
```
jsumners, line 1529: What is "chunk"? Also, overall, the sentence needs some rewording. What is it asserting about the specified log properties? "so that you is avoided to do it" I believe is unnecessary.

docs/help.md
jsumners: I would just check that `exports.logger.level` matches what I expect, but sure, you could use `proxyquire` to do it as well. My point is, this lone sentence doesn't explain why a reader would want to reach for `proxyquire`. And, as such, makes it a frivolous sentence that will only cause confusion.

lib/test.js
jsumners, line 1: The name of this file is going to get caught up in our unit test runs:

https://github.com/pinojs/pino/blob/c109804c6ce4b4545f3dfe1d021ceb698582fe15/.taprc.yaml#L7-L8

Also, please add jsdoc blocks to each function in this file.

docs/help.md
ruddenchaux: You are right; let me remove it.

lib/test.js
ruddenchaux: The unit test runs will take all the files ending with `.test.js` inside the `test` folder. `lib/test.js` is outside the `test` folder, so it will not be caught up by unit test runs. `test/test.test.js` will be taken as it should be.

docs/api.md
ruddenchaux: let me remove it

docs/api.md
ruddenchaux: yep

lib/test.js
```diff
@@ -0,0 +1,107 @@
...
+
+const assert = require('assert')
+const os = require('os')
+const split = require('split2')
+
+/**
+ * Create a Pino destination stream to easily inspect the logs processed by Pino.
+ * @param [options]: The object options to control the behavior of the stream when an error happen. Default: { destroyOnError: false, emitErrorEvent: false }
+ * @returns {import('node:stream').Transform} A Transform stream to be used as destination for the pino function
+ * @example
```
jsumners, line 11: This should be documented like:

```js
/**
 * Description.
 *
 * @param {object} [params]
 * @param {boolean} [destroyOnError=false] Description.
 * @param {boolean} [emitErrorEvent=false] Description.
 *
 * @returns {object} A stream.
 */ 
function foo(params) {}
```

lib/test.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * Check if the chunk is equal to the expected value.
+ * @param {object} chunk: The chunk to be tested
+ * @param {object} expected: The expected value to be tested
+ * @param {assert.deepStrictEqual} is: The assert function to be used.
+ * @throws If the expected value is not equal to the chunk value
+ */
+function check (chunk, expected, is) {
```
jsumners, line 40:
```suggestion
/**
 * Check if the chunk is equal to the expected value.
 *
 * @param {object} chunk The chunk to be tested.
 * @param {object} expected The expected value to be tested.
 * @param {assert.deepStrictEqual} is The assert function to be used.
 *
 * @throws If the expected value is not equal to the chunk value.
 */
function check (chunk, expected, is) {
```

lib/test.js
```diff
@@ -0,0 +1,107 @@
...
+ * @returns A promise that resolves when the expected value is equal to the stream value
+ * @throws If the expected value is not equal to the stream value
+ * @example
+ * const stream = pino.test.sink()
+ * const logger = pino(stream)
+ * logger.info('hello world')
+ * const expected = { msg: 'hello world', level: 30 }
+ * await pino.test.once(stream, expected)
+ */
+async function once (stream, expected, is = assert.deepStrictEqual) {
```
jsumners, line 63:
```suggestion
/**
 * Assert that a single log is expected.
 *
 * @param {import('node:stream').Transform} stream The stream to be tested.
 * @param {object} expected The expected value to be tested.
 * @param {function} [assert=assert.deepStrictEqual] The assert function to be used.
 *
 * @returns A promise that resolves when the expected value is equal to the stream value.
 * @throws If the expected value is not equal to the stream value
 *
 * @example
 * const stream = pino.test.sink()
 * const logger = pino(stream)
 * logger.info('hello world')
 * const expected = { msg: 'hello world', level: 30 }
 * await pino.test.once(stream, expected)
 */
async function once (stream, expected, is = assert.deepStrictEqual) {
```

lib/test.js
```diff
@@ -0,0 +1,107 @@
...
+ * const logger = pino(stream)
+ * logger.info('hello world')
+ * logger.info('hi world')
+ * const expected = [
+ *   { msg: 'hello world', level: 30 },
+ *   { msg: 'hi world', level: 30 }
+ * ]
+ * await pino.test.consecutive(stream, expected)
+ */
+async function consecutive (stream, expected, is = assert.deepStrictEqual) {
```
jsumners, line 98:
```suggestion
/**
 * Assert that consecutive logs are expected.
 *
 * @param {import('node:stream').Transform} stream The stream to be tested.
 * @param {object} expected The expected value to be tested.
 * @param {function} [assert=assert.deepStrictEqual] The assert function to be used.
 *
 * @returns A promise that resolves when the expected value is equal to the stream value.
 * @throws If the expected value is not equal to the stream value.
 *
 * @example
 * const stream = pino.test.sink()
 * const logger = pino(stream)
 * logger.info('hello world')
 * logger.info('hi world')
 * const expected = [
 *   { msg: 'hello world', level: 30 },
 *   { msg: 'hi world', level: 30 }
 * ]
 * await pino.test.consecutive(stream, expected)
 */
async function consecutive (stream, expected, is = assert.deepStrictEqual) {
```

docs/api.md
```diff
@@ -1380,6 +1384,157 @@ Exposes the Pino package version. Also available on the logger instance.
...
+
+await pino.test.consecutive(stream, expected, is) // doesn't throw an error
+await pino.test.consecutive(stream, [{ msg: 'bye world', level: 30 }], is) // throw an error
+```
+
+Both `pino.test.once` and `pino.test.consecutive` internally
+- assert log message `time` is less than or equal to the current time
+- assert log message `pid` matches the current process id
+- assert log message `hostname` matches the current hostname
+- uses the default `deepStrictEqual` assert function of the `node:assert` module.
```
jsumners, line 1532: This part applies to two sections but is only present in one of the sections. So people are likely to miss it if they only reach the `pino.test.once` section. It should be present in both sections and specific to each.

docs/api.md
```diff
@@ -1380,6 +1384,157 @@ Exposes the Pino package version. Also available on the logger instance.
...
+
+See
+* [Transform](https://nodejs.org/api/stream.html#class-streamtransform)
+* [split2](https://github.com/mcollina/split2#readme)
```
jsumners, line 1536: Why would someone need to reference these external resources? What is this "See" section referring to?

ruddenchaux: no longer needed; let me remove it.

lib/test.js
```diff
@@ -32,41 +28,46 @@ function sink ({ destroyOnError = false, emitErrorEvent = false } = {}) {
...
 /**
  * Check if the chunk is equal to the expected value.
- * @param {object} chunk: The chunk to be tested
- * @param {object} expected: The expected value to be tested
- * @param {assert.deepStrictEqual} is: The assert function to be used.
- * @throws If the expected value is not equal to the chunk value
+ *
+ * @param {object} chunk The chunk to be tested.
+ * @param {object} expected The expected value to be tested.
+ * @param {nodeAssert.deepStrictEqual} assert The assert function to be used.
```
jsumners, line 34: `{nodeAssert.deepStrictEqual}` is not a valid type. See https://jsdoc.app/tags-type. It should be `{function}`.

mcollina (changes requested):
Please convert the jest tests to tap. Jest is used here to ensure everything works there, but it would be a nightmare to debug anything there.

mcollina:
> In general, I think most people just need an easy to find, clear bit of, documentation of the example shown in https://github.com/pinojs/pino/issues/1274#issuecomment-997971329

Maybe we could just do a standalone module?

ruddenchaux:
> Please convert the jest tests to tap. Jest is used here to ensure everything works there, but it would be a nightmare to debug anything there.

I have already tested all the new APIs with tap `test/test.test.js`. I have also included some jest tests `test/jest/test.spec.js` to ensure everything works there, but I will remove it if you think it is useless.

ruddenchaux:
> Can you please send this PR to https://github.com/pinojs/pino-test? I think it's better if it's in a separate package.

Done https://github.com/pinojs/pino-test/pull/1

mcollina:
thx, closing thins

PR #1922, opened 2024-03-12
Add pino-discord-webhook transport information

fabulousgk:
This PR adds a discord webhook transport to the list of known transports.

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -422,6 +422,7 @@ PRs to this document are welcome for any new transports!
 + [pino-axiom](#pino-axiom)
 + [pino-opentelemetry-transport](#pino-opentelemetry-transport)
 + [@axiomhq/pino](#@axiomhq/pino)
++ [pino-discord-webhook](#pino-discord-webhook)
 
 ### Legacy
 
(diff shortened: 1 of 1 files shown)


mcollina:
Can you please add a few tests?
How could someone get the webhook url? Are there security tokens involved?

fabulousgk:
I added information about how to create a webhook and get the URL.

I have tried adding testing, but I have no experience with test suites in JS. I looked at the Slack one, but my setup is a bit different.

The main issue is I am not sure how to test the actual sending as that is wrapped in the callback function.

So I can test that the transport is created, but nothing beyond that.  If someone gives me hints I will add it.

I will alos ask elsewere.

I noticed that there are transports without tests, is this required for publising?

mcollina:
> I noticed that there are transports without tests, is this required for publising?

I double checked and only the logtail one does not have tests. This is unfortunate and it slipped.

Overally, yes, I consider tests to be a requirement for inclusion in this list.

---

To test this I would spin up an http server and pass its URL as the webhook. Then, verify this endpoint is called with the expected output.

fabulousgk:
Well, I added a test.  The idea of a webserver seemed not impossible since what I am doing is simply calling a class in the discord.js library, which handles the rest of the interaction.  So I am testing that that call occurs.

mcollina (approved):
lgtm

PR #1929, opened 2024-03-22
Fix deprecated dependency notation

mbtools:
Minor change to replaces a semver v1 notation (with leading `v`) with semver v2.

PS: The npm semver package most likely will deprecate semver v1 support in next major.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "atomic-sleep": "^1.0.0",
     "fast-redact": "^3.1.1",
     "on-exit-leak-free": "^2.1.0",
-    "pino-abstract-transport": "v1.1.0",
+    "pino-abstract-transport": "1.1.0",
     "pino-std-serializers": "^6.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -104,7 +104,7 @@
     "atomic-sleep": "^1.0.0",
     "fast-redact": "^3.1.1",
     "on-exit-leak-free": "^2.1.0",
-    "pino-abstract-transport": "v1.1.0",
+    "pino-abstract-transport": "1.1.0",
```
mcollina, line 107:
```suggestion
    "pino-abstract-transport": "^1.1.0",
```

mcollina (changes requested):
good catch! I think it should be a caret.

mbtools:
Makes sense. Changed

jsumners (approved):

mcollina (approved):
lgtm

PR #1946, opened 2024-04-22
Update pino types for browser.formatters

KatelynKim:
Add the missing types for the `browser.formatters` option.


mcollina (changes requested):
Can you add a test for these types? We use tsd.

KatelynKim:
> Can you add a test for these types? We use tsd.

Hello, I added a test. Could you please check? Thank you!

mcollina (approved):
lgtm

Issue #1948, opened 2024-04-24
Import error when using worker with pkg


nagyszabi:
Hi there!

When using pkg (more specifically [yao-pkg](https://github.com/yao-pkg/pkg)) the transport stream builder is unable to load files.
This is happening on node v20 while using CommonJS modules throughout the whole project, and it's throwing the following error:
```
TypeError [ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING]: A dynamic import callback was not specified.
    at importModuleDynamicallyCallback (node:internal/modules/esm/utils:213:9)
    at eval (eval at <anonymous> (C:\snapshot\backend\node_modules\real-require\src\index.js:4:20), <anonymous>:3:1)
    at loadTransportStreamBuilder (C:\snapshot\backend\node_modules\pino\lib\transport-stream.js:28:19)
    at C:\snapshot\backend\node_modules\pino\lib\worker.js:14:22
    at Array.map (<anonymous>)
    at C:\snapshot\backend\node_modules\pino\lib\worker.js:13:39
    at start (C:\snapshot\backend\node_modules\thread-stream\lib\worker.js:61:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
Emitted 'error' event on ThreadStream instance at:
    at Immediate.<anonymous> (C:\snapshot\backend\node_modules\thread-stream\index.js:369:12)
    at process.processImmediate (node:internal/timers:478:21)

Node.js v20.11.1
```

This is essentially [the same issue I encountered in the thread-stream package](https://github.com/pinojs/thread-stream/issues/143).

And just like in the other issue, adding the `ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING` error code [here](https://github.com/nagyszabi/pino/blob/d9911b0ec0ddcf71c6371da7be363e454e9f93e9/lib/transport-stream.js#L34-L37) fixes the problem:
```ts
else if (error.code === undefined || error.code === 'ERR_VM_DYNAMIC_IMPORT_CALLBACK_MISSING') {
      // When bundled with pkg, an undefined error is thrown when called with realImport
      fn = realRequire(decodeURIComponent(target))
}
```

I am working on creating a PR with [the same fix as I did for thread-stream](https://github.com/pinojs/thread-stream/pull/144).

PR #1949, opened 2024-04-24
fix(transport-stream): Fix import error when using pkg with node v20

nagyszabi:
Fixes: #1948 

Note:
Because the changes to `thread-stream` aren't published yet, tests will fail.
I tested the code locally, where I manually changed the `thread-stream` package in order to be able to test the changes.

### DIFF
--- a/lib/transport-stream.js
+++ b/lib/transport-stream.js
@@ -31,8 +31,10 @@ async function loadTransportStreamBuilder (target) {
     // See this PR for details: https://github.com/pinojs/thread-stream/pull/34
     if ((error.code === 'ENOTDIR' || error.code === 'ERR_MODULE_NOT_FOUND')) {
       fn = realRequire(target)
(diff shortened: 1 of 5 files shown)


mcollina:
done

nagyszabi:
I fixed the errors for Node v21. I was checking for stderr to be an empty string but it turns out it's a deprecation warning for punnycode:
```
(node:1234) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

```

So I added a regex test which checks exactly the warning above. Anything else will fail the test.

For Node 18 on MacOS I assume it's not related to my commit since it fails on another test, which works in Linux and Windows.

But please run the tests again, and if there is a problem on MacOS I'll look deeper into it. Thank you!

mcollina (approved):
lgtm

Issue #1302, opened 2022-01-18
How can I use multiple targets and pipelines?


bmaupin:
I'd like to be able to send logs to syslog and to stdout.

I know I need a pipeline for syslog (pino-syslog+pino-socket), and that works fine, e.g.

```typescript
const pino = require('pino');
const logger = pino({
  transport: {
    pipeline: [
      {
        target: 'pino-syslog',
      },
      {
        target: 'pino-socket',
        options: {
          port: 5140,
          mode: 'udp',
        },
      },
    ],
  },
});
```
My guess would be that I'd need to combine `pipeline` with `targets` so that I can send the logs to multiple targets; one target for stdout (I'm testing with pino-pretty but the format doesn't really matter) and the other target for the syslog pipeline.

I tried this:

```typescript
const logger = pino({
  transport: {
    targets: [
      {
        target: 'pino-pretty',
      },
      {
        pipeline: [
          {
            target: 'pino-syslog',
          },
          {
            target: 'pino-socket',
            options: {
              port: 5140,
              mode: 'udp',
            },
          },
        ],
      },
    ],
  },
});
```

But it throws this error: `TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined`

Thanks!

mcollina:
This is not currently possible. You can't mix & match.
However it would be amazing to support that, it's would be a really nice feature to add.
However I don't have time to work on it, so it would be great to have a PR.

dbacarel:
I'll give it a try - I could see this as a nice to have in preparation to the v8 release

Starkrights:
Heya! I was just thinking the same thing and came across this issue. It looks like @dbacarel's fork hasn't had any commits in relation to this particular feature, so I started poking around myself to see what a potential PR might look like.

From what I understand, the `transport` function checks for the `fullOptions.targets` and `fullOptions.pipeline` properties, and sets the target passed to `buildStream` as either `worker.js` or `worker-pipeline.js`, which return a stream that executes the expected functionality (worker: input->multistream to all targets | worker-pipeline: input->pipeline(first stream -> last stream)`

With the way those workers are written, I see a couple of potential solutions. 

## Solution 1: Abuse the existing functionality of worker.js 
[...]

PR #1954, opened 2024-04-27
Adding support for mix&match pipelines 

dbacarel:
This PR wants to address the use-case of  mix matching pipelines and targets, see discussion #1302.
Currently only `pipeline` or `targets` can be defined.

As a stretch, the current prototype proposes to introduce support for named pipelines that can be referred within the definition of a target.

```javascript
    targets: [
      {
        target: join(__dirname, '..', 'fixtures', 'to-file-transport.js'),
        options: { destination: '/path/to/A'},
        pipeline: 'pipelineA'
      },
      {
        target: join(__dirname, '..', 'fixtures', 'to-file-transport.js'),
        options: { destination: '/path/to/B' }
      },
      {
        target: join(__dirname, '..', 'fixtures', 'to-file-transport.js'),
        options: { destination: '/path/to/C'},
        pipeline: 'pipelineB'
      }
    ],
    pipelines: {
      pipelineA: [
        {
          target: join(__dirname, '..', 'fixtures', 'transform-1.js')
        },
        {
          target: join(__dirname, '..', 'fixtures', 'transform-2.js')
        }
      ],
      pipelineB: [
        {
          target: join(__dirname, '..', 'fixtures', 'transform-3.js')
        },
        {
          target: join(__dirname, '..', 'fixtures', 'transform-4.js')
        }
      ]
    }
  })
```

More details will follow

### Update
The idea of _named_  pipelines was abandoned.
This PR extends `targets` to support  mixing in pipeline definitions within the array of transport configurations

Example:
```javascript
const pino = require('pino')
  targets: [{
    level: 'info',
    target: 'pino-pretty' // must be installed separately
  }, {
    level: 'trace',
    target: 'pino/file',
    options: { destination: '/path/to/store/logs' }
  }, {
    pipeline: [{
      target: 'pino-syslog' // must be installed separately
    }, {
      target: 'pino-socket' // must be installed separately
    }]
  }
  ]
})
pino(transport)
```


mcollina:
I think this is too complex, would avoid the "naming" altogether.

dbacarel:
One of the advantages that I saw in adopting _named_ pipeline would have been enabling the implementation of _one input, multiple outputs_, similarly to `tee`. Potentially, the transformation for an input would have been produced only once (for a given pipeline) and only then passed through to the destination target(s).

On the user side, this simplifies the configuration since the pipeline requires to be defined only once.
On the implementation side, it would been easier to identify identical pipes, if the optimization mentioned above were to be implemented.

Besides the technical aspects, I also understand that it also comes down to how common such use case is so, I don't feel like pushing this idea too much.

mcollina:
I feel most devs won't understand that.

dbacarel:
Pity but fair enough. 

I'll revert the changes related to the named pipelines in favor to a configuration like the following one, looks intuitive enough to me.

```javascript
const pino = require('pino')
  targets: [{
    level: 'info',
    target: 'pino-pretty' // must be installed separately
  }, {
    level: 'trace',
    target: 'pino/file',
    options: { destination: '/path/to/store/logs' }
  }, {
    pipeline: [{
      target: 'pino-syslog' // must be installed separately
    }, {
      target: 'pino-socket' // must be installed separately
    }]
  }
  ]
})
pino(transport)
```

dbacarel:
Removing the Draft mark as the code is now in a more decent shape for some feedback.

Besides the implementation of the support for pipelines illustrated above,  I also thought to get rid of `worker-pipeline` and unify the logic into one worker only, `worker.js`.

The following schema illustrates how the streams in `targets` with pipelines and `pipeline` are handled in the new worker 

```
┌──────────────────────────────────────────────────┐   ┌─────┐                
│                                                  │   │     │                
│                                                  │   │  p  │                
│                   target    ┌───────────────┐    │   │  i  │                
│                 ──────────► │               │    │   │  n  │                
│   targets         target    │  pino.        │    │   │  o  │                
│ ────────────►   ──────────► │  multistream  ├────┼──►│  .  │       source   
│                   target    │               │    │   │  m  │         │      
│                 ──────────► └───────────────┘    │   │  u  │         │write 
│                                                  │   │  l  │         ▼      
│                  pipeline   ┌───────────────┐    │   │  t  │      ┌────────┐
│                 ──────────► │ PassThrough   ├────┼──►│  i  ├──────┤        │
│                             └───────────────┘    │   │  s  │ write│ Thread │
│                                                  │   │  t  │◄─────┤ Stream │
│                  pipeline   ┌───────────────┐    │   │  r  │      │        │
│                 ──────────► │ PassThrough   ├────┼──►│  e  │      └────────┘
│                             └───────────────┘    │   │  a  │                
│                                                  │   │  m  │                
│                                                  │   │     │                
│                        OR                        │   │     │                
│                                                  │   │     │                
│  pipeline     ┌──────────────┐                   │   │     │                
│ ────────────► │ PassThrough  ├───────────────────┼──►│     │                
│               └──────────────┘                   │   │     │                
│                                                  │   │     │                
└──────────────────────────────────────────────────┘   └─────┘                
```

There're probably some naïve decisions here with impacts I don't see, for example pipelines defined via `pipeline` property now need to go through an additional stream layer represented by `pino.multistream`.

mcollina:
`pino.multistream()` has a few additional consideration (such as the minimum level), that makes this architecture a bit more fragile.

I would recommend:

* [ ] moreving the 2nd multistream for all targets, just reusing the top one.
* [ ] add a special cases for 1 target or 1 pipeline that bypass all the heavy machinery, essentially being used by ThreadStream directly (same as it is right now).

dbacarel:
> pino.multistream() has a few additional consideration (such as the minimum level), that makes this architecture a bit more fragile.

Yes, that's something I wanted follow up a bit as I wasn't sure about the implications of logging level vs pipelines in this context. I'll read more about this.

About the recommended changes:

> moreving the 2nd multistream for all targets, just reusing the top one.

There is actually one layer of `pino.multistream`, the current state looks in fact like this
```
┌────────────────────────────────────────────────┐     ┌─────┐                
│                                                │     │     │                
│                                                │     │     │                
│                   target                       │     │     │                
│               │ ───────────────────────────────┼────►│     │                
│   targets     │   target                       │     │  p  │                
│ ────────────► │ ───────────────────────────────┼────►│  i  │       source   
│               │   target                       │     │  n  │         │      
│               │ ───────────────────────────────┼────►│  o  │         │write 
│               │                                │     │  .  │         ▼      
│               │  pipeline   ┌───────────────┐  │     │  m  │      ┌────────┐
│               │ ──────────► │ PassThrough   ├──┼────►│  u  ├──────┤        │
│               │             └───────────────┘  │     │  l  │ write│ Thread │
│               │                                │     │  t  │◄─────┤ Stream │
│               │  pipeline   ┌───────────────┐  │     │  i  │      │        │
│               │ ──────────► │ PassThrough   ├──┼────►│  s  │      └────────┘
│                             └───────────────┘  │     │  t  │                
│                                                │     │  r  │                
│                                                │     │  e  │                
│                        OR                      │     │  a  │                
│                                                │     │  m  │                
│  pipeline                   ┌───────────────┐  │     │     │                
│ ───────────────────────────►│ PassThrough   ├──┼────►│     │                
│                             └───────────────┘  │     │     │                
│                                                │     │     │                
└────────────────────────────────────────────────┘     └─────┘                
```
I drew the wrong schema, my bad.


> add a special cases for 1 target or 1 pipeline that bypass all the heavy machinery, essentially being used by ThreadStream directly (same as it is right now).

Makes sense, will do.

dbacarel:
So
 
- [X] moreving the 2nd multistream for all targets, just reusing the top one.
No actions here, there was already one single layer of multistream involved. My drawing was wrong. 

 - [X] add a special cases for 1 target or 1 pipeline that bypass all the heavy machinery, essentially being used by ThreadStream directly (same as it is right now).
I added a check [here](https://github.com/pinojs/pino/pull/1954/files#diff-df9477f338c214ed04a5e566edd666461eee57595f4b8800d286075824c2ad2cR126)  to skip it in cases either one target or one pipeline only is defined in `targets`.


> pino.multistream() has a few additional consideration (such as the minimum level), that makes this architecture a bit more fragile.

Now regarding this,  I thought that we could introduce the optional [level property](https://github.com/pinojs/pino/pull/1954/files#diff-bb1e88b0612556147dae1eff5d068f3260d32f523c81d54efc3a4a4750a9b182R262) to pipeline definitions as well, in this way `multistream` would know how to properly sort them.
I added this [UT](https://github.com/pinojs/pino/pull/1954/files#diff-ee57362c2c29571b8522651c34ac9d717f4bc3104ed2f66fc55cc18057d1bc6aR86)

mcollina (approved):
lgtm

jsumners:
Sorry. I didn't have a chance over the weekend. I was going to try and review today.

dbacarel:
Thank you 🥳

PR #1961, opened 2024-04-30
Support file URLs when configuring multiple transports

haines:
pino currently supports file URLs in single transports

```ts
pino.transport({ target: "file:///path/to/transport.js" });
```

but in multiple transports it does not:

```ts
pino.transport({ targets: [{ target: "file:///path/to/transport.js" }] });
```

throws

```ts
{
  code: "ERR_INVALID_URL",
  input: "file://file:///path/to/transport.js",
}
```

This PR adds a check for the presence of the `file://` prefix before prepending it to the target.


mcollina (approved):
lgtm

PR #1971, opened 2024-05-12
apply serializers to args once before asObject or transmit

emmyakin:
- Fixes https://github.com/pinojs/pino/issues/1970 
  -  `applySerializers` functioin mutates `args` to logger.*, this cause args to be updated twice in `asObject` and `transmit`  
  - Apply serializers once before `asObject`

- Also includes fix for https://github.com/pinojs/pino/issues/1966
  - Current tests were already failing on the main branch, since `formatters.level` does not spread the returned object, but updates `logData.level = { level: 30, label: 'info' }`

PR #1976, opened 2024-05-15
use boolean flag to ensure serializers are applied once

emmyakin:
```
var logger = pino({
  serializers: {
    extras(data) {
      return { serializedExtras: data };
    },
  },
  browser: {
    serialize: ['extras'],
    asObject: true,
    transmit: {
      send(level, o) {
        is(o.messages[0].extras.serializedExtras, 'world');
      },
    },
  },
})
```
When `opt.browser.transmit` is defined, serializers are not applied to `write`

Conditions when serializers should be applied
- when `opts.serialize` is `true` or `defined`
 - serializers should be applied to `write` and `transmit` operations
- when `opts.serialize` is `false` or `undefined`
 - serializers should be applied to `transmit` operation only

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -302,8 +302,10 @@ function createWrap (self, opts, rootLogger, level) {
       const proto = (Object.getPrototypeOf && Object.getPrototypeOf(this) === _console) ? _console : this
       for (var i = 0; i < args.length; i++) args[i] = arguments[i]
 
-      if (opts.serialize && !opts.transmit) {
+      var argsIsSerialized = false
+      if (opts.serialize) {
         applySerializers(args, this._serialize, this.serializers, this._stdErrSerialize)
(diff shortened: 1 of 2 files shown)


jsumners:
I am not following. Can you please provide a direct statement of the problem this is solving?

mcollina (approved):
lgtm

emmyakin:
@jsumners

> I am not following. Can you please provide a direct statement of the problem this is solving?

Sure

I had created this PR https://github.com/pinojs/pino/pull/1971 to fix serializers being applied twice to logs, i.e, before `write` and in `transmit`

The fix only applies serializers if transmit is not defined, which breaks the original implementation, where we want apply serializers for `write` operations even if `transmit` is defined, but `options.serialize` is true

PR #1986, opened 2024-06-09
Fixing browser side child log issue (#960) child level can now be set at cr…

stevel032:
With this PR, a browser side child logger level can now be set at creation time. More discussions and sample code to demonstrate this problem are in #960.


mcollina (approved):
lgtm

PR #1995, opened 2024-06-24
Update pino.d.ts

miketamis:
Made it so you can't put objects after msg unless it has %s etc in it.

I keep making the mistake of putting the metadata object after the msg. This makes it so you can't do that unless you were intending on using the values in an interpreted output.


mcollina (changes requested):
can you a test for this? we use `tsd` for types

PR #2001, opened 2024-07-04
call onChild when instanciating new child browser logger

hollykurt:
onChild callback is currently not called when instanciating a child broser logger.
This PR corrects this by passing the callback passed to the parent pino ops into the browser context and calling it on new Child.

Test also included.


mcollina (approved):
lgtm

jsumners (approved):

mcollina (approved):
lgtm

PR #2003, opened 2024-07-10
chore: use node: imports

Cangit:
(available since node 14.18)


jsumners:
I'm not clear what this offers. Will it result in issues from people bundling things?

Cangit:
There have been quite some discussions on the nodejs repo when the feature was introduced. Recently added modules don't have unprefixed versions available, like `node:sqlite` and `node:test`. [Nodejs recommends](https://github.com/nodejs/node/issues/38343) the `node:` syntax for requiring core modules in the docs. [Modern linters like biomejs have it on by default](https://biomejs.dev/linter/rules/use-nodejs-import-protocol/). It supports a more extensive standard library in JavaScript. It makes it clear for beginners that you are adding a Node.js builtin module. There are some microscopic cache(skip) benefits and security benefits.

If you are worried about deno and bun support (don't know if pino runs as is currently?) they handle the `node:` syntax fine. My understanding is that pinojs v9 only supports nodejs >=18, so its not breaking in that regard either.

Disclaimer: I am fully aware all of these reasons are opinionated and none are hard hitting. I do however feel obliged to suggest whatever nodejs proper/core brings out to help move the ecosystem forward.

jsumners:
Yes, all of that is true. I am asking what benefit it provides this code base. Any usage of modules that require the prefix will, of course, utilize the prefix. We know that people run this dependency, `pino`, through bundlers. So the big question is: will these changes cause issues with those things and result in complaints here?

Cangit:
Ah, sorry I misunderstood. I will do some research.

Jarred-Sumner:
In Bun, `node:` prefixes are stripped and fully work without those

mcollina (approved):
lgtm

trentm:
Just a data point: This indirectly breaks fastify@4 with Node.js v14 versions before 14.18.

jsumners:
> Just a data point: This indirectly breaks fastify@4 with Node.js v14 versions before 14.18.

Nice. I suspected we'd see issues. We'll revert.

trentm:
> We'll revert.

Please don't feel obliged just for *me*. I am able to cope in my usages. I just have some repos that try to maintain 14.17 as a min-supported Node. In my case I just needed to limit the scope of *testing*. My runtime deps haven't been broken.

jsumners:
I also forgot that this is in the v9 release line and we do not support v14 with `pino@9`: https://github.com/pinojs/pino/blob/8aafa88139890b97aca0d32601cb5ffdd9bda1eb/docs/lts.md?plain=1#L46

trentm:
So, perhaps it is an issue on *fastify@4* that it took a PR to bump its pino dep from 8 to 9?  This PR (which was merged ... without any review?) https://github.com/fastify/fastify/pull/5431

https://github.com/fastify/fastify/blob/4.x/docs/Reference/LTS.md#schedule

jsumners:
😖 I have feelings about Dependabot.

Issue #1861, opened 2023-12-07
Log level functionality broken when describing `transmit`


kyranjamie:
We're using Pino in the [Leather web extension project](https://github.com/leather-wallet/extension/blob/fd5be9b0cde701f720207535ed09c1ce436dc60c/src/shared/logger.ts#L21-L34), with the config:

```
const pinoLogger = pino({
  enabled: !IS_TEST_ENV,
  level: 'info',
  browser: {
    asObject: false,
    transmit: {
      level: 'info',
      send(_level, logEvent) {
        if (!chrome) return;
        logs$.next(logEvent);
      },
    },
  },
});
```

Not sure when this was introduced, but lately, `pinoLogger.debug` statements have been logging to console, despite the level being `info`. In debugging this, I've noticed that when removing the `browser.transmit` object, the logging behaviour works as expected.

jsumners:
Please provide a [minimal reproducible example](https://en.wikipedia.org/wiki/Minimal_reproducible_example). Doing so will help us diagnose your issue. It should be the bare minimum code needed to trigger the issue, and easily runnable without any changes or extra code.

You may use a GitHub repository to host the code if it is too much to fit in a [code block](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks) (or two).

adragich:
Hi, I have the same bug in my app.

There is how to reproduce: https://playcode.io/1935966

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

adragich:
Do you run `browser-test` in CI? Locally it complains about outdated airtap strategy:

``` 
yarn browser-test
$ airtap --local 8080 test/browser*test.js
The --local flag has been removed in favor of providers.

Providers must be installed separately. You might like:

- airtap-manual      (https://github.com/airtap/manual)
- airtap-playwright  (https://github.com/airtap/playwright);
- airtap-system      (https://github.com/airtap/system).
error Command failed with exit code 1.
```

adragich:
Also having a problem with a commit.

Do you have any guideline for contributing except for https://github.com/pinojs/pino/blob/main/CONTRIBUTING.md? Environment requirements, how to test locally etc. I get error in precommit hook (tried with a dummy change in README file):

```
​ FAIL ​ test/pkg/pkg.test.js
 ✖ Command failed: ./test/pkg/index-node14

[...]

PR #2005, opened 2024-07-12
fix(browser): should respect log level with transmit

adragich:
Potential fix for https://github.com/pinojs/pino/issues/1861

Added early return in the case level of the message is below transmit level

Added unit test for the case


jsumners (approved):
Looks good to me.

mcollina (approved):
lgtm

PR #2008, opened 2024-07-15
chore: enable n/prefer-node-protocol eslint rule

techvlad:
Enables rule and enforces style for future contributions https://github.com/eslint-community/eslint-plugin-n/blob/HEAD/docs/rules/prefer-node-protocol.md

Initial work was done here
- https://github.com/pinojs/pino/pull/2003


package.json
```diff
@@ -58,6 +58,9 @@
   "bugs": {
     "url": "https://github.com/pinojs/pino/issues"
   },
+  "engines": {
+    "node": "^18 || >=20"
+  },
```
techvlad, line 63: From my understanding `pino` supports only next node versions: `^18 || >=20`
Required by `eslint-plugin-n` package

jsumners:
I expect that this project will switch to using https://www.npmjs.com/package/neostandard

mcollina:
Yes, exactly

techvlad:
ok, got it, thank you!

PR #2027, opened 2024-08-13
docs: Update transports.md

Imgodmaoyouknow:
In the below example, erros logs will be send to `my-transport.mjs`, all logs will be send to `some-file-transport`.

Moreover, the option to send those logs is `level`.


mcollina (approved):
lgtm

Issue #1556, opened 2022-09-13
Pino in the browser prints the log bindings weirly


segevfiner:
When using pino in the browser multiple child log bindings are just printed one after the other rather than coalesced into a single object: e.g.
```ts
logger.child({ foo: 'foo'}).child({ bar: 'bar'}).info("Hello, %s!", "World);
```

Prints:
```
{ foo: 'foo'}   { bar: 'bar' }   Hello, World!
```

Which is kinda weird, as in node you just get a single object.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

segevfiner:
Offf this is even worse. It prints:
```
{ foo: 'foo'}   { bar: 'bar' }   Hello, %s!    "World"
```

segevfiner:
We might want to handle the console more like this:
```ts
        const consoleLevel = level === "fatal" ? "error" : level;

        console[consoleLevel](
          // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
          ...logEvent.messages,
          Object.assign({}, ...logEvent.bindings)
        );
```

But I'll need to dig into the code a bit to figure out how to make Pino behave like that.

segevfiner:
One idea is to just move bindings to the end instead of at the start:
```ts
diff --git a/browser.js b/browser.js
index 8a72166..f3cf3a5 100644
--- a/browser.js
+++ b/browser.js
@@ -258,11 +258,11 @@ function applySerializers (args, serialize, serializers, stdErrSerialize) {
 
 function bind (parent, bindings, level) {
   return function () {
-    const args = new Array(1 + arguments.length)
-    args[0] = bindings
-    for (var i = 1; i < args.length; i++) {
-      args[i] = arguments[i - 1]
+    const args = new Array(arguments.length + 1)
+    for (var i = 0; i < args.length - 1; i++) {
+      args[i] = arguments[i]
     }
+    args[args.length - 1] = bindings
     return parent[level].apply(this, args)
   }
 }
```

But I have no idea what this will possibly break. I think there is some code that assumes the bindings are at the start. In addition, it would be nice to flatten the bindings into a single object, and possibly add some prefix (User configurable?) before them at the end of the message so we will get something like this instead:
```
Hello, World!    bindings: { foo: 'foo', bar: 'bar' }
```
[...]

PR #2051, opened 2024-09-26
Add asObjectBindingsOnly, allowing write to receive only the bindings as object

segevfiner:
This allows to pass the format string and args as
is the underlying console when to keep their
special handling, for when using `write` to only
modify the default way the printing to the console
looks rather than writing to an alternate
destination entirely.

The name of the option and its behavior is obviously open for discussion

Part of #1556

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -109,6 +109,7 @@ function pino (opts) {
     transmit,
     serialize,
     asObject: opts.browser.asObject,
+    asObjectBindingsOnly: opts.browser.asObjectBindingsOnly,
     formatters: opts.browser.formatters,
     levels,
     timestamp: getTimeFunction(opts),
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):

mcollina:
Can. you add some docs?

segevfiner:
Can you point me where to add them? Also do consider the naming and the exact semantics of this API, see the issues I opened for what I'm trying to eventually achieve.

mcollina:
https://github.com/pinojs/pino/blob/main/docs/browser.md

I think the name is ok.

jsumners (commented):
Is this applicable to the core logger?

segevfiner:
What do you mean? It's a variation of a flag that only exists in the browser version, where the output interface is different.

mcollina (approved):
lgtm

jsumners (approved):

Morriz:
@mcollina has this not been released yet? Its been over 3 months already, but its not available in `9.7.0` it seems.

PR #2052, opened 2024-09-29
fix: Ensure silent log level is respected with browser.transmit option

SrsDanny:
The silent log level currently does not work in the browser when the `browser.transmit` option is used.

- When `options.level` is set to `"silent"` and `options.browser.transmit` is used, everything gets logged as if `options.level` was the most verbose.
- When `options.level` is non-silent (e.g. `"info"`) and `options.browser.transmit.level` is `"silent"`, the `options.browser.transmit.send` method is called for every log event above `options.level`, which should not occur.

This PR uses the `levelToValue` method which correctly handles the silent level to determine the `transmitValue` and `methodValue`, so the silent level works even when `options.browser.transmit` is used.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -280,8 +280,8 @@ function set (self, opts, rootLogger, level) {
     if (!opts.transmit) return
 
     const transmitLevel = opts.transmit.level || self.level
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #2034, opened 2024-08-22
feat: accept level in capitals

balaji-atoa:
### 💬**BASED ON**:
https://github.com/pinojs/pino/issues/2022#issuecomment-2304332920

### 🚀**UPDATES**: 
- converted level to lower case if passed in capitals.

✅ Added UnitTests.

PR #2056, opened 2024-10-07
fix: Lowercase the custom levels only when useOnlyCustomLevels is true

ondrej-langr:
Typescript enums usually have uppercased keys. [This](https://github.com/pinojs/pino/pull/2034/files) change completely breaks implementation when the `useOnlyCustomLevels` option is set to `true` and the enum of custom levels includes `DEBUG` key.

### DIFF
--- a/pino.js
+++ b/pino.js
@@ -90,12 +90,11 @@ function pino (...args) {
   const { opts, stream } = normalize(instance, caller(), ...args)
 
   if (
-    opts.level && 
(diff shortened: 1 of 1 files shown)


mcollina:
CI is failing

ondrej-langr:
@mcollina hey, can you please look at this when you have some time? Thanks

mcollina (changes requested):

Issue #1801, opened 2023-09-08
Pino will not print logs when the event loop is blocked, but console.log will


vedantroy:
I'm running into an odd error where Pino is not printing logs unless an exception is thrown.

Here's my config:
```
const opts: PinoOpts = {
  redact: {
    // these just cause clutter
    paths: ["pid", "hostname"],
    remove: true,
  },
  transport: {
    targets: [],
  },
}
opts.transport.targets.push({
  level: "trace",
  target: "pino-pretty",
  options: {
  },
})

const logger = pino(opts)

console.log("BEFORE ...")
logger.info("SOME LOG MESSAGE ....")
console.log("AFTER ...")

// if I comment the below line, I don't see any logs
throw new Error("hi")
```
What could be be going on? Is there some sort of buffering?

vedantroy:
Ok, I figured out the issue. When the event-loop is almost blocked because there's an infinite while-loop, pino will not print anything. The problem is -- this is quite bad, because it is precisely when the event loop is blocked that I don't want my logs to go silent.

I'd rather sacrifice lots of performance, then have pino not prevent logs when the event loop is blocked. Example:

mcollina:
Regarding `console.log`: that is true only if you are logging to a TTY. If you are logging to another process, or a file, it would be asynchronous as well, with exactly the same behavior (at some point the process will stop logging and will crash with out of memory).

Regarding transports, I actually think we can support this behavior by allowing users to set `sync: true` in [`ThreadStream`](https://github.com/pinojs/pino/blob/619e049aed18dd79a9fdfac5dceab86c34e9bda0/lib/transport.js#L21) [constructor](https://github.com/pinojs/thread-stream#usage). I actually added that with this specific purpose long ago but in the end I didn't feel the need. If you want to send a PR, it would be awesome.

This is generically a tradeoff between various concerns and it really depends on your application. The defaults here are generically the best tradeoff we could came up with. 
You can read more about this topic in https://adventures.nodeland.dev/archive/the-dangers-of-logging-and-other-adventures/.

karankraina:
@mcollina Since the PR is now merged, should this be closed ?

PR #2063, opened 2024-10-13
feature: allow sync flag to be passed to ThreadStream for synchronous…

karankraina:
Allow `sync` flag to be passed to ThreadStream for sync logging.
Ref - #1801

### DIFF
--- a/test/transport/syncTrue.test.js
+++ b/test/transport/syncTrue.test.js
@@ -1,0 +1,55 @@
+'use strict'
+
+const pino = require('../..')
+const { join } = require('node:path')
+const { readFileSync } = require('node:fs')
+const { test } = require('tap')
+const { file } = require('../helper')
+
+test('thread-stream sync true should log synchronously', async (t) => {
+  const outputPath = file()
+
+  function getOutputLogLines () {
+    return (readFileSync(outputPath)).toString().trim().split('\n').map(JSON.parse)
+  }
+
+  const transport = pino.transport({
+    target: join(__dirname, '..', 'fixtures', 'to-file-transport.js'),
+    options: { destination: outputPath },
+    sync: true
+  })
+  const instance = pino(transport)
+
+  var value = { message: 'sync' }
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  let interrupt = false
+  let flushData
+  let loopCounter = 0
+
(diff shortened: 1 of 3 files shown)


karankraina:
@mcollina Do you think this works ?

mcollina (changes requested):

karankraina:
@mcollina Added unit test.

mcollina (approved):
lgtm

test/transport/syncTrue.test.js
```diff
@@ -0,0 +1,55 @@
...
+  })
+  const instance = pino(transport)
+
+  var value = { message: 'sync' }
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
+  instance.info(value)
```
mcollina, line 29: I think you could do `instance.flushSync()` and then just try reading the file. It should work.

karankraina: Wouldn't calling `flushSync` force synchronously flush all messages to the stream ? That would pass even if `sync` is `false` ? I tried to simulate a event loop block immediately after calling `instance.info`. I think I need to update the loop duration by some bit. Am i missing something?

mcollina: ah no prob. This is ok

mcollina:
You might need to pass `flush: true` to https://nodejs.org/api/fs.html#filehandlecreatewritestreamoptions in https://github.com/pinojs/pino/blob/f220965c3e3885b8479c37ec880e43315e1a2373/test/fixtures/to-file-transport.js#L8. Otherwise Node.js would not `fsync` the file descriptor, so you might not be able to read it.

karankraina:
@mcollina Ah yess, updated PR. `flush: true` fixed it.  Added loop duration by a bit just in case.

test/transport/syncTrue.test.js
karankraina: Updated PR. Should pass now :)

karankraina:
> @mcollina Ah yess, updated PR. `flush: true` fixed it. Added loop duration by a bit just in case.

Ubuntu 22 failing for some reason. :( Need to re-check this. Had worked on my fork

test/transport/syncTrue.test.js
karankraina: @mcollina can you retrigger the CI actions please

Issue #1991, opened 2024-06-15
`stdSerializers` cannot be avoided for Request and Response


gerardolima:
When any logger method (ie `info`) receives a _Request_, pino serializes the object using `pino.stdSerializers.req` and stores in `req` property; this value is, then, passed to the actual _serializer_ function given by the options. This also happens to _Response_ objects:

```ts
const opts = {
  serializers: { 
    'req': mySerializerReq,
    'res': mySerializerRes,
}
const logger = pino(opts)
logger.info(myReq) // serialized as `mySerializerReq(pino.stdSerializers.req(myReq))`
logger.info(myRes) // serialized as  `mySerializerRes(pino.stdSerializers.res(myRes))`
```

This is inefficient, performance-wise, and also can lead to _unexpected behaviour_, when the provided serializer requires some property from the object that is removed by `pino.stdSerializers.*`.

gerardolima:
I tested using this script, below:
```ts
import {describe, it} from 'node:test'
import assert from 'node:assert/strict'
import pino from 'pino';

const makeWriter = () => {
  const acc = [];
  return {
    write: (chunk) => { acc.push(chunk); },
    entries: () => [...acc],
  };
};

describe('stdSerializers', () => {

  it('serializes `.mySpecialProperty` for Errors', () => {
    // this test is OK
     
    // `err` property is resolved at `pino/lib/proto.js:write`
    // see https://github.com/pinojs/pino/blob/8db130eba0439e61c802448d31eb1998cebfbc98/lib/proto.js#L191

    const myErr = new Error()
    myErr.mySpecialProperty = Math.random()

    const writer = makeWriter()
    const opts = { serializers: { 'err': obj => ({ mySpecialProperty: obj.mySpecialProperty }) } }
    const logger = pino(opts, writer)

    logger.info(myErr)

    const entries = writer.entries()
    assert.equal(entries.length, 1)

    const obj = JSON.parse(entries[0])
    assert.deepEqual(obj.err, {mySpecialProperty: myErr.mySpecialProperty})
  })

  it('serializes `.mySpecialProperty` for Requests', () => {
    // this test fails

    // error property is resolved at `pino/lib/tools.js:genLog:LOG` -> `pino-std-serializers/index.d.ts:mapHttpRequest`
    // see https://github.com/pinojs/pino/blob/8db130eba0439e61c802448d31eb1998cebfbc98/lib/tools.js#L44
[...]

PR #2081, opened 2024-11-09
feat(serializers): avoid implicit sanitization

gerardolima:
When serializers are defined for HTTP Request or HTTP Response, do not run the correspondent `stdSerializers` before calling the provided serializer functions.

ISSUE  #1991


lib/proto.js
```diff
@@ -193,6 +198,10 @@ function write (_obj, msg, num) {
     if (msg === undefined) {
       msg = _obj.message
     }
+  } else if (_obj instanceof IncomingMessage) {
+    obj = { [requestKey]: _obj }
+  } else if (_obj instanceof ServerResponse) {
+    obj = { [responseKey]: _obj }
```
gerardolima, line 204: If preferred, we can assess _type_ from properties, as previously done in `lib/tools.js`:
```js
  } else if (_obj.method && _obj.headers && _obj.socket) {
    obj = { [requestKey]: _obj }
  } else if (typeof _obj.setHeader === 'function') {
    obj = { [responseKey]: _obj }
```

lib/tools.js
```diff
@@ -132,6 +129,10 @@ function asJson (obj, msg, num, time) {
         value = serializers[key](value)
       } else if (key === errorKey && serializers.err) {
         value = serializers.err(value)
+      } else if (key === requestKey && serializers.req) {
+        value = serializers.req(value)
+      } else if (key === responseKey && serializers.res) {
+        value = serializers.res(value)
```
gerardolima, line 135: Behaviour of all `stdSerializers` follow the same pattern.

pino.d.ts
```diff
@@ -408,6 +408,14 @@ declare namespace pino {
...
          */
         errorKey?: string;
+        /**
+         * The string key for the 'Request' in the JSON object. Default: "req".
+         */
+        requestKey?: string;
+        /**
+         * The string key for the 'Response' in the JSON object. Default: "res".
+         */
+        responseKey?: string;
```
gerardolima, line 418: The key for request and response are now explicitly configurable. Previously, these had fixed values, set on pino-std-serializers ([mapHttpRequest](https://github.com/pinojs/pino-std-serializers/blob/master/lib/req.js#L98) and [mapHttpResponse](https://github.com/pinojs/pino-std-serializers/blob/master/lib/res.js#L45)).

test/http.test.js
```diff
@@ -160,34 +154,31 @@ test('http response support', async ({ ok, same, error, teardown }) => {
   server.close()
 })
 
-test('http response support via a serializer', async ({ ok, same, error, teardown }) => {
+test('http response support via a serializer', async ({ match, error }) => {
```
mcollina, line 157: can you please keep the previous test and only add a new one?

gerardolima: Original tests restored. Also fixed linter issues.

pino.d.ts
gerardolima: maybe this change makes the exported functions `mapHttpRequest` and `mapHttpResponse` deprecated, in [pino-std-serializers](https://github.com/pinojs/pino-std-serializers)?

gerardolima:
ok, I was running isolated tests and only now I noticed there were some assertions on the default serialisers; I am fixing them right now and soon the tests will be fixed; sorry for the silly mistake

mcollina (approved):
lgtm

lib/proto.js
jsumners: I think property presence will be better here. These _could_ be extended objects, e.g. a `FastifyRequest`.

gerardolima: done

gerardolima:
> The issue mentioned deprecating the current implementation and switching to the new one in the next major. I don't see this being implemented in this PR. Have I missed it?

This changeset only modifies the following behviour: neither logged _Request_ nor _Response_ objects are not transformed by their corresponding `stdSerializers` unconditionally, at the start of `LOG` ([lib/tools.js](https://github.com/pinojs/pino/pull/2081/files#diff-7f88797c7db38a6af7a3283e3af30f2cdd3c2489bd9c85a7a25b32ec2485f724L43)).

This change in behaviour is asserted in test/http.test.js, lines [86](https://github.com/pinojs/pino/pull/2081/files#diff-07b203d934292ad265e6f428d9aaaf287e55032e8f497cc0c030d276a1dea245R86) and [240](https://github.com/pinojs/pino/pull/2081/files#diff-07b203d934292ad265e6f428d9aaaf287e55032e8f497cc0c030d276a1dea245R86), with tests that create an arbitrary property that should be removed from them, if their standard serializers ran.

jsumners:
> This changeset only modifies the following behviour: neither logged _Request_ nor _Response_ objects are not transformed by their corresponding `stdSerializers` unconditionally, at the start of `LOG` ([lib/tools.js](https://github.com/pinojs/pino/pull/2081/files#diff-7f88797c7db38a6af7a3283e3af30f2cdd3c2489bd9c85a7a25b32ec2485f724L43)).
> 
> This change in behaviour is asserted in test/http.test.js, lines [86](https://github.com/pinojs/pino/pull/2081/files#diff-07b203d934292ad265e6f428d9aaaf287e55032e8f497cc0c030d276a1dea245R86) and [240](https://github.com/pinojs/pino/pull/2081/files#diff-07b203d934292ad265e6f428d9aaaf287e55032e8f497cc0c030d276a1dea245R86), with tests that create an arbitrary property that should be removed from them, if their standard serializers ran.

This is a breaking change, correct?

gerardolima:
> This is a breaking change, correct?

Yes, this is correct. When custom serializers _are_ provided and they depend on the changes made by the standard serializers on the target objects (ie: to _sanitize_ them?), they will be impacted.

Though running standard serializers when custom serializers are given is not exaclty _obvious behaviour_ -- and I think nor documented --, I understand there might be code in the wild that dependes on it ([Hyrum's Law](https://www.hyrumslaw.com/)) and publishing this as _breaking change_ should be the safest strategy.

jsumners:
> Yes, this is correct. When custom serializers _are_ provided and they depend on the changes made by the standard serializers on the target objects (ie: to _sanitize_ them?), they will be impacted.

Then we are missing the deprecation and opt-in for the current version.

gerardolima:
> Then we are missing the deprecation and opt-in for the current version.

I'm sorry, but I'm not familiar with the versioning process in Pino -- I've already checked [CONTRIBUTING.md](https://github.com/pinojs/pino/blob/main/CONTRIBUTING.md). Would it suffice updating major version in [package.json](https://github.com/pinojs/pino/blob/main/package.json#L3C4-L3C11)?  Could you, please help me?

jsumners:
1. https://github.com/pinojs/pino/issues/1991#issuecomment-2243151339 suggested that the original functionality be deprecated. That means the current behavior _does not change_, but will generate a warning when the functionality is used.
2. This PR changes the published functionality, thus introducing a breaking change and necessitating a new major version to be published after it is merged.
3. We avoid needing to publish a new major version by deprecating the current functionality and providing an opt-in path to the functionality provided in this PR.
4. We utilize https://www.npmjs.com/package/process-warning to issue deprecation warnings. I provided links to _historical_ code where we have issued deprecation warnings so that you can learn how to implement such a deprecation in this feature. A clearer example may be seen in https://github.com/pinojs/pino/commit/6c538159dc83b190889781f80051962138714bb8. The next available deprecation code is currently commented out in the deprecations source file (PINODEP010).

gerardolima:
hey, @jsumners, the new example is super clear about showing the warning message, thank you! About the `opt-in` flow, the idea should be to create some new property in options to act like a feature-flag / pragma?

gerardolima:
hey, @mcollina, this PR got stalled and outdated. I can rebase my branch and fix the conflicts, but first I'd like to understand whether you are still interested in changing the behaviour of `stdSerializers` for _Request_ and _Response_. A `no` response is perfectly fine :)

jsumners:
Sorry for the delay. I have it on my priority list to address Pino issues first when I am in a position to work on OSS things.

gerardolima:
hey, @mcollina and @jsumners, I rebased the branch and resolved conflicts; this PR is ready for review again:

Summary of changes:
1. Add property `future` to the constructor _options_. This property enables _opt-in_ behaviour changes (all keys are `false` by default);
2. When logging a `Request` or `Response` object, depending on `future.skipUnconditionalStdSerializers`:
    - when `false`
        1. Behaviour is not changed (based on `mapHttpRequest`, and `mapHttpResponse`); and
        2. The warning [PINODEP010](https://github.com/pinojs/pino/pull/2081/files#diff-57e657ba6573b3d7f37a78b90f247e13bc16ae1849b3188e123f2d879a4c648eR14) is emmitted;
    - when `true`
        1. Behaviour is changed and value is serialized either by the corresponding `stdSerializer` or by a serializer provided in constructor _options_ (neither `mapHttpRequest`, nor `mapHttpResponse` is executed).

mcollina (approved):
lgtm

mcollina:
Can you drop the `future` prefix?

gerardolima:
> Can you drop the `future` prefix?

sure, @mcollina, but how should I guard this breaking change not to force bumping a major version, as @jsumners pointed? Is it just a matter of naming? Should I use, some `"opt-in"`, instead of `"future"` option?

gerardolima:
hey, @mcollina and @jsumners, just a friendly reminder that this PR is half-an-year old, already; please let me know whether this is still relevant to you and whether you want any change 🙏🏼

gerardolima:
hello again, @mcollina and @jsumners, it's been a while since we've last spoken; I understand that you folks are quite busy, but I wonder whether it still makes sense to keep this PR open.

PR #2112, opened 2024-12-23
test: skip `pkg` test in CITGM

targos:
CI hosts are not guaranteed to be able to run `pkg` correctly. See for example https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/3527/nodes=osx11/testReport/junit/(root)/citgm/pino_v9_6_0/


mcollina (approved):
lgtm

PR #2136, opened 2025-02-25
docs: add hono example

jansepke:
example on how pino could be used with the hono framework

### DIFF
--- a/docs/web.md
+++ b/docs/web.md
@@ -12,6 +12,7 @@ web framework ecosystem.
   - [Pino with Node core `http`](#pino-with-node-core-http)
   - [Pino with Nest](#pino-with-nest)
   - [Pino with H3](#pino-with-h3)
+  - [Pino with Hono](#pino-with-hono)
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

I think we should end up creating a pino-hono library at some point.

docs/web.md
```diff
@@ -267,3 +268,49 @@ router.get(
...
+import { Hono } from 'hono';
+import { requestId } from 'hono/request-id';
+import { pino } from 'pino';
+import { pinoHttp } from 'pino-http';
+
+declare module 'hono' {
+  interface ContextVariableMap {
+    logger: pino.Logger;
+  }
+}
```
jsumners, line 291: Which version of JavaScript is this? I haven't ever seen such.

jansepke:
I opened an issue at hono to discuss if they want to make integration of a logger easier or if every logger should add a dedicated library for hono https://github.com/honojs/hono/issues/3963

PR #2146, opened 2025-03-11
build(dependabot): reduce npm updates to monthly

Fdawgs:
Same reasoning as https://github.com/fastify/fastify/pull/5939.


mcollina (approved):
lgtm

PR #2161, opened 2025-04-28
test: move browser-child.test.js to node test runner

Puppo:
This PR moves `browser-child.test.js` to Node test runner

### DIFF
--- a/test/browser-child.test.js
+++ b/test/browser-child.test.js
@@ -1,8 +1,8 @@
 'use strict'
-const test = require('tape')
+const test = require('node:test')
 const pino = require('../browser')
 
-test('child has parent level', ({ end, same, is }) => {
+test('child has parent level', (t, end) => {
   const instance = pino({
     level: 'error',
     browser: {}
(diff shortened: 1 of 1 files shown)


test/browser-child.test.js
```diff
@@ -1,32 +1,32 @@
 'use strict'
-const test = require('tape')
+const test = require('node:test')
```
jsumners, line 2: I don't know if this is what we want to do. I think `tape` is used for some browser reason. @mcollina do you know?

jsumners (changes requested):

test/browser-child.test.js
Puppo: @jsumners @mcollina, any updates on this one?

mcollina: Yes I would keep tape for this is possible. This is run via https://github.com/pinojs/pino/blob/a42487988dfa2cf4624e38a46673657bd9772a1f/package.json#L11. But I'm not sure this has been exercised in a long time.

Puppo: Perfect, I'll close this pr

PR #2159, opened 2025-04-27
Include pino-roll in Pino transports list​

mustakimkr:
This PR adds documentation for the `pino-roll` transport in `docs/transports.md`, as requested in issue #2153 (see https://github.com/pinojs/pino/issues/2153#issuecomment-2833391961)

PR #2167, opened 2025-04-29
Fix order of pino-roll

jsumners:
Fixes the order of `pino-roll` from #2159. The sections looked out of order in the GitHub interface, but they were actually correct except for the new one. This fixes it.


mcollina (approved):
lgtm

PR #2184, opened 2025-05-03
test: formatters.test.js to the node test runner

Puppo:
This PR moves `formatters.test.js` to the node test runner

### DIFF
--- a/test/formatters.test.js
+++ b/test/formatters.test.js
@@ -4,11 +4,20 @@
 const { hostname } = require('node:os')
 const { join } = require('node:path')
 const { readFile } = require('node:fs').promises
-const { test } = require('tap')
+const { test } = require('node:test')
 const { sink, once, watchFileCreated, file } = require('./helper')
 const pino = require('../')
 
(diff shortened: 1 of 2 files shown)


test/formatters.test.js
```diff
@@ -4,11 +4,20 @@
 const { hostname } = require('node:os')
 const { join } = require('node:path')
 const { readFile } = require('node:fs').promises
-const { test } = require('tap')
+const { test } = require('node:test')
 const { sink, once, watchFileCreated, file } = require('./helper')
 const pino = require('../')
 
-test('level formatter', async ({ match }) => {
+function match (t, obj, expected) {
```
jsumners, line 11: Subsequent to the other PR being updated with `@jsumners/assert-match` and being merged, this PR needs to be updated to use that assertion function.

jsumners (changes requested):

test/formatters.test.js
Puppo: yep, done!

jsumners (approved):

PR #2185, opened 2025-05-03
test: move hooks.test.js to the node test runner

Puppo:
This PR moves `hooks.test.js` to the node test runner


test/hooks.test.js
```diff
@@ -1,25 +1,34 @@
...
-tap.test('log method hook', t => {
-  t.test('gets invoked', async t => {
+function match (t, obj, expected) {
+  const checkObj = Object.keys(expected).reduce((acc, key) => {
+    acc[key] = obj[key]
+    return acc
+  }, {})
+
+  t.assert.deepStrictEqual(checkObj, expected)
+}
```
jsumners, line 14: It's not available in v20 yet, but I would prefer we use https://github.com/nodejs/node/pull/54630. In the interim, we can copy https://github.com/newrelic/node-newrelic/blob/ebf3aa946be1e6802b567f2bfeef0b0c13613375/test/lib/custom-assertions/match.js into a file like `test/assert-match.js`. I'm sure it will be used in more suites than this one.

jsumners (changes requested):

test/hooks.test.js
Puppo: @jsumners done. There are other PR that need this `match` assertion. I'll wait for this one to be merged before refactoring them.

jsumners (approved):

jsumners (changes requested):
Aw hell. I just realized the match function is licensed Apache-2.0 and Pino is MIT. I think we can include it, but we might need to update the licenses file. @mcollina do you know?

jsumners:
That's a fair solution. I can do it with my account after work.

jsumners:
@Puppo please use https://www.npmjs.com/package/@jsumners/assert-match

```sh
npm install -D @jsumners/assert-match
```

Puppo:
> @Puppo please use https://www.npmjs.com/package/@jsumners/assert-match
> 
> ```shell
> npm install -D @jsumners/assert-match
> ```

Done!

jsumners (approved):

PR #2188, opened 2025-05-03
test: levels.test.js to node test runner

Puppo:
This PR moves `levels.test.js` to the node test runner

### DIFF
--- a/test/levels.test.js
+++ b/test/levels.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { test, describe } = require('node:test')
+const { test } = require('node:test')
 const { sink, once, check } = require('./helper')
 const pino = require('../')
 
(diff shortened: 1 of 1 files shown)


test/levels.test.js
```diff
@@ -519,15 +525,15 @@ test('changing level from info to silent and back to info in child logger', asyn
...
 
   child.level = 'info'
   child.info('hello world')
   result = await once(stream, 'data')
-  check(equal, result, expected.level, expected.msg)
+  check(t.assert.strictEqual, result, expected.level, expected.msg)
 })
 
-test('changing level respects level comparison set to', async ({ test, end }) => {
+describe('changing level respects level comparison set to', async () => {
```
jsumners, line 536: I would prefer that we do not use `describe`. I'd rather see:

```js
test('something', async t => {
  await t.test('another thing', t => t.equal(1, 1))
})
```

jsumners (changes requested):

jsumners (approved):

PR #2192, opened 2025-05-05
test: move multistream.test.js to node test

Puppo:
This PR moves `multistream.test.js` to node test

### DIFF
--- a/test/multistream.test.js
+++ b/test/multistream.test.js
@@ -3,14 +3,14 @@
 const writeStream = require('flush-write-stream')
 const { readFileSync } = require('node:fs')
 const { join } = require('node:path')
-const test = require('tap').test
+const test = require('node:test').test
 const pino = require('../')
 const multistream = pino.multistream
 const proxyquire = require('proxyquire')
 const strip = require('strip-ansi')
 const { file, sink } = require('./helper')
 
-test('sends to multiple streams using string levels', function (t) {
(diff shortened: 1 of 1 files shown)


test/multistream.test.js
```diff
@@ -29,11 +29,11 @@ test('sends to multiple streams using string levels', function (t) {
   log.info('info stream')
   log.debug('debug stream')
   log.fatal('fatal stream')
-  t.equal(messageCount, 9)
-  t.end()
+  t.assert.strictEqual(messageCount, 9)
+  end()
```
jsumners, line 33: I don't think we need `end()` in these tests. With `tap`, we needed `t.end` because we didn't set forth a plan or return a promise. It looks to me like (at least most of) the tests in this suite are synchronous. `node:test`, at its most basic, only requires that a test not throw an error and the code of the test be fully synchronous.

jsumners (changes requested):

Puppo:
@jsumners I've also simplified two tests using `t.assert.throw` https://github.com/pinojs/pino/pull/2192/commits

test/multistream.test.js
jsumners: I left this comment on the first synchronous test, but there are still many others. Please update them as well, @Puppo.

Puppo: @jsumners done! I removed `end` where possible.

jsumners (approved):

PR #2198, opened 2025-05-05
test: move broken-pipe.test.js to node test

Puppo:
This PR moves `broken-pipe.test.js` to node test

### DIFF
--- a/test/broken-pipe.test.js
+++ b/test/broken-pipe.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const t = require('tap')
+const t = require('node:test')
 const { join } = require('node:path')
 const { fork } = require('node:child_process')
 const { once } = require('./helper')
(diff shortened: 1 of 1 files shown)


test/broken-pipe.test.js
```diff
@@ -52,6 +52,8 @@ t.test('let error pass through', ({ equal, plan }) => {
   process.nextTick(() => stream.emit('error', new Error('kaboom')))
 
   stream.on('error', (err) => {
-    equal(err.message, 'kaboom')
+    t.assert.strictEqual(err.message, 'kaboom')
   })
+
+  process.nextTick(end)
```
Puppo, line 58: Without `end`, node test exists before the end.

jsumners: Can we put it in `stream.on('end')`?

jsumners (changes requested):

test/broken-pipe.test.js
Puppo: Nope, I got this error.

Screenshot 2025-05-06 at 19 00 56

jsumners: I suspect `nextTick` is going to be flaky. I think we should:

```js
test('thing', async t => {
	const { promise, resolve } = Promise.withResolvers()
	const stream = createStream()

	stream.on('end', () => resolve())

	await promise
})
```

Puppo: @jsumners, this is the only solution I found that works.
If I don't call the `stream.end()`, the promise is still pending, and the `close` event won't called.
Let me know if you have another solution in mind that I didn't catch.

jsumners (approved):
I didn't realize `Promise.withResolvers` is not available in Node.js 20. This solution looks good to me.

PR #2201, opened 2025-05-05
test: move big.test.js to node test

Puppo:
This PR moves `big.test.js` to node test

### DIFF
--- a/test/transport/big.test.js
+++ b/test/transport/big.test.js
@@ -32,9 +32,6 @@ test('eight million lines', { skip }, async (t) => {
   let count = 0
   await pipeline(createReadStream(destination), split(), new Writable({
     write (chunk, enc, cb) {
-      if (count % (toWrite / 10) === 0) {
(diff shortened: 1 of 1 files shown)


test/transport/big.test.js
```diff
@@ -33,11 +33,11 @@ test('eight million lines', { skip }, async ({ equal, comment }) => {
   await pipeline(createReadStream(destination), split(), new Writable({
     write (chunk, enc, cb) {
       if (count % (toWrite / 10) === 0) {
-        comment(`read ${count}`)
+        t.diagnostic(`read ${count}`)
```
jsumners, line 36: Unfortunately, `t.diagnostic` does not work the same way as `tap.comment`. It only reports things at the end of a test run. We should just remove this `if` block.

jsumners (changes requested):

jsumners (approved):

PR #2214, opened 2025-05-08
test: move core.test.js to node test

Puppo:
This PR moves `core.test.js` to node test

### DIFF
--- a/test/transport/core.test.js
+++ b/test/transport/core.test.js
@@ -294,7 +294,7 @@ test('pino.transport with an array including a pino-pretty destination', async (
     msg: 'hello'
   })
   const actual = (await readFile(dest2)).toString()
-  match(strip(actual), /\[.*\] INFO.*hello/, t)
+  t.assert.match(strip(actual), /\[.*\] INFO.*hello/)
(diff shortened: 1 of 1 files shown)


test/transport/core.test.js
```diff
@@ -280,23 +281,23 @@ test('pino.transport with an array including a pino-pretty destination', async (
...
-  same(result1, {
+  t.assert.deepStrictEqual(result1, {
     pid,
     hostname,
     level: 30,
     msg: 'hello'
   })
   const actual = (await readFile(dest2)).toString()
-  match(strip(actual), /\[.*\] INFO.*hello/)
+  match(strip(actual), /\[.*\] INFO.*hello/, t)
```
jsumners, line 297: I believe we can use the standard `t.assert.match` here. This is matching a regular expression against a string.

jsumners (changes requested):

test/transport/core.test.js
Puppo: Done

jsumners (approved):

PR #2216, opened 2025-05-08
test: move transport-stream.test.js to node test

Puppo:
This PR moves `transport-stream.test.js` to node test


test/transport-stream.test.js
```diff
@@ -1,14 +1,15 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
+const { mock } = require('cjs-mock')
```
Puppo, line 4: I installed `cjs-mock` because mocking in the node test runner is experimental.

jsumners: I pushed a commit. I prefer dependency injection for such things.

Puppo: LGTM!

jsumners (approved):

PR #2219, opened 2025-05-09
test: move pkg/index.js to node test

Puppo:
This PR moves `pkg/index.js` to node test

### DIFF
--- a/test/pkg/index.js
+++ b/test/pkg/index.js
@@ -4,7 +4,7 @@ const os = require('node:os')
 const { join } = require('node:path')
 const { readFile } = require('node:fs').promises
 const { watchFileCreated, file } = require('../helper')
-const { test } = require('node:test')
+const assert = require('node:assert/strict')
 const pino = require('../../pino')
 
 const { pid } = process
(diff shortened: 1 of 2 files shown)


test/pkg/index.js
Puppo, line 1: pkg cannot resolve `node:test`.
This is the best idea that came to my mind.
Please let me know what you think.

test/pkg/pkg.config.json
```diff
@@ -7,10 +7,10 @@
             "../../node_modules/pino-abstract-transport/index.js"
         ],
         "targets": [
-            "node14",
```
Puppo, line 10: Remove node14 because it doesn't support `node:assert/strict`.
I don't know if we want to keep node16 and node18 or remove them.

jsumners: Remove them and add 24.

Puppo: Removed them, but node24 isn't supported yet.
Screenshot 2025-05-09 at 14 09 35

jsumners (approved):

Issue #2120, opened 2025-01-23
Types issue with ESM setup


iamchathu:
The types seems broken in ESM Typescript setup with module resolution set to "node16". 

Pino is `commonjs` and normally should work with the project without issues.

Here is the reproduction minimal setup.

https://stackblitz.com/edit/stackblitz-starters-kuphd3r7?file=tsconfig.json,src%2Findex.ts

mcollina:
I think this would do the trick:

```
import pino from 'pino';

const logger = pino({});
```

iamchathu:
Still the error is the same

```sh
Type 'typeof import("/home/projects/stackblitz-starters-kuphd3r7/node_modules/pino/pino")' has no call signatures.

3 const logger = pino({});
                 ~~~~


Found 1 error in src/index.ts:3

```

alexgleason:
Agreed, see: https://arethetypeswrong.github.io/?p=pino%409.6.0

![Image](https://github.com/user-attachments/assets/4adbe802-6abd-4774-92cf-ebe76e765e61)

More info: https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/MissingExportEquals.md

colinmollenhour:
So is this currently the only/correct way to use it without throwing TS errors?

```
import pino from 'pino';

const logger = pino.default({});
```

Tobbe:
I'm building for both CJS and ESM, so I had to do `const logger = (pino.default || pino)({});` to both please TS and to make both the outputs work.

cluah:
you can also manually `import { pino } from 'pino'`

but more people will end up here as that's not the default VSCode auto-import behavior

PR #2223, opened 2025-05-19
Fix TS export condition

mcollina:
Fixes https://github.com/pinojs/pino/issues/2120

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -739,26 +739,16 @@ declare namespace pino {
     //// Top level variable (const) exports
 
 
-}
+    /**
+     * Provides functions for serializing objects common to many projects.
+     */
+    export const stdSerializers: typeof pinoStdSerializers;
 
-//// Callable default export
-
-declare function pino<CustomLevels extends string = never, UseOnlyCustomLevels extends boolean = boolean>(optionsOrStream?: pino.LoggerOptions<CustomLevels, UseOnlyCustomLevels> | pino.DestinationStream): pino.Logger<CustomLevels, UseOnlyCustomLevels>;
-declare function pino<CustomLevels extends string = never, UseOnlyCustomLevels extends boolean = boolean>(options: pino.LoggerOptions<CustomLevels, UseOnlyCustomLevels>, stream?: pino.DestinationStream | undefined): pino.Logger<CustomLevels, UseOnlyCustomLevels>;
-
-// Merge static properties and functions into the pino function via namespace merging
-declare namespace pino {
-    // Constants and functions - make them available as static properties
-    const version: string;
-    const levels: LevelMapping;
-    const stdSerializers: typeof pinoStdSerializers;
-    const stdTimeFunctions: {
-        epochTime: TimeFn;
-        unixTime: TimeFn;
-        nullTime: TimeFn;
-        isoTime: TimeFn;
-    };
-    const symbols: {
+    /**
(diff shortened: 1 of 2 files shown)


mcmxcdev:
Any change of getting this finished and merged? Only dependency blocking us from upgrading to ESM in our monorepo.

mcollina:
Can you test this PR @mcmxcdev ?

mcmxcdev:
> Can you test this PR @mcmxcdev ?

Sure thing, I will try to have a look later today!

mcmxcdev:
I was actually able to test this out quickly by copying over `pino.d.ts` from this PR to the project I am working on.

Looks great, no more `This expression is not callable. Type 'typeof import(".../node_modules/.pnpm/pino@9.7.0/node_modules/pino/pino")' has no call signatures.` errors!

pino.d.ts
```diff
@@ -838,66 +839,15 @@ declare namespace pino {
...
-export interface TransportBaseOptions extends pino.TransportBaseOptions {}
-export interface TransportMultiOptions extends pino.TransportMultiOptions {}
-export interface TransportPipelineOptions extends pino.TransportPipelineOptions {}
-export interface TransportSingleOptions extends pino.TransportSingleOptions {}
-export interface TransportTargetOptions extends pino.TransportTargetOptions {}
-
-// Bundle all top level exports into a namespace, then export namespace both
-// as default (`import pino from "pino"`) and named variable
-// (`import {pino} from "pino"`).
-export { pino as default, pino };
```
UncleGrim, line 899: I'd like to understand why this line of code was not treated as a major version bump as it causes a breaking change. 

The exported interfaces and types above are problematic as well, but this line breaks projects that are currently importing pino like:

```
import { pino } from "pino";
```

Resulting in: 

`error TS2595: 'pino' can only be imported by using a default import.`

noudadrichem: ➕1 on this one. It's breaking our setup.

We have a helper wrapping Pino and updating from 9.7.0 to 9.8.0 breaks. When changing it from named to default import the following error is thrown when calling `pino()`

```
This expression is not callable.
  Type 'typeof import("/Developer/proj/node_modules/pino/pino")' has no call signatures.ts(2349)
```

This is our helper wrapping Pino.
```ts
import pino, { Logger, LoggerOptions } from "pino";
import { ENVIRONMENT } from "...";

export class LoggerHelper {
    // Ref; Static so it can be used by Fastify.
    public static config: LoggerOptions;

    private readonly logger: Logger;
    private readonly level: string;

    constructor() {
        this.level = ENVIRONMENT.LOG_LEVEL;

        LoggerHelper.config = {
            level: this.level,
            timestamp: pino.stdTimeFunctions.isoTime,
        };

        if (ENVIRONMENT.NODE_ENV === "local") {
            LoggerHelper.config.transport = {
                target: "pino-pretty",
                options: {
                    translateTime: "HH:MM:ss",
                    ignore: "pid,hostname",
                    colorize: true,
                },
            };
        }

        this.logger = pino(LoggerHelper.config); // <-- This call is what errors out
    }

    info(message: string, data?: Record<string, unknown>) {
        this.logger.info(data, message);
    }

    error(message: string, data: Record<string, unknown> | Error) {
        this.logger.error(data, message);
    }

    debug(message: string, data: Record<string, unknown>) {
        this.logger.debug(data, message);
    }
}
```

PR #2224, opened 2025-05-25
docs: update transports.md

marklai1998:
`pino-datadog-transport` is no longer maintained

Ref: https://github.com/theogravity/datadog-transports

I've created a separate package to maintain it forward, with extra features like batching and proper AWS Lambda support

### DIFF
--- a/docs/transports.md
+++ b/docs/transports.md
@@ -451,7 +451,6 @@ PRs to this document are welcome for any new transports!
 + [pino-cloudwatch](#pino-cloudwatch)
 + [pino-couch](#pino-couch)
 + [pino-datadog](#pino-datadog)
-+ [pino-datadog-transport](#pino-datadog-transport)
 + [pino-gelf](#pino-gelf)
 + [pino-http-send](#pino-http-send)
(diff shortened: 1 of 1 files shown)


docs/transports.md
```diff
@@ -662,9 +662,36 @@ $ node foo | pino-datadog --key blablabla
...
+  level: "error", // minimum log level that should be sent to datadog
+})
+pino(transport)
+```
+
+[datadog-logger-integrations]: https://github.com/marklai1998/datadog-logger-integrations
+[Datadog]: https://www.datadoghq.com/
+
 <a id="pino-datadog-transport"></a>
 ### pino-datadog-transport
```
marklai1998, line 691: Not sure we should keep the entry here or not, one side there are still users using it

On the other hand, I don't think we should encourage user to use deprecated package, so I removed the entry in the index, and add a note here

marklai1998:
@mcollina can you help take a look a this? I don't wanna hang this PR for too long
Huge thanks!

marklai1998:
@mcollina up

PR #2230, opened 2025-06-11
Enhance logger function typing

samchungy:
An even bigger upgrade to https://github.com/pinojs/pino/pull/1995

This prevents metadata objects from being provided after a message and also guarantees type safety when you are using string templates.

### DIFF
--- a/test/types/pino.test-d.ts
+++ b/test/types/pino.test-d.ts
@@ -44,7 +44,6 @@ expectError(info('The object is %j', 'not a JSON'));
 expectError(info('The object is %O', 'not an object'));
 expectError(info('The answer is %d and the question is %s with %o', 42, { incorrect: 'order' }, 'unknown'));
 
-
 // object types with messages
 info({ obj: 42 }, "hello world");
 info({ obj: 42, b: 2 }, "hello world");
@@ -53,6 +52,9 @@ info({ a: 1, b: '2' }, 'hello world with %s', 'extra data');
 
 // metadata with messages type errors
 expectError(info({ a: 1, b: '2' }, 'hello world with %s', 123));
+
+// metadata after message
+expectError(info('message', { a: 1, b: '2' }));
 
 // multiple strings without template
 expectError(info('string1', 'string2', 'string3'));
--- a/test/types/pino.test-d.ts
+++ b/test/types/pino.test-d.ts
@@ -35,6 +35,8 @@ info('The json is %j', { a: 1, b: '2' });
 info('The object is %O', { a: 1, b: '2' });
 info('The answer is %d and the question is %s with %o', 42, 'unknown', { correct: 'order' });
 info('Missing placeholder is fine %s');
+declare const errorOrString: string | Error;
+info(errorOrString)
 
 // placeholder messages type errors
 expectError(info('Hello %s', 123));
--- a/test/types/pino.test-d.ts
+++ b/test/types/pino.test-d.ts
@@ -10,10 +10,53 @@ const error = log.error;
 
 info("hello world");
 error("this is at error level");
-info("the answer is %d", 42);
+
+// primative types
+info('simple string');
+info(true)
+info(42);
+info(3.14);
+info(null);
+info(undefined);
+
+// object types
+info({ a: 1, b: '2' });
+info(new Error());
+info(new Date());
+info([])
+info(new Map());
+info(new Set());
+
+// template messages
+info('Hello %s', 'world');
+info('The answer is %d', 42);
+info('The object is %o', { a: 1, b: '2' });
+info('The json is %j', { a: 1, b: '2' });
+info('The object is %O', { a: 1, b: '2' });
+info('The answer is %d and the question is %s with %o', 42, 'unknown', { correct: 'order' });
+
+// template messages type errors
+expectError(info('Hello %s', 123));
+expectError(info('Hello %s', false));
+expectError(info('The answer is %d', 'not a number'));
+expectError(info('The object is %o', 'not an object'));
+expectError(info('The object is %j', 'not a JSON'));
+expectError(info('The object is %O', 'not an object'));
+expectError(info('The answer is %d and the question is %s with %o', 42, { incorrect: 'order' }, 'unknown'));
+
+
+// object types with messages
 info({ obj: 42 }, "hello world");
 info({ obj: 42, b: 2 }, "hello world");
 info({ obj: { aa: "bbb" } }, "another");
+info({ a: 1, b: '2' }, 'hello world with %s', 'extra data');
+
+// metadata with messages type errors
+expectError(info({ a: 1, b: '2' }, 'hello world with %s', 123));
+
+// multiple strings without template
+expectError(info('string1', 'string2', 'string3'));
+
 setImmediate(info, "after setImmediate");
 error(new Error("an error"));
 
--- a/test/types/pino.test-d.ts
+++ b/test/types/pino.test-d.ts
@@ -27,16 +27,16 @@ info([])
 info(new Map());
 info(new Set());
 
-// template messages
+// placeholder messages
 info('Hello %s', 'world');
 info('The answer is %d', 42);
 info('The object is %o', { a: 1, b: '2' });
 info('The json is %j', { a: 1, b: '2' });
 info('The object is %O', { a: 1, b: '2' });
 info('The answer is %d and the question is %s with %o', 42, 'unknown', { correct: 'order' });
-info('Missing template is fine %s');
+info('Missing placeholder is fine %s');
 
-// template messages type errors
+// placeholder messages type errors
 expectError(info('Hello %s', 123));
 expectError(info('Hello %s', false));
 expectError(info('The answer is %d', 'not a number'));
@@ -44,7 +44,7 @@ expectError(info('The object is %o', 'not an object'));
 expectError(info('The object is %j', 'not a JSON'));
 expectError(info('The object is %O', 'not an object'));
 expectError(info('The answer is %d and the question is %s with %o', 42, { incorrect: 'order' }, 'unknown'));
-expectError(info('Extra message %s', 'after template', 'not allowed'));
+expectError(info('Extra message %s', 'after placeholder', 'not allowed'));
 
 // object types with messages
 info({ obj: 42 }, "hello world");
@@ -52,7 +52,7 @@ info({ obj: 42, b: 2 }, "hello world");
 info({ obj: { aa: "bbb" } }, "another");
 info({ a: 1, b: '2' }, 'hello world with %s', 'extra data');
 
-// Extra message after template
+// Extra message after placeholder
 expectError(info({ a: 1, b: '2' }, 'hello world with %d', 2, 'extra' ));
 
 // metadata with messages type errors
@@ -61,7 +61,7 @@ expectError(info({ a: 1, b: '2' }, 'hello world with %s', 123));
 // metadata after message
 expectError(info('message', { a: 1, b: '2' }));
 
-// multiple strings without template
+// multiple strings without placeholder
 expectError(info('string1', 'string2'));
 expectError(info('string1', 'string2', 'string3'));
 
--- a/test/types/pino.test-d.ts
+++ b/test/types/pino.test-d.ts
@@ -34,6 +34,7 @@ info('The object is %o', { a: 1, b: '2' });
 info('The json is %j', { a: 1, b: '2' });
 info('The object is %O', { a: 1, b: '2' });
 info('The answer is %d and the question is %s with %o', 42, 'unknown', { correct: 'order' });
+info('Missing template is fine %s');
 
 // template messages type errors
 expectError(info('Hello %s', 123));
@@ -43,12 +44,16 @@ expectError(info('The object is %o', 'not an object'));
 expectError(info('The object is %j', 'not a JSON'));
 expectError(info('The object is %O', 'not an object'));
 expectError(info('The answer is %d and the question is %s with %o', 42, { incorrect: 'order' }, 'unknown'));
+expectError(info('Extra message %s', 'after template', 'not allowed'));
 
 // object types with messages
 info({ obj: 42 }, "hello world");
 info({ obj: 42, b: 2 }, "hello world");
 info({ obj: { aa: "bbb" } }, "another");
 info({ a: 1, b: '2' }, 'hello world with %s', 'extra data');
+
+// Extra message after template
+expectError(info({ a: 1, b: '2' }, 'hello world with %d', 2, 'extra' ));
 
 // metadata with messages type errors
 expectError(info({ a: 1, b: '2' }, 'hello world with %s', 123));
@@ -57,6 +62,7 @@ expectError(info({ a: 1, b: '2' }, 'hello world with %s', 123));
 expectError(info('message', { a: 1, b: '2' }));
 
 // multiple strings without template
+expectError(info('string1', 'string2'));
 expectError(info('string1', 'string2', 'string3'));
 
 setImmediate(info, "after setImmediate");
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -338,7 +338,7 @@ declare namespace pino {
         : Acc;
 
     interface LogFn {
-        <T, TMsg extends string = string>(obj: Exclude<T, string>, msg?: TMsg, ...args: ParseLogFnArgs<TMsg> | []): void;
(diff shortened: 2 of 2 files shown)


test/types/pino.test-d.ts
```diff
@@ -10,10 +10,61 @@ const error = log.error;
...
+info(new Set());
+
+// placeholder messages
+info('Hello %s', 'world');
+info('The answer is %d', 42);
+info('The object is %o', { a: 1, b: '2' });
+info('The json is %j', { a: 1, b: '2' });
+info('The object is %O', { a: 1, b: '2' });
+info('The answer is %d and the question is %s with %o', 42, 'unknown', { correct: 'order' });
+info('Missing placeholder is fine %s');
```
samchungy, line 37: I've allowed strings with missing template data but if we want to be more strict and require that they add it we can just change the type of `args` from `...args: ParseLogFnArgs<TMsg> | []` to `...args: ParseLogFnArgs<TMsg>` .

But, this is technically valid

tadhglewis:
lgtm 🚀

samchungy:
In case you want something for a changelog, we wrote something like this for ours. Feel free to throw it away/tweak it however you would like @mcollina 🙏 

---
'@seek/logger': major
---

Enforce stricter typing for logger methods

This is a **breaking change** to the types, improving type safety by enforcing stricter parameter typing on all logger methods.

Existing code that passes metadata after the message string will need to be updated. This pattern was previously silently ignoring the metadata, but now triggers a type error to prevent data loss.

**Before (no longer works):**

```ts
logger.error('my message', { err, metadata });
```

**After (correct usage):**

```ts
logger.error({ err, metadata }, 'my message');
```

This ensures all metadata is properly captured and logged.

This update also enforces type checking for placeholder values in message strings:

**Before (no longer works):**

```ts
logger.error({ err, metadata }, 'my message error code: %d', 'Bad Request');
```

**After (correct usage):**

```ts
logger.error({ err, metadata }, 'my message error code: %d', 400);
```

samchungy:
One small side-effect of this which someone in my team ran into is if you're trying to access the `calls` via a Jest Mock - you may lose typing from the 2nd argument onwards due to the conditional types that I've added.

Eg. 

```ts
logger.warn.mock.calls[0]?.[1]
```

You may need to change that to:

```ts
logger.warn.mock.calls[0]?.at(1)
```

dcitron:
@samchungy we noticed an oddity -- we're using a `logMethod` hook as suggested in the documentation -- https://getpino.io/#/docs/api?id=message

The hook allowed us to write code like this:

```typescript
// ...
  } catch (err) {
    pino.error(getContext(), 'Failed to send error to Slack:', err);
  }
```

And also this:

```typescript
console.info = (...args) => {
  pino.info(getContext(), ...args);
};
```

After this change to Pino, it seems that now we have to define a type that represents the prior definition of logFn and use it like this:

```typescript
type UnsafePinoLogFn = (obj: unknown, msg?: string, ...args: unknown[]) => void;

// ...
  } catch (err) {
    (pino.error as unknown as UnsafePinoLogFn)(
      getContext(),
      'Failed to send error to Slack:',
      err
    );

// ...

console.info = (...args) => {
  (pino.info as unknown as UnsafePinoLogFn)(getContext(), ...args);
};

// ... etc
```

Does that seem to make sense?

samchungy:
Is `getContext()` returning a placeholder string? Otherwise err would be getting lost to the void.

This change is bound to get people coming here, but unless you're using a placeholder string, Pino actually drops everything after a message string.

I would suggest something like: 

pino.error({ ...getContext(), err }, 'Failed to send error to Slack:');

dcitron:
Oh, `getContext()` is returning an object.

So, maybe this is unorthodox, but we're monkeypatching `console.error` etc with pino methods, so as a result, we have a logMethod hook that supports multi-argument logging with formatting placeholders and error/object handling the way `console.log` does.:

```typescript
/**
 * A custom logMethod hook for Pino that mimics console.log behavior
 * for handling multiple arguments beyond the first input.
 */
const consoleLogPinoHook = {
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  logMethod(args: any[], method: (...args: any[]) => void) {
    if (args.length <= 2) {
      // Default behavior for a single argument
      method.apply(this, args);
      return;
    }

    // Handle multiple arguments like console.log does
    // Format: first argument is the message (potentially with format specifiers)
    // followed by any number of arguments to be substituted or concatenated
    const contextArg = args.shift();
    const firstArg = args[0];
    const restArgs = args.slice(1);

    // If the first argument is a string and contains format specifiers,
    // handle them similar to how console.log would
    { ... }
    
    // Else no format specifiers or first arg is not a string
    // Just concatenate everything with spaces, like console.log does
    { ... }

  },
};
```

samchungy:
The change I made where was to help prevent users from sending their metadata to the void for a logger with standard configuration.

I think defining something custom would be inevitable for your case

You can restore the previous types with an overload/declaration merging

Eg.
```ts
declare module "pino" {
  interface LogFn {
    <T extends object>(obj: T, msg?: string, ...args: any[]): void;
  }
}
```

I would keep this next to your monkey patching code.

dcitron:
That makes sense! Thank you!

Dhruv-Garg79:
I am also facing type error after recent update. The below code used to work earlier.
```js
class Logger {
	private readonly requestInfo: TRequestInfo;

	constructor(reqInfo: TRequestInfo) {
		this.requestInfo = reqInfo;
	}

	public debug(msg: string, ...args: any[]): void {
		pinoLogger.debug(this.requestInfo, msg, ...args);
	}
}
```

If I change the new type definition to have `...args: any[]` instead of `...args: ParseLogFnArgs<TMsg> | []`. It will start working for me again. But, is there a better way to do it?
```js
interface LogFn {
        <T, TMsg extends string = string>(obj: T, msg?: T extends string ? never: TMsg, ...args:any]): void;
        <_, TMsg extends string = string>(msg: TMsg, ...args: ParseLogFnArgs<TMsg> | []): void;
    }
```

samchungy:
> I am also facing type error after recent update. The below code used to work earlier.
> 
> ```js
> class Logger {
> 	private readonly requestInfo: TRequestInfo;
> 
> 	constructor(reqInfo: TRequestInfo) {
> 		this.requestInfo = reqInfo;
> 	}
> 
> 	public debug(msg: string, ...args: any[]): void {
> 		pinoLogger.debug(this.requestInfo, msg, ...args);
> 	}
> }
> ```
> 
> If I change the new type definition to have `...args: any[]` instead of `...args: ParseLogFnArgs<TMsg> | []`. It will start working for me again. But, is there a better way to do it?
> 
> ```js
> interface LogFn {
>         <T, TMsg extends string = string>(obj: T, msg?: T extends string ? never: TMsg, ...args:any]): void;
>         <_, TMsg extends string = string>(msg: TMsg, ...args: ParseLogFnArgs<TMsg> | []): void;
>     }
> ```

Before you consider changing the types, are you also similarly patching Pino? Have you read why the change is happening? Are you aware args aren't being logged unless msg is a placeholder string?

AaronMoat:
@Dhruv-Garg79 if I'm understanding your need correctly, you could make the `debug` method also use the new generics:

```
class Logger {
	private readonly requestInfo: TRequestInfo;

	constructor(reqInfo: TRequestInfo) {
		this.requestInfo = reqInfo;
	}

	public debug<TMsg extends string = string>(
        msg: TMsg,
        ...args: pino.ParseLogFnArgs<TMsg>
    ): void {
		pinoLogger.debug(this.requestInfo, msg, ...args);
	}
}
```

This allows you to have type safety in your custom class.

[Playground](https://www.typescriptlang.org/play/?ssl=7&ssc=1&pln=21&pc=1#code/JYWwDg9gTgLgBGYA7CcBmUIjgckSnAbgChiYBPMAUzgBUAlKgRwFcqBnGASSTVQF44LJAGsUAdyQliAYwhJOCZBAAyEAObqqUOIPwQAFAEppMgDYBDduzhrN2uAG9iASDBRgANwswaUKhYAJvJm5HD+rBzcvBAAXHSMkZw8fNIucgowUCwyMNAGESlxCcxsyTFGTq4uMAAWwOwAdBFl0Xy64cxFJC4AvqRuLABGZsAycIFUQyzqADy0ALLs6nBUAB6+SIE2nB5IK4K7yOoAfAbEcJdXlyDL8YvLADQX11eN7xZQ6uzx+o0ACp92FQ7AAxJAAQS+7HmS1OL0uRninggwECVRcbmUdi0UEak2m6gMdQazVKUSKjzgt3UVPejU+3xMrn6-Vk8kUZg6SCo4lsGlxBkcvWZxDM+KmMwMOFqVDMZggOFFAHplXAAAIwdgAWnW1FyuqgmCgYolhOlsvliqpODq2ioStIqo1Wt1a31MENxtNBKlMrlCrgAFJ2I6fZKif6rcHQzbxNAzIElUA)

Dhruv-Garg79:
> Before you consider changing the types, are you also similarly patching Pino?

No

> Have you read why the change is happening?

yes, but it will lead to lot of changes for us, that's why I asked if there was any other better way to handle it instead of just monkey patching types.

> Are you aware args aren't being logged unless msg is a placeholder string?

yes. and that doesn't affect us. since we are using a small wrapper over pino.


Hey @AaronMoat, that's a good idea, but that will require us to make a lot of changes wherever we are using the logger. Because value can be T or undefined where %s or %j is used in our case.

AaronMoat:
Hi @Dhruv-Garg79 — are you able to show a problematic TS Playground example? I'm not following the problem, nor what T is in your comment.

Dhruv-Garg79:
@AaronMoat I have updated your TS playground with the issue. The problem arises when the templated value is null or undefined.

[playground](https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBGYA7CcBmUIjgckSnAbgChiYBPMAUzgBUAlKgRwFcqBnGASSTVQF44LJAGsUAdyQliAYwhJOCZBAAyEAObqqUOIPwQAFAEppMgDYBDduzhrN2uAG9iASDBRgANwswaUKhYAJvJm5HD+rBzcvBAAXHSMkZw8fNIucgowUCwyMNAGESlxCcxsyTFGTq4uMAAWwOwAdBFl0Xy64cxFJC4AvqRuLABGZsAycIFUQyzqADy0ALLs6nBUAB6+SIE2nB5IK4K7yOoAfAbEcJdXlyDL8YvLADQX11eN7xZQ6uzx+o0ACp92FQ7AAxJAAQS+7HmS1OL0uRninggwECVRcbmUdi0UEak2m6gMdQazVKUSKjzgt3UVPejU+3xMrn6-Vk8kUZg6SCo4lsGlxBkcvWZxDM+KmMwMOFqVDMZggOFFAHplXAAAIwdgAWnW1FyuqgmCgYolhOlsvliqpODq2ioStIqo1Wt1a31MENxtNBKlMrlCrgAFJ2I6fZKif6rcHQzbxNAzIEw2YqPBvGY2HB4kd9nAAD5CLZUNDIKjowTCSYlnmBEji32Ry2BkM4Knptii9s0QQ4eNQRM4cPmqPN2NwLsmIA)

samchungy:
I mean I think that makes sense there, you're trying to pass something that could be `null` or `undefined` to a placeholder which is expecting a string.

You can always bypass it with some casts or coercion.

Either:

```ts
l.debug('hello %s', value ?? '');

l.debug('hello %s', value as string);
```

Or using `any` on your own logger implementation. Alternatively, write your own version of `ParseLogFnArgs` which accepts `null` or `undefined` and use a cast to pass it to pino.

eg. `pinoLogger.debug(this.requestInfo, msg, ...args as ParseLogFnArgs<TMsg>);`

```ts
	public debug<TMsg extends string = string>(
        msg: TMsg,
        ...args: any
    ): void {
		pinoLogger.debug(this.requestInfo, msg, ...args);
	}
```

Or declare the override as suggested above.

clemens:
At the risk of asking a stupid question: Why is this done as part of a minor release when it clearly breaks the public API (even if it's "only" the types) of a lot of functions on the main/only object? And if we say this is acceptable, shouldn't the changelog at least mark it as a breaking change (instead, it's even called "enhance", which suggests an improvement)?

Dhruv-Garg79:
@samchungy I get it, but since it's logging. I think we should not need to handle it by casting or coercion. For now, I can override the types, but it should allow null/undefined directly.

samchungy:
> I think we should not need to handle it by casting or coercion

We can agree to disagree but I think if your template is expecting a string, the type should show that.

Unless you were to introduce alternate identifiers eg. %ns (nullish string?), I think the types should reflect what the templates identifiers represent.

How would you coerce null or undefined into %d which is expecting a number?

That being said, go with `any`s if you don't want types.

Dhruv-Garg79:
> We can agree to disagree but I think if your template is expecting a string, the type should show that.

yes we can. It should be use case dependent, and the thing we are talking about is logging right now. All types, according to me should allow null/undefined by default.

> Unless you were to introduce alternate identifiers eg. %ns (nullish string?), I think the types should reflect what the templates identifiers represent.

Types like %ns, %nd, %nj are a good middle ground for nullish types, which allow both null and undefined values.

I have changes my types to be like this for now
```
type PlaceholderTypeMapping<T extends pino.PlaceholderSpecifier> = T extends 'd'
	? number | null | undefined
	: T extends 's'
		? string | null | undefined
		: T extends 'j' | 'o' | 'O'
			? object | null | undefined
			: never;
```

Also shouldn't we have a separate type for boolean as well? something like %b?

samchungy:
>All types, according to me should allow null/undefined by default.

I think TypeScript largely disagrees with you 🤷

>Also shouldn't we have a separate type for boolean as well? something like %b?

This is open source, so PR away my friend.

Dhruv-Garg79:
> I think TypeScript largely disagrees with you 🤷

You took my statement out of context. I said for logging, of course, not for everything.

What I wrote was - yes we can. It should be use case dependent, and the thing we are talking about is logging right now. All types, according to me should allow null/undefined by default.

> This is open source, so PR away my friend.

Sure.

klippx:
> At the risk of asking a stupid question: Why is this done as part of a minor release when it clearly breaks the public API (even if it's "only" the types)

It is not only the types, this is a super breaking change that disagrees with everything stated in the docs: https://getpino.io/#/docs/api?id=logging-method-parameters

This is pretty common usage in our code base:
```
      this.logger.warn('Got fallback values from REDA', {
        pno12: body.REDAIndividualValuesRequest.Order.PNO12,
        options: body.REDAIndividualValuesRequest.Order.Options,
        clientName: clientName ?? 'unknown',
      });
```

It is used as per docs, but after this PR it is "wrong" according to types (but it works in runtime)

How was this approved and merged? I don't even... 🤷

samchungy:
> > At the risk of asking a stupid question: Why is this done as part of a minor release when it clearly breaks the public API (even if it's "only" the types)
> 
> It is not only the types, this is a super breaking change that disagrees with everything stated in the docs: https://getpino.io/#/docs/api?id=logging-method-parameters
> 
> This is pretty common usage in our code base:
> 
> ```
>       this.logger.warn('Got fallback values from REDA', {
>         pno12: body.REDAIndividualValuesRequest.Order.PNO12,
>         options: body.REDAIndividualValuesRequest.Order.Options,
>         clientName: clientName ?? 'unknown',
>       });
> ```
> 
> It is used as per docs, but after this PR it is "wrong" according to types (but it works in runtime)
> 
> How was this approved and merged? I don't even... 🤷

Well the metadata there doesn't get logged 🤷‍♂️ (have you checked your logs?)

I'm not disagreeing that it should be a breaking change

klippx:
Then at the docs needs an urgent update. But it would have been great if this was a v10.0.0 release for everyone's sanity, it's kind of hard to patch the docs now. Usually it is preferred to have a version selector for the docs, now they change in the middle of v9.

But it seems this change is legit then and that I might go through with the upgrade. Wishing there was a codemod for this :)

samchungy:
> Then at the docs needs an urgent update. But it would have been great if this was a v10.0.0 release for everyone's sanity, it's kind of hard to patch the docs now. Usually it is preferred to have a version selector for the docs, now they change in the middle of v9.
> 
> But it seems this change is legit then and that I might go through with the upgrade. Wishing there was a codemod for this :)

I didn't write the docs but where does it say that's valid?

klippx:
Ouf.... I guess I can't read - apologies. You are right. The docs seem to agree more with this PR and you. 🙇

samchungy:
> Ouf.... I guess I can't read - apologies. You are right. The docs seem to agree more with this PR and you. 🙇

No probs! I knew this was gonna irk some people so I was hoping it would be a major release but I promise it was done purely out of good intentions 🙏

PR #2232, opened 2025-06-16
feat: introduce `logger.msgPrefix` getter

indutny-signal:
When creating a logger or a child logger API allows providing `msgPrefix` string that gets appended to every formatted log line. However, there is no way to get this string from existing logger instance which makes it inaccessible from within the `hooks.logMethod`.

This commit adds the getter method and types for `logger.msgPrefix`.

cc @mcollina

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -22,6 +22,7 @@
   * [logger\[Symbol.for('pino.serializers')\]](#serializers)
   * [Event: 'level-change'](#level-change)
   * [logger.version](#version)
+  * [logger.msgPrefix](#msgPrefix)
 * [Statics](#statics)
   * [pino.destination()](#pino-destination)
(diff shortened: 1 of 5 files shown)


mcollina:
@indutny-signal hi Fedor! Sorry for the long delay! Can you add a quick test for this?

indutny-signal:
@mcollina sorry, I was OOO myself. Here you go!

mcollina (approved):
lgtm

mcollina:
Can you add it to https://github.com/pinojs/pino/blob/main/docs/api.md?

indutny-signal:
@mcollina good point! Just added some. Let me know if it looks good!

indutny-signal:
Thank you!

PR #2254, opened 2025-08-08
feat(types): add LogFnFields

samchungy:
Hello, we recently went down a little bit of a rabbit hole with wanting to enforce some typing on the logging function.

This feature adds the ability for us to add some global types to the logger functions.

This came about because we had some internal configuration where we wanted to add some typing to.

We have the ability to set our `retention` days by adding it to our log

eg.

```ts
logger.info({ logOptions: { retention: 90 }, err }, 'Something failed');
```

But we wanted `logOptions` to be properly typed without relying on users setting an overload everytime.

eg.

```ts
logger.info<LoggerOptions>({ logOptions: { retention: 90 }, err }, 'Something failed');
```

So I came up with introducing a module type augmentation which would allow us to set a single module augmentation:

```ts
declare module "pino" {
  interface LogFnFields {
    logOptions?: {
      retention: 30 | 60 | 90
    },
    logoptions?: never,
  }
}
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -37,6 +37,9 @@
   * [DestinationStream](#destinationstream)
 * [Types](#types)
   * [Level](#level-1)
+* [TypeScript](#typescript)
+  * [Module Augmentation](#module-augmentation)
+  * [LogFnFields Interface](#logfnfields-interface)
 
 <a id="export"></a>
 ## `pino([options], [destination]) => logger`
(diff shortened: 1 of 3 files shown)


mcollina:
Can you rebase? Sorry for the delay

mcollina (approved):
lgtm

mcollina:
Sorry, can you resolve the conflicts?

PR #2258, opened 2025-08-11
Fix TypeScript types

### LINKED ISSUE #2255: Regression: version 9.8.0 modifies export functionality.
Our team noticed in a project that relies on Pino, that our builds started to fail during ci/cd. We have not made any modification and the error is complaining about our usage of `pino`. 

*Error Message:*
```
 error TS2595: 'pino' can only be imported by using a default import.
```

I believe the issue is related to this commit included in 9.8.0: https://github.com/pinojs/pino/pull/2223/files

I have left the following [comment](https://github.com/pinojs/pino/pull/2223/files#r2262203895) for convenience.

joshkel:
Copying from #2255 for discussion:

From what I understand:

1. A typical CommonJS default export (`module.exports = pino`) should be represented in a TypeScript `.d.ts` file as `export = pino`.
2. If you want to export both types and a default (so that callers can do, e.g., `import pino, { type Level } from 'pino'`, you can do a namespace plus `export =`:
    ```ts
    namespace pino {
      export type Level = "fatal" | "error" | "warn" | "info" | "debug" | "trace";
    }
    declare function pino(msg: string);
    export = pino;
    ```
3. If you want to export types, a default, _and_ a named function (so that callers can do either of `import pino, { type Level } from 'pino'` or `import { pino, type Level } from 'pino'`, then the only way I can figure out how to do that is to duplicate the `declare function`. That's ugly but doable.
    ```ts
    namespace pino {
      export type Level = "fatal" | "error" | "warn" | "info" | "debug" | "trace";
      declare function pino(msg: string);
    }
    declare function pino(msg: string);
    export = pino;
    ```
4. If you want to export types, a default, and a named function _and also_ expose all exports as types on the named function (so callers can do, e.g., `import { pino } from 'pino'; const level: pino.Level;` - [like here](https://github.com/pinojs/pino/blob/v9.7.0/test/types/pino.test-d.ts#L4)), then I'm stumped - I can't figure out how to make a namespace refer to itself.

This PR takes approach 3. Therefore, it has a few backward compatibility changes relative to 9.7.0 and before:

* Code such as the following will not work:
    ```ts
    import { pino } from 'pino';
    const level: pino.Level;
    ```
    It can be updated to `import pino from 'pino'; const level: pino.Level;` or `import { Level } from 'pino'; const level: Level;`
* The [`P` namespace](https://github.com/pinojs/pino/blob/v9.7.0/pino.d.ts#L888) is no longer supported.
* The `import pinoCjsImport = require ("../../pino");` line in [pino-import.test-d.ts](https://github.com/pinojs/pino/blob/v9.7.0/test/types/pino-import.test-d.ts) and its reference to `pinoCjsImport.default()` started failing ("Import assignment cannot be used when targeting ECMAScript modules. Consider using 'import * as ns from "mod"', 'import {a} from "mod"', 'import d from "mod"', or another module format"). To fix this, I renamed the test to `pino-import.test-d.cts` (to force it to compile as a CommonJS file) and removed the `pinoCjsImport.default()` reference. Therefore, code using either of these patterns may fail - but, from what I understand, these aren't typical usage patterns, and they have straightforward alternatives.

Fixes #2255

### DIFF
--- a/pino.d.ts
+++ b/pino.d.ts
@@ -20,124 +20,120 @@ import * as pinoStdSerializers from "pino-std-serializers";
 import type { SonicBoom, SonicBoomOpts } from "sonic-boom";
 import type { WorkerOptions } from "worker_threads";
 
-
-
-//// Non-exported types and interfaces
-
-// ToDo https://github.com/pinojs/thread-stream/issues/24
-type ThreadStream = any
-
-type TimeFn = () => string;
-type MixinFn<CustomLevels extends string = never> = (mergeObject: object, level: number, logger:pino.Logger<CustomLevels>) => object;
-type MixinMergeStrategyFn = (mergeObject: object, mixinObject: object) => object;
-
-type CustomLevelLogger<CustomLevels extends string, UseOnlyCustomLevels extends boolean = boolean> = { 
-    /**
-     * Define additional logging levels.
-     */
-    customLevels: { [level in CustomLevels]: number };
-    /**
-     * Use only defined `customLevels` and omit Pino's levels.
-     */
-    useOnlyCustomLevels: UseOnlyCustomLevels;
- } & {
-    // This will override default log methods
-    [K in Exclude<pino.Level, CustomLevels>]: UseOnlyCustomLevels extends true ? never : pino.LogFn;
- } & {
-    [level in CustomLevels]: pino.LogFn;
- };
-
-/**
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

bcomnes:
Thanks for this.

PR #2262, opened 2025-08-18
fix: TypeScript regression with boolean values in %s placeholders

### LINKED ISSUE #2259: Regression: TS - cannot log boolean values - Pino 9.9.0
Pino 9.9.0 cannot accept booleans as log placeholders any more, given the Typescript definitions:

```typescript
const biteMe = true
logger.debug("can I bite? %s, biteMe)
```

This now generates a typescript compiler error: 

`error TS2345: Argument of type '[boolean]' is not assignable to parameter of type '[] | [string]'.

This has worked for years for me.

mcollina:
Fixes #2259 - TypeScript regression introduced in v9.9.0 where boolean values could not be used with %s placeholders in logging methods.

The %s placeholder now correctly accepts all primitive types (string, number, boolean, bigint, null, undefined, symbol) as it should, matching the runtime behavior where these values are converted to strings.

- Updated PlaceholderTypeMapping to allow all primitives for %s
- Added comprehensive test coverage for all primitive types with %s
- Removed incorrect type error expectations for boolean and number with %s

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -757,7 +757,7 @@ from `msg` to another string as preferred.
 The `message` string may contain a printf style string with support for
 the following placeholders:
 
-* `%s` – string placeholder
+* `%s` – string placeholder, every non-string value passed in will have `.toString()` called.
 * `%d` – digit placeholder
 * `%O`, `%o`, and `%j` – object placeholder
 


samchungy (approved):
Ah okay - could we update the docs if %s accepts all primitives? I was going off this:

https://getpino.io/#/docs/api?id=message-string

Issue #2260, opened 2025-08-18
stdTimeFunctions property not found in TypeScript typings


dc7290:
Hi,

When using pino with TypeScript, I encountered the following error:

```ts
const { pino } = require('pino');
pino.stdTimeFunctions;
//              ^ error TS2339: Property 'stdTimeFunctions' does not exist on type '...'.
```

It seems that stdTimeFunctions is available at runtime, but the TypeScript definitions do not include it.

Environment:

pino: v9.9.0

Expected behavior:
TypeScript typings should expose pino.stdTimeFunctions without error.

Example of modification:

How about modifying it as follows?

`pino.d.ts`

```diff
 namespace pino {
-    //// Nested version of default export for TypeScript/Babel compatibility
-
-    /**
-     * @param [optionsOrStream]: an options object or a writable stream where the logs will be written. It can also receive some log-line metadata, if the
-     * relative protocol is enabled. Default: process.stdout
-     * @returns a new logger instance.
-     */
-    function pino<CustomLevels extends string = never, UseOnlyCustomLevels extends boolean = boolean>(optionsOrStream?: LoggerOptions<CustomLevels, UseOnlyCustomLevels> | DestinationStream): Logger<CustomLevels, UseOnlyCustomLevels>;
-
-    /**
-     * @param [options]: an options object
-     * @param [stream]: a writable stream where the logs will be written. It can also receive some log-line metadata, if the
-     * relative protocol is enabled. Default: process.stdout
-     * @returns a new logger instance.
-     */
-    function pino<CustomLevels extends string = never, UseOnlyCustomLevels extends boolean = boolean>(options: LoggerOptions<CustomLevels, UseOnlyCustomLevels>, stream?: DestinationStream | undefined): Logger<CustomLevels, UseOnlyCustomLevels>;
+    //// Nested version of default export for TypeScript/Babel compatibility
+    /**
+     * Mirrors the default export including its static properties, so that
+     * `const { pino } = require('pino')` preserves the callable type *and*
+     * static members like `stdTimeFunctions`.
+     */
+    const pino: typeof import("pino");
 }

```

MobinAskari:
Same issue in 9.10.0.
Every update breaks something to the point that I've pinned 9.7.0 as the only installable version. 
Hopefully the situation will be improved.

abhu85:
## Verification: Issue Fixed in Current Version

I tested this with **pino v9.14.0** (latest) and the issue appears to be resolved.

### Evidence

[...]

PR #2263, opened 2025-08-19
types: expose stdTimeFunctions on nested namespace pino while preserv…

tawseefnabi:
…ing callability; fix TS typings for CJS destructuring (#2260)


dc7290:
I thought it would be resolved for stdTimeFunctions, but not for other method type definitions?

mcollina (approved):
lgtm

PR #2278, opened 2025-09-05
Add support for `Pear` and `Bare` runtimes

yassernasc:
Hello, this PR adds out-of-box support for [Bare](https://github.com/holepunchto/bare) JavaScript runtime and therefore [Pear](https://github.com/holepunchto/pear) runtime also.

A [subpath import](https://nodejs.org/docs/latest/api/packages.html#subpath-imports) field was added in the `package.json`; it is a map that covers every built-in Node.js module that is used internally by Pino and happens to not be provided by default by `Bare`. A intuitive case is the global variable `process` which requires an extra step to be available.

`{ with: { imports: './package.json' } }` is a feature from `Bare` module resolution that dynamically applies the declared `subpath imports`. When executing Pino in Node.js, the `{ with: { imports: './package.json' } }` is ignored by the Node.js module resolution, resulting in no performance impacts.

A unit test was added to ensure that the basic usage works as intended in `Bare`. I would like to follow the pattern and use `tap` as the other test files do, but for that, it will be needed some additional work to add `Bare` portability to `tap` and its dependencies as well. So I used the already `Bare`-friendly test framework [Brittle](https://github.com/holepunchto/brittle) for this particular case.

Open to suggestions and thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Pino works on [Bare](https://github.com/holepunchto/bare) with the [`pino-bare`]
 
 ### Pear
 
-Pino works on [Pear](https://docs.pears.com), which is built on [Bare](https://github.com/holepunchto/bare), with the [`pino-bare`](https://github.com/pinojs/pino-bare) compatability module.
+Pino works on [Pear](https://docs.pears.com), which is built on [Bare](https://github.com/holepunchto/bare), with the [`pino-bare`](https://github.com/pinojs/pino-bare) compatibility module.
 
 
 ## Install


davidmarkclements:
@mcollina per our discussion.

This would mean that Pino supports Node.js, Bare and Pear - we can of course always create a module that wraps but thought it makes sense to offer it directly for consideration

davidmarkclements:
also side note, since Mobile has core support from Bare, this would also unlock Pino for Mobile apps

davidmarkclements:
@yassernasc needs ci workflow, and needs to run all the tests on bare

jsumners (commented):
I think these alternative runtimes should focus on implementing compatibility with the de facto standard runtime instead of insisting ecosystem modules comply with their design decisions. To be clear: I'm not in favor of the whole PR. I would be in favor of a PR that adds an alternative runtimes CI workflow and doesn't add any further requirements on Pino itself.

yassernasc:
Thanks for your time @mcollina and @jsumners, it's true that the first changes proposed here are quite invasive to Pino's core. We are constantly researching how to approach the integration with the Node.js ecosystem without losing the principles that motivates the Bare project. Your feedback was very valuable.

All the integration logic is now placed at: https://github.com/pinojs/pino-bare 🫶

davidmarkclements:
@mcollina we now have https://github.com/pinojs/pino-bare, I've updated the readme in this PR to signpost to pino-bare, via a new Runtimes header

PR #2279, opened 2025-09-05
Add WHATWG Console API adapter

mcollina:
## Summary
- Implements Console constructor that wraps Pino loggers
- Full Node.js Console API compatibility with TypeScript support
- All 21 Console methods implemented with proper type definitions

## Usage
```js
const pino = require('pino')
const logger = pino()
const console = new pino.Console(logger)
console.log('Hello world!') // logs via pino.info()
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@
 
 * [Benchmarks ⇗](/docs/benchmarks.md)
 * [API ⇗](/docs/api.md)
+* [Console Adapter ⇗](/docs/console-adapter.md)
 * [Browser API ⇗](/docs/browser.md)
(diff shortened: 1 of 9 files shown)


mcollina:
@jsumners do you think this is better in core rather or have a pino-console module?

jsumners:
I am pretty sure we are talking about https://console.spec.whatwg.org/

That being the case, I'd suggest the independent module. Specifically because:

> This specification is an early work in progress that welcomes feedback to refine toward more precise and compatible definitions.

mcollina:
Closing in favor of https://github.com/pinojs/pino-console

Issue #2289, opened 2025-09-18
pino@9.10.0 breaks on Node v18.18 due to missing diagnostics_channel.tracingChannel


aryamohanan:
Hi Pino team 👋

According to [Pino’s LTS policy](https://github.com/pinojs/pino/blob/main/docs/lts.md), Pino 9.x supports Node 18, 20, and 22 (latest releases). However, in CI environments using Node v18.18, we’re seeing:

```
TypeError: diagChan.tracingChannel is not a function
    at Object.<anonymous> (node_modules/pino/lib/tools.js:31:29)
```

This is because `diagnostics_channel.tracingChannel()` was only added in Node.js v18.19+, so `pino@9.10.0` crashes on earlier 18.x releases.

see: https://nodejs.org/api/diagnostics_channel.html#diagnostics_channeltracingchannelnameorchannels

**Environment:**

* Node: v18.18.2
* Pino: v9.10.0
* OS: Linux (CI), macOS (local)

**Expected behavior:**

It would be helpful if:

* `pino@9.10.0` either gracefully detects `diagChan.tracingChannel` and disables tracing if missing,
* or clearly states in documentation/release notes that Node >= 18.19 is required, not just “Node 18”.

MukulWBD:
I am still facing the same issue even after you reverted the latest updates.
Everything was fine till yesterday 

<img width="1099" height="483" alt="Image" src="https://github.com/user-attachments/assets/dad0dd8f-b6c1-44e6-9aed-609d72603204" />

mcollina:
@aryamohanan thanks for reporting! We didn't catch this because as it's clearly written in the [LTS document](https://github.com/pinojs/pino/blob/f8e4161b801f59d0b394e82a0cd8f8a482cd9be8/docs/lts.md?plain=1#L17-L21):

> Major releases will be tested and verified against all Node.js release lines that are supported by the [Node.js LTS policy](https://github.com/nodejs/Release) within the LTS period of that given Pino release line. __This implies that only the latest Node.js release of a given line is supported.__

How can it be clearer?

mcollina:
@aryamohanan a PR to make this detected would be highly welcomed.

cc @jsumners-nr

aryamohanan:
Thanks for the clarification, @mcollina!

We have updated our code base to pin pino@9.9.5 to fix the issue temporarily. I’m happy to open a PR to handle this gracefully.

mcollina:
Go for it!

[...]

PR #2281, opened 2025-09-10
Add tracing channels

jsumners-nr:
This PR adds a tracing channel for the `asJson` function. APM tools can utilize this channel for their tracing needs. At New Relic, this is the only channel we need. Other vendors may like a bit more, so I'm tagging them here to get their input.

I'm opening this in draft mode while the feedback is collected. Once we have "yeah, this is good" from enough people, I will add documentation and convert the PR to ready.

attn: @AbhiPrasad @bengl @bizob2828 @timfish

PR #2290, opened 2025-09-19
fix: gracefully handle missing diagChannel.tracingChannel on Node < 18.19

aryamohanan:
This PR fixes a compatibility issue when running Pino on Node.js 18.x versions earlier than **18.19**, where `diagnostics_channel.tracingChannel` is not available.

### Problem

In Node 18.18 and earlier, `diagChannel.tracingChannel` does not exist. This caused Pino’s `asJson` tracing logic to throw an error at runtime.

### Solution

* Added **feature detection** for `diagChannel.tracingChannel`.
* When unavailable, provide a minimal no-op fallback:

  * `hasSubscribers: false`
  * `traceSync()` simply calls the provided function
* Ensures `asJson` works normally, even without tracing support.

### Notes

* Added inline comments to clarify that `tracingChannel` is available starting with Node 18.19+.
* Kept the fallback implementation minimal to avoid performance impact.

### Reference

This issue was added in the latest Pino release, https://github.com/pinojs/pino/pull/2281 and diagnostics_channel.tracingChannel was introduced in Node.js starting from v18.19 (see [Node.js docs](https://nodejs.org/api/diagnostics_channel.html#diagnostics_channeltracingchannelnameorchannels)).

closes https://github.com/pinojs/pino/issues/2289

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, windows-latest, ubuntu-latest]
-        node-version: [18, 20, 22]
+        node-version: [18, '18.18', 20, 22]
         exclude:
           - os: windows-latest
             node-version: 22
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -28,7 +28,19 @@ const {
 const { isMainThread } = require('worker_threads')
 const transport = require('./transport')
 
-const asJsonChan = diagChan.tracingChannel('pino_asJson')
+let asJsonChan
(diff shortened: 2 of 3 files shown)


mcollina:
cc @jsumners

aryamohanan:
@mcollina I tested this on macOS with Node.js v20, and it passed. Could this be a flaky test in CI? Would it be possible to rerun the test?

jsumners (commented):
I'm not going to block this fair solution. But my opinion is that it is just adding stuff to remove for the next major (which is long overdue and something I hope to get done soon). I think we are clear that we only support the most current release in each release line, and v18.19 wasn't even the most current v18 (there were two releases of .19 and 9 releases of v18.20).

Issue #2071, opened 2024-10-28
--frozen-intrinsics: Cannot assign to read only property 'prepareStackTrace' of function 'function Error() { [native code] }'


NickKelly1:
When the NodeJS process runs with `--frozen-intrinsics`, creating a pino logger throws an error.

```
> node --env-file ./.env --enable-source-maps --disable-proto=throw --disallow-code-generation-from-strings --frozen-intrinsics ./build/main.js

(node:1764737) ExperimentalWarning: Frozen intristics is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
<cwd>/node_modules/.pnpm/pino@9.5.0/node_modules/pino/lib/caller.js:9
  Error.prepareStackTrace = noOpPrepareStackTrace
                          ^

TypeError <Object <Object <[Object: null prototype] {}>>>: Cannot assign to read only property 'prepareStackTrace' of function 'function Error() { [native code] }'
    at getCallers (<cwd>/node_modules/.pnpm/pino@9.5.0/node_modules/pino/lib/caller.js:9:27)
    at pino (<cwd>/node_modules/.pnpm/pino@9.5.0/node_modules/pino/pino.js:90:48)
    at server (<cwd>/src/commands/server.ts:8:17)
    at main (<cwd>/src/main.ts:58:8)
    at <anonymous> (<cwd>/src/main.ts:63:26)
    at ModuleJob.run (node:internal/modules/esm/module_job:268:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:543:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)

Node.js v23.1.0
ELIFECYCLE Command failed with exit code 1.
[nodemon] app crashed - waiting for file changes before starting...
```

Due to this line

https://github.com/pinojs/pino/blob/231adffeccd71f26627a43260c09bfb61a98fcaa/lib/caller.js#L9

mcollina:
Thanks for reporting! Now that Node.js has [`util.getCallSite()`](https://nodejs.org/api/util.html#utilgetcallsiteframes), I think we can rely on that instead of patching `Error` to achieve this if that new function is available. Would you like to send a Pull Request to address this? Remember to add unit tests.

NickKelly1:
Sure thing, I'll create a a PR for it.

NickKelly1:
[...]

PR #2294, opened 2025-09-22
fix: use util.getCallSites when able to allow frozen intrinsics

CarlOlson:
Issue: https://github.com/pinojs/pino/issues/2071

The examples pass now with `--frozen-intrinsics`.


jsumners:
Tests fail on Windows.

PR #2297, opened 2025-09-26
docs(ecosystem): add eslint-plugin-pino to ecosystem page

orzarchi:
Adding (my) new eslint plugin for pino users that catches this common error of passing in arguments to pino methods in the wrong order:

```javascript
// right
logger.info( { importantContext: 2 }, "very useful log message") // perfect log record

// wrong
logger.info("very useful log message", { importantContext: 2 }) // whoops, context is thrown away
```


jsumners (changes requested):
Nice work. But please add it to the list in alphabetical order.

orzarchi:
Fixed 😄

PR #2298, opened 2025-09-26
Replace fast-redact with slow-redact

mcollina:
Replaces fast-redact dependency with slow-redact for improved maintainability and security.

## Summary
- Updates package.json to use slow-redact ^0.2.0 instead of fast-redact
- Updates lib/redaction.js to use slow-redact while maintaining API compatibility
- Adds custom rx regex for path parsing to maintain existing functionality

## Test Results  
- 5 test failures out of 1386 total tests (0.36% failure rate)
- All core redaction functionality working correctly
- Minor edge cases with invalid path validation and empty string bracket notation

## Breaking Changes
None - maintains backward compatibility with existing pino redaction API

### DIFF
--- a/lib/redaction.js
+++ b/lib/redaction.js
@@ -1,13 +1,10 @@
 'use strict'
 
-const fastRedact = require('fast-redact')
+const slowRedact = require('slow-redact')
 const { redactFmtSym, wildcardFirstSym } = require('./symbols')
-const { rx, validator } = fastRedact
 
-const validate = validator({
-  ERR_PATHS_MUST_BE_STRINGS: () => 'pino – redacted paths must be strings',
-  ERR_INVALID_PATH: (s) => `pino – redact paths array contains an invalid path (${s})`
-})
+// Custom rx regex equivalent to fast-redact's rx
+const rx = /[^.[\]]+|\[([^[\]]*?)\]/g
 
 const CENSOR = '[Redacted]'
 const strict = false // TODO should this be configurable?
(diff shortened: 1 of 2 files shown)


mcollina:
Are you ok with a minor?

jsumners:
Let's just issue v10 with this plus some other changes I'm about to start prepping. It's long past time.

kibertoad:
@jsumners @mcollina should we also use the opportunity to drop support for node 18?

mcollina:
My main concern for doing a minor with this is to provide a solution to https://github.com/davidmarkclements/fast-redact/issues/75. Otherwise we are going to have to fork fast-redact and publish under a different name to avoid scanners - I'm afraid it would take too long for that CVE to go away (I hope to be wrong).

Overall happy with a major. I'm happy to set the fork up if needed.

mcollina:
On a second note, Fastify is already 20+, so we can easily update there if the only breaking is dropping 18.

jsumners:
> My main concern for doing a minor with this is to provide a solution to [davidmarkclements/fast-redact#75](https://github.com/davidmarkclements/fast-redact/issues/75). Otherwise we are going to have to fork fast-redact and publish under a different name to avoid scanners - I'm afraid it would take too long for that CVE to go away (I hope to be wrong).
> 
> Overall happy with a major. I'm happy to set the fork up if needed.

Okay. We'll just use this clause:

https://github.com/pinojs/pino/blob/550f8c2efc00ec72fd5d3b1d3d114b610a9cfdfb/docs/lts.md?plain=1#L25-L36

> On a second note, Fastify is already 20+, so we can easily update there if the only breaking is dropping 18.

That should be it. We've just been too busy/exhausted to get it done.

mcollina:
It's not really a security vulnerability, it's literally just appeasing security scanners and the nice people that published an invalid CVE on fast-redact without even pinging us. Meh.

jsumners:
> It's not really a security vulnerability, it's literally just appeasing security scanners and the nice people that published an invalid CVE on fast-redact without even pinging us. Meh.

I know. But not like we can do anything else.

PR #2299, opened 2025-09-27
Convert tests to node:test

jsumners:
WIP

### DIFF
--- a/test/broken-pipe.test.js
+++ b/test/broken-pipe.test.js
@@ -1,47 +1,37 @@
 'use strict'
 
-const t = require('tap')
+const test = require('node:test')
+const assert = require('node:assert')
 const { join } = require('node:path')
 const { fork } = require('node:child_process')
+const tspl = require('@matteo.collina/tspl')
 const { once } = require('./helper')
 const pino = require('..')
 
 if (process.platform === 'win32') {
-  t.skip('skipping on windows')
+  console.log('skipping on windows')
   process.exit(0)
 }
 
-if (process.env.CITGM) {
-  // This looks like a some form of limitations of the CITGM test runner
-  // or the HW/SW we run it on. This file can hang on Node.js v18.x.
-  // The failure does not reproduce locally or on our CI.
-  // Skipping it is the only way to keep pino in CITGM.
-  // https://github.com/nodejs/citgm/pull/1002#issuecomment-1751942988
-  t.skip('Skipping on Node.js core CITGM because it hangs on v18.x')
-  process.exit(0)
-}
-
-function test (file) {
+function testFile (file) {
   file = join('fixtures', 'broken-pipe', file)
-  t.test(file, { parallel: true }, async ({ equal }) => {
+  test(file, async () => {
     const child = fork(join(__dirname, file), { silent: true })
     child.stdout.destroy()
 
     child.stderr.pipe(process.stdout)
 
     const res = await once(child, 'close')
-    equal(res, 0) // process exits successfully
+    assert.equal(res, 0) // process exits successfully
   })
 }
 
-t.jobs = 42
+testFile('basic.js')
+testFile('destination.js')
+testFile('syncfalse.js')
 
-test('basic.js')
-test('destination.js')
-test('syncfalse.js')
-
-t.test('let error pass through', ({ equal, plan }) => {
-  plan(3)
+test('let error pass through', async (t) => {
+  const plan = tspl(t, { plan: 3 })
   const stream = pino.destination({ sync: true })
 
   // side effect of the pino constructor is that it will set an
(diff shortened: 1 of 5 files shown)


test/broken-pipe.test.js
jsumners, line 1: @mcollina while I'm working on the rest of this, could you verify that this test is working as designed? I'm not entirely clear I understand it, but I have a suspicion it's trying to create file system contention.

test/esm/index.test.js
```diff
@@ -1,15 +1,14 @@
 'use strict'
 
-const t = require('tap')
 const semver = require('semver')
 
 const { isYarnPnp } = require('../helper')
 
 if (!semver.satisfies(process.versions.node, '^13.3.0 || ^12.10.0 || >= 14.0.0') || isYarnPnp) {
```
mcollina, line 7: You can just remove this block, esm is always supported

test/pkg/index.js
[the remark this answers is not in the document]
mcollina: Can you restore them? The successor is https://github.com/yao-pkg/pkg.

test/transport/big.test.js
```diff
@@ -26,18 +28,18 @@ test('eight million lines', { skip }, async ({ equal, comment }) => {
...
+  const toWrite = 8 * 1_000_000
   let count = 0
   await pipeline(createReadStream(destination), split(), new Writable({
     write (chunk, enc, cb) {
-      if (count % (toWrite / 10) === 0) {
-        comment(`read ${count}`)
-      }
+      // if (count % (toWrite / 10) === 0) {
+      //   comment(`read ${count}`)
+      // }
```
mcollina, line 39: just delete

test/transport/core.test.js
```diff
@@ -1,643 +0,0 @@
-'use strict'
```
mcollina, line 1: why was this deleted?

test/transport/core.test.ts
```diff
@@ -1,15 +1,17 @@
+import test from 'node:test'
```
mcollina, line 1: Why is this a ts file?

test/broken-pipe.test.js
```diff
@@ -1,47 +1,37 @@
...
 
-if (process.env.CITGM) {
-  // This looks like a some form of limitations of the CITGM test runner
-  // or the HW/SW we run it on. This file can hang on Node.js v18.x.
-  // The failure does not reproduce locally or on our CI.
-  // Skipping it is the only way to keep pino in CITGM.
-  // https://github.com/nodejs/citgm/pull/1002#issuecomment-1751942988
-  t.skip('Skipping on Node.js core CITGM because it hangs on v18.x')
-  process.exit(0)
-}
```
mcollina, line 22: I would keep skipping this on CITGM

test/errorKey.test.js
```diff
@@ -1,34 +0,0 @@
-'use strict'
```
mcollina, line 1: why was this removed?

test/pkg/index.js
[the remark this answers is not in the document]
jsumners: Do we really need it?

test/transport/core.test.js
jsumners: Because it was nested under `core.test.js` in my IDE. I thought that meant it was a generated file.

test/broken-pipe.test.js
jsumners: But v18 isn't supported?

test/errorKey.test.js
jsumners: It's renamed to `error-key.test.js`.

test/pkg/index.js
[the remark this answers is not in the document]
jsumners: I tried adding it back. First I get an error about no version satisfying v24. Skipping v24, I then get an error about `node:test` not existing. I don't want to spend any more time on this thing that I don't think we need. Please push up changes that work for you if you disagree.

test/broken-pipe.test.js
mcollina: I'm a bit unsure this test would actually run on the servers CITGM runs on, so it's better to keep skipping it.

test/transport/core.test.ts
mcollina: I think we should wort this out and keep only the js file if identical.

mcollina:
We should cut a 9.x branch before landing this to main, or start preparing a `next` branch with this change

jsumners:
> We should cut a 9.x branch before landing this to main, or start preparing a `next` branch with this change

I'll create a 9.x branch before I merge.

Puppo:
@jsumners, sorry for the question, but why did you rework all these files when we've already done them in the [next](https://github.com/pinojs/pino/tree/next) branch months before?

jsumners:
Simply losing track of what is where. I can't recall, but I don't think I saw we even have a `next` branch when I did this.

PR #2308, opened 2025-10-01
fix!: use safer types for censor function signature (#2307)

### LINKED ISSUE #2307: Use `unknown` instead of `any` for redactOptions.censor function definition
The latest release exposed the fact that a project I'm working on had defined an invalid pino `censor` function definition.  Specifically we had written something that incorrectly typed `value` as a `string`.  Because censor is typed as `any` this didn't get caught by tsc.

To be more concrete, [the censor signature](https://github.com/pinojs/pino/blob/9b6901f1a3ff524d3f984560193c5c1959ed2db6/pino.d.ts#L57) says value is type `any`:

```
        censor?: string | ((value: any, path: string[]) => any);
```

When in reality a censor author should explicitly type `value` as `unknown` because all censors should, at runtime, assume nothing about the type of `value` -- if the censor requires any type-specific behavior (e.g. string functionality) then it needs to first utilize a type guard to be safe.

This can be conveyed by pinot with an update to:

```
        censor?: string | ((value: unknown, path: string[]) => any);
```

Relatedly it probably wouldn't to be more explicit about the expected return type of a censor (e.g. the `=> any` could be `=> string | undefined` or `=> unknown`).

slifty:
This PR updates censor function signatures to require type safety (using `unknown` instead of `any`)

This improvement could result in type errors for anybody who had been using unsafe censor definitions.

Resolves #2307


mcollina:
We have tests for the types. Could you add one?

slifty:
@mcollina sorry about that; added two tests (one to ensure that return values are not type-constrained, one to ensure that paramater values *are* type constrained)

PR #2312, opened 2025-10-03
chore: add .npmignore to exclude unnecessary files

mcollina:
## Summary

Adds `.npmignore` file to exclude development and CI-related files from npm packages:
- `.github/` directory
- `docsify/` directory  
- `.nyc_output/` directory (test coverage)
- `coverage/` directory
- `*.png` files

## Impact

Reduces package size from **1.6 MB** to **141.3 kB** (91% reduction).

## Test Plan

- [x] Verified with `npm pack --dry-run` 
- [x] All tests pass

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,5 @@
+*.png
+.github/
+docsify/
+.nyc_output/
+coverage/


.npmignore
```diff
@@ -0,0 +1,5 @@
+*.png
+.github/
+docsify/
+.nyc_output/
+coverage/
```
jsumners, line 5: This is why we haven't done it before: these are already covered by `.gitignore`. With this PR merged, we'll have to maintain overlap. I wish `.npmignore` would augment instead of replace.

jsumners (approved):
I don't really like having to maintain the same list in two files. But I won't block it.

PR #2320, opened 2025-10-13
feat(types): use ThreadStream type from thread-stream

CHC383:
The `ThreadStream` type has been defined and evolved in the `thread-stream` package for years, I think it should be used here, unless there are any other reasons.

Besides, I feel https://github.com/pinojs/thread-stream/issues/24 can be closed.


CHC383:
@mcollina @jsumners Any thoughts on this type improvement? If it’s not needed, I can close the PR.

jsumners:
I have no interest in TypeScript things. @kibertoad may have an opinion.

mcollina (approved):
lgtm

Issue #2328, opened 2025-10-21
[Bug] Objects with generics cannot be logged


SpencerKaiser:
After the recent changes to support more types functionality in the context of `logFn` (#2254) Pino log functions can no longer be used to log objects with generic properties:

Version: 9.14.0

Repro:

``` typescript
const myFunc = <T extends string = never>(arg: Record<T, string>) => {
  const logger = pino()
  logger.info(arg) // Really nasty type error due to a circular reference
}
```

I know this wasn't intended to be a breaking change but it should probably be reverted on 9.x and revisited on 10.x.

rozzilla:
👀

rozzilla:
@samchungy, let me know what you think, but on my side I don't see how to make the example from @SpencerKaiser working without reverting [this change](https://github.com/pinojs/pino/pull/2254/files#diff-bb1e88b0612556147dae1eff5d068f3260d32f523c81d54efc3a4a4750a9b182R349) 🤔 

The issue with the code above seems related to the fact that `T` in `Record<T, string>` is a generic placeholder. For this reason, TypeScript cannot definitively prove that `Record<T, string>` is compatible with `Record<T, string> & {}` (where `{}` comes from `LogFnFields`) for all possible types of `T`.

@SpencerKaiser, I know it's not ideal, but as of now, the only way to avoid this type error is to remove the generic `T`:
```ts
const myFunc = (arg: Record<string, string>) => {
  const logger = pino()
  logger.info(arg)
}
```

SpencerKaiser:
@rozzilla unfortunately that's not an option for me ☹️ the generic is serving a very important purpose in that code so I unfortunately needed to roll back and lock the version. 

I don't imagine this will come up often for teams but I basically have three options unless yall revert and/or find a proper solution (I tried; I got nothing 🤷‍♂️):
- lock this version indefinitely (not sustainable)
- Use `patch-package` to revert the change yourself
- Migrate to an alternative 😢

samchungy:
I'll take a look this arvo! Sorry about that!

SpencerKaiser:
Zero worries, I'm sure it's a pretty niche use-case tbh... I can't imagine that would have come up during testing at all haha

PR #2329, opened 2025-10-22
fix(types): Correct conditional type handling for generic log function arguments

samchungy:
Resolves https://github.com/pinojs/pino/issues/2328

Allow generics to passthrough


mcollina (approved):
lgtm

PR #2372, opened 2026-01-09
fix: allow passing string, number, null for %o placeholder

### LINKED ISSUE #2371: Type is not assignable to type `object`
I really like the type-checking pino provides for the template strings, but the type for `%o` is too strict: it does not allow strings, numbers or even nullable objects. All of the following work at runtime, but are disallowed by type-checking (side note: I prefer using %o over %s for strings when I want clear string bounds):

```ts
log.info("uploaded %o", '"a file" and did something else')  // uploaded '"a file" and did something else'
log.info("uploaded %o", 123)  // uploaded 123
log.info("uploaded %o", null)  // uploaded null
```

Image

I think the only type that doesn't work is `undefined`, so `%o` could use `{} | null` instead of `object`?

rChaoz:
Closes #2371

### DIFF
--- a/test/types/pino.test-d.ts
+++ b/test/types/pino.test-d.ts
@@ -54,12 +54,19 @@ info('All primitive types %s %s %s %s %s %s %s', 'string', 123, true, BigInt(123
 declare const errorOrString: string | Error;
 info(errorOrString)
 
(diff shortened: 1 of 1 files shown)


mcollina:
Can you add a type test with tsd?

PR #2395, opened 2026-02-20
chore: upgrade borp to v1.0.0

Tony133:
Proposals:

- Upgrade `borp` dependency from `^0.21.0` to `^1.0.0` ( see here: https://github.com/mcollina/borp/releases/tag/v1.0.0 )
- Rename `core.transpiled.test.ts` to `core.transpiled.test.mjs`
- Update `core.transpiled.test.mjs`


test/transport/core.transpiled.test.mjs
```diff
@@ -1,20 +1,24 @@
...
 import * as url from 'node:url'
 
-import { watchFileCreated } from '../helper'
-import pino from '../../'
+import { fileURLToPath } from 'node:url'
+import { watchFileCreated } from '../helper.js'
+import pino from '../../pino.js'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
```
mcollina: Why?

mcollina (changes requested):

test/transport/core.transpiled.test.mjs
Tony133: > import.meta.dirname and import.meta.filename exists

I checked, I added them because these tests were failing  ( see screenshot )

screenshot-error


screenshot-test-error

in fact removing them gives me errors in the tests

mcollina: I'm a bit lost on this test tbh. I thought keeping the `.ts` extension was part of the actual test.

PR #2419, opened 2026-04-02
chore: upgrade typescript to v6.0.2

Tony133:
Proposal:

- Upgrade typescript v6.0.2 🔥

Changes made:

- Bump `typescript`dependency  from `~5.9.2` to `~6.0.2`
- Update `tsconfig.json`:
  - `target`: `es6` → `es2022`
  - `lib`: `["es2015", "dom"]` → `["es2022"]`
  - `module`: `commonjs` → `node16`
  - `moduleResolution`: `node16` 
  - `types`: `["node"]`
- Update `transpile.cjs`

This PR should resolve this PR: https://github.com/pinojs/pino/pull/2417

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,11 +1,13 @@
 {
   "compilerOptions": {
-    "target": "es6",
-    "lib": [ "es2015", "dom" ],
-    "module": "commonjs",
+    "target": "es2022",
+    "lib": ["es2022"],
+    "module": "node16",
+    "moduleResolution": "node16",
     "noEmit": true,
     "strict": true,
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "types": ["node"]
   },
   "exclude": [
     "./test/types/**/*.tst.*",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,11 +6,10 @@
     "moduleResolution": "node16",
     "noEmit": true,
     "strict": true,
-    "esModuleInterop": true,
     "types": ["node"]
   },
   "exclude": [
-    "./test/types/**/*.tst.*",
+    "./test/types/**/*",
     "./*.d.ts"
   ]
 }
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -1,14 +1,15 @@
...
+    "module": "node16",
+    "moduleResolution": "node16",
     "noEmit": true,
     "strict": true,
-    "esModuleInterop": true
+    "types": ["node"]
   },
   "exclude": [
-    "./test/types/**/*.tst.*",
+    "./test/types/**/*",
```
mrazauskas: Not sure about this change. See explanation in #2409 

Perhaps the simplest would be not to exclude anything (i.e. allow TypeScript to include all `.ts` file automatically) and to keep only a single `tsconfig.json` at the root? In a way, `pino/test/types/tsconfig.json` seems to be redundant.

(Or perhaps the opposite? The root `tsconfig.json` and `tsc` in the `test-types` script is not needed?)

Tony133: > Not sure about this change. See explanation in [#2409](https://github.com/pinojs/pino/pull/2409)
> 
> Perhaps the simplest would be not to exclude anything (i.e. allow TypeScript to include all `.ts` file automatically) and to keep only a single `tsconfig.json` at the root? In a way, `pino/test/types/tsconfig.json` seems to be redundant.
> 
> (Or perhaps the opposite? The root `tsconfig.json` and `tsc` in the `test-types` script is not needed?)



Thanks for the review! Yes, I confirm we can set the `tsconfig.json` like this:

```json
{
    "compilerOptions": { 
         "target": "es2022",
         "lib": ["es2022"],
         "module": "node16",
         "moduleResolution": "node16",
         "noEmit": true,
         "strict": true,
         "types": ["node"]
     },
     "exclude": [
         "./*.d.ts"
     ]
}

```

Tony133: see commit here: 2061f03

mcollina (approved):
lgtm

mcollina:
@mrazauskas ptal

mrazauskas:
lgtm

mcollina (approved):
lgtm

PR #2421, opened 2026-04-11
fix: add named `default` export

### LINKED ISSUE #2152: How to import pino with NodeNext module type
This problem was raised here: https://github.com/pinojs/pino/issues/543 but never resolved, and is still a problem.

When you configure a TypeScript project to use NodeNext modules, then you get a compiler error saying that pino is not a function. I tried many suggestions for different ways to write the import statement, and also using require instead, but so far nothing works.

The simplest reproduction is this source file
```typescript
import { default as pino } from 'pino'

export default pino()
```

My tsconfig.json is
```json
{
  "compilerOptions": {
    "rootDir": "./",
    "outDir": "./dist",
    "lib": ["es2020"],
    "target": "es2020",
    "module": "NodeNext",
    "moduleResolution": "NodeNext",
    "esModuleInterop": true,
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "types": ["node"],
    "baseUrl": "src",
  }
}
```

mrazauskas:
Close #2152, because I believe all cases should be covered.

This is a continuation of #2223 and #2258.

Instead of duplicating function declarations inside of the namespace (see #2258 for explanation), one can do:

```diff
  namespace pino {
    export type Level = "fatal" | "error" | "warn" | "info" | "debug" | "trace";
-   declare function pino(msg: string);
+   export { pino as default, pino }
  }
  declare function pino(msg: string);
  export = pino;
```

This allows removing duplicate code and enable the disabled `pinoCjsImport.default()` test. The rest of tests are passing. I added few more.

By the way, `fastify` is using similar [approach](https://github.com/fastify/fastify/blob/b06a196b694c0c7aed53976cd77456f1ad7d4c9f/fastify.d.ts#L205-L208). Also a detailed [explanation](https://github.com/arethetypeswrong/arethetypeswrong.github.io/blob/main/docs/problems/MissingExportEquals.md) can be found at `arethetypeswrong`.

### DIFF
--- a/test/types/pino-import.tst.cts
+++ b/test/types/pino-import.tst.cts
@@ -1,30 +1,34 @@
 import { expect } from 'tstyche'
 
 import * as pinoStar from '../../pino.js'
-import { type default as P, default as pino, pino as pinoNamed } from '../../pino.js'
+import pinoDefault, { default as pino, pino as pinoNamed } from '../../pino.js'
 import pinoCjsImport = require ('../../pino.js')
 const pinoCjs = require('../../pino.js')
 const { P: pinoCjsNamed } = require('pino')
 
 const log = pino()
-expect(log.info).type.toBe<P.LogFn>()
-expect(log.error).type.toBe<P.LogFn>()
+expect(log.info).type.toBe<pino.LogFn>()
+expect(log.error).type.toBe<pino.LogFn>()
 
+expect(pinoDefault()).type.toBe<pino.Logger>()
 expect(pinoNamed()).type.toBe<pino.Logger>()
-expect(pinoNamed()).type.toBe<P.Logger>()
 expect(pinoStar.default()).type.toBe<pino.Logger>()
 expect(pinoStar.pino()).type.toBe<pino.Logger>()
-// expect(pinoCjsImport.default()).type.toBe<pino.Logger>()
+expect(pinoCjsImport()).type.toBe<pino.Logger>()
+expect(pinoCjsImport.default()).type.toBe<pino.Logger>()
 expect(pinoCjsImport.pino()).type.toBe<pino.Logger>()
 expect(pinoCjsNamed()).type.toBe<any>()
 expect(pinoCjs()).type.toBe<any>()
-expect(pinoNamed.stdTimeFunctions.isoTimeNano).type.toBe<P.TimeFn>()
-expect(pinoNamed.stdTimeFunctions.isoTimeNano()).type.toBe<string>()
 
-const levelChangeEventListener: P.LevelChangeEventListener = (
-    lvl: P.LevelWithSilent | string,
+expect(pinoDefault.stdTimeFunctions.isoTimeNano()).type.toBe<string>()
+expect(pinoNamed.stdTimeFunctions.isoTimeNano()).type.toBe<string>()
+expect(pinoStar.stdTimeFunctions.isoTimeNano()).type.toBe<string>()
+expect(pinoCjsImport.stdTimeFunctions.isoTimeNano()).type.toBe<string>()
+
+const levelChangeEventListener: pino.LevelChangeEventListener = (
+    lvl: pino.LevelWithSilent | string,
     val: number,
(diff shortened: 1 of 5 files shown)


test/types/pino-import.tst.cts
```diff
@@ -1,30 +1,34 @@
 import { expect } from 'tstyche'
 
 import * as pinoStar from '../../pino.js'
-import { type default as P, default as pino, pino as pinoNamed } from '../../pino.js'
+import pinoDefault, { default as pino, pino as pinoNamed } from '../../pino.js'
```
mrazauskas, line 4: Here the default import was missing. Sounds like a good candidate for a test.

I removed `type default as P` for brevity, because `default as pino` is the same. Without the `type` modifier, but that does not change anything in this case (as far as I can understand). Same namespace is available as `P` and `pino`.

test/types/pino-import.tst.cts
```diff
@@ -1,30 +1,34 @@
...
-expect(pinoNamed()).type.toBe<P.Logger>()
 expect(pinoStar.default()).type.toBe<pino.Logger>()
 expect(pinoStar.pino()).type.toBe<pino.Logger>()
-// expect(pinoCjsImport.default()).type.toBe<pino.Logger>()
+expect(pinoCjsImport()).type.toBe<pino.Logger>()
+expect(pinoCjsImport.default()).type.toBe<pino.Logger>()
 expect(pinoCjsImport.pino()).type.toBe<pino.Logger>()
 expect(pinoCjsNamed()).type.toBe<any>()
 expect(pinoCjs()).type.toBe<any>()
-expect(pinoNamed.stdTimeFunctions.isoTimeNano).type.toBe<P.TimeFn>()
```
mrazauskas, line 21: Line is removed, because `TimeFn` is not exported (not sure how this currently works): https://github.com/pinojs/pino/blob/40cc80c4f59ae8695a5d25073d032895888ce299/pino.d.ts#L26

test/types/pino-import.tst.js
mrazauskas, line 1: The `require()` imports resolve to `any` in `.ts` files. So I moved tests here, because in `.js` files they are typed. This means that a line like `pino.log` would be reported as an error:

Image

mcollina: why is is resolving to any?

mrazauskas: I was trying to say that `const ... = require()` syntax in a `.js` file replicates the usage of APIs from JavaScript file. This is why I added these new test as a `.js` file and wanted to explain the reason.

---

Here is the `any` pitfall. When `const ... = require()` is used in a `.ts` file, TypeScript interprets `require()` as just another function. Its typings are defined in `@types/node` as `require(id: string): any`. It will always return `any`. See these imports in current test files:

https://github.com/pinojs/pino/blob/40cc80c4f59ae8695a5d25073d032895888ce299/test/types/pino-import.tst.cts#L6-L7

and here are the tests:

https://github.com/pinojs/pino/blob/40cc80c4f59ae8695a5d25073d032895888ce299/test/types/pino-import.tst.cts#L19-L20

---

Note that the second import is `require('pino')`, which should error with "Cannot find module 'pino' or its corresponding type declarations", but it doesn’t. This is correct because `require()` is a function that takes a `string`. Also `P` is not exported, but there is no error, because it exists on the `any` return type.

In a `.ts` file only `import ... = require ()` syntax is interpreted as import syntax (and would be [transpiled by TypeScript](https://www.typescriptlang.org/docs/handbook/modules/reference.html#export--and-import--require)). But this is not the same as `const ... = require()`. For instance, `import { pino } = require ('../../pino.js')` is not allowed, but `const { pino } = require ('../../pino.js')` works fine in a `.js` file.

These new tests avoid the pitfall of `require(id: string): any` definition and test the exports as mentioned here:

https://github.com/pinojs/pino/blob/40cc80c4f59ae8695a5d25073d032895888ce299/pino.d.ts#L874-L877

mcollina (approved):
lgtm

PR #2423, opened 2026-04-16
chore(deps): remove `pre-commit`

kshitijanurag:
Remove `pre-commit` because it is no longer actively maintained.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
   "bin": {
     "pino": "./bin.js"
   },
-  "precommit": "test",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/pinojs/pino.git"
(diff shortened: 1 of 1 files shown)


jsumners:
As before, remove it completely. We no longer want to use post install scripts.

kshitijanurag:
> As before, remove it completely. We no longer want to use post install scripts.

Removed

jsumners:
> > As before, remove it completely. We no longer want to use post install scripts.
> 
> Removed

_Almost_ complete. Line 41 should also be removed (GH won't let me add a suggestion on an unaffected line).

kshitijanurag:
> > > As before, remove it completely. We no longer want to use post install scripts.
> > 
> > 
> > Removed
> 
> _Almost_ complete. Line 41 should also be removed (GH won't let me add a suggestion on an unaffected line).

Thanks for pointing out, removed it.

jsumners (approved):

PR #2425, opened 2026-04-24
docs: use @hono/structured-logger for Pino with Hono

brunorodmoreira:
## What

Replaces the `pino-http` example in the **Pino with Hono** section of `docs/web.md` with the newly released [`@hono/structured-logger`](https://github.com/honojs/middleware/blob/main/packages/structured-logger/README.md) middleware.

## Why

The current snippet wires `pino-http` (an Express-style middleware) into Hono by:

- reaching into `@hono/node-server` internals (`c.env.incoming` / `c.env.outgoing`),
- manually forwarding `hono/request-id` into the request object,
- wrapping the call in a `new Promise` to bridge the Express `(req, res, next)` signature.

`@hono/structured-logger` is a first-class Hono middleware with pino as one of its documented integrations. The recipe becomes:

- shorter and easier to read,
- no longer coupled to `@hono/node-server` internals,
- portable across every runtime Hono supports (Node, Bun, Deno, Workers, Lambda, Edge),
- still request-scoped via `rootLogger.child({ requestId })`.

## Scope

Docs-only. No code changes, no dep changes in the repo itself.

### DIFF
--- a/docs/web.md
+++ b/docs/web.md
@@ -274,28 +274,25 @@ See the [pino-http README](https://npm.im/pino-http) for more info.
 ## Pino with Hono
 
 ```sh
-npm install pino pino-http hono
+npm install pino @hono/structured-logger hono
 ```
 
 ```js
 import { serve } from '@hono/node-server';
 import { Hono } from 'hono';
 import { requestId } from 'hono/request-id';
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #2436, opened 2026-05-11
fix: handle read-only prepareStackTrace

### LINKED ISSUE #2071: --frozen-intrinsics: Cannot assign to read only property 'prepareStackTrace' of function 'function Error() { [native code] }'

raashish1601:
Fixes #2071.

## Summary
- keep using `util.getCallSites()` / `util.getCallSite()` when available
- make the legacy `Error.prepareStackTrace` fallback return `undefined` instead of throwing when `prepareStackTrace` is read-only
- add a regression test that simulates the fallback path without `util.getCallSites()` and a read-only `Error.prepareStackTrace`

## Validation
- `npm run lint`
- `node --test test/caller.test.js`
- `npx borp --timeout 60000 test/caller.test.js`
- `node --frozen-intrinsics -e "require('./pino')(); console.log('ok')"`

`npm test` starts successfully through lint, then fails on this Windows shell in `npm run transpile` because `test/fixtures/ts/transpile.cjs` shells out to `mv`, which is not available in Windows CMD/PowerShell.

### DIFF
--- a/test/caller.test.js
+++ b/test/caller.test.js
@@ -2,6 +2,7 @@
 
 const test = require('node:test')
 const assert = require('node:assert')
+const util = require('node:util')
 const loop = require('./fixtures/caller-loop.js')
 
 test('returns a callstack of absolute paths', () => {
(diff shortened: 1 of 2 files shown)


test/caller.test.js
```diff
@@ -18,3 +19,33 @@ test('returns a callstack of absolute paths', () => {
   assert.match(callers[6], /^[/\\]fixtures[/\\]caller-loop\.js$/)
   assert.match(callers[7], /^[/\\]caller\.test\.js$/)
 })
+
+test('does not throw when prepareStackTrace is read-only', () => {
```
mcollina, line 23: Can you move this to its own file? So we don't have a problem with tests added later.

mcollina (changes requested):

PR #2437, opened 2026-05-11
Fix pipeline transport worker options

### LINKED ISSUE #1856: Worker options for custom transport in pipeline not passed
Hi, thanks for your work on pino!
I am trying to create a pipelined transport and pass it a message port, like so:
```typescript
import pino, { TransportPipelineOptions } from 'pino';
import { MessageChannel } from 'node:worker_threads';

const { port1, port2 } = new MessageChannel();
const logger = pino({
    transport: {
      pipeline: [
        {
          target: './transport.ts',
          worker: {
            workerData: {
              port: port1,
            },
            transferList: [port1],
          },
        },
        {
          target: 'pino-pretty',
          options: {
            sync: true,
            translateTime: 'HH:MM:ss Z',
            ignore: 'pid,hostname',
          },
        },
      ],
    } satisfies TransportPipelineOptions,
  });
```

When I run this code, it throws this error:
```

    TypeError: Object that needs transfer was found in message but not listed in transferList

      32 |               port: port1,
      33 |             },
    > 34 |             transferList: [port1],
         |                                ^
      35 |           },
      36 |         },
```
However, if I do not use a pipeline (omitting the pino-pretty transport) and create the transport like so
```typescript
 pino(
    pino.transport({
      target: './transport.ts',
      worker: {
        workerData: {
          port: port1,
        },
        transferList: [port1],
      },
    }),
  );
```
everything works correctly. 

Is there potentially an option I'm not passing or passing incorrectly to make this work with a pipeline?
Thanks!

raashish1601:
Fixes #1856.

This updates pipeline transport setup so worker options defined on individual pipeline transport entries are not serialized directly into the pino worker configuration. Instead, `workerData` is forwarded to that transport's options and nested `transferList` entries are carried up to the actual worker thread options.

That lets transports in a pipeline receive worker data containing transferable values such as `MessagePort` without triggering `Object that needs transfer was found in message but not listed in transferList`.

Validation:
- `npx borp --timeout 60000 test/transport/pipeline.test.js`
- `node --test test/transport/pipeline.test.js`
- `npm run lint`

### DIFF
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -176,6 +176,43 @@ function buildStream (filename, workerData, workerOpts, sync, name) {
   return stream
 }
 
+function setupTransportWorkerOptions (dest, workerOpts) {
+  const { worker, ...transportOptions } = dest
+
+  if (!worker) {
+    return { dest: transportOptions, workerOpts }
+  }
+
+  const { workerData, transferList, ...streamWorkerOpts } = worker
+
+  if (workerData !== undefined) {
+    transportOptions.options = {
+      ...transportOptions.options,
+      workerData
+    }
+  }
+
+  for (const key of Object.keys(streamWorkerOpts)) {
+    if (workerOpts[key] === undefined) {
+      workerOpts = {
+        ...workerOpts,
+        [key]: streamWorkerOpts[key]
+      }
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #2438, opened 2026-05-12
fix(browser): serialize err.cause in errWithCause

### LINKED ISSUE #2057: [bug] in the browser, `pino.stdSerializers.errorWithCause` does not serialize `cause` prop
When using `Pino.stdSerializers.errWithCause` in the browser the "cause" property of errors are not, in fact, serialized.

Video demonstrating the behavior.
https://www.loom.com/share/916e9677c50946f99a5772adef18e992?sid=e775a176-ceed-48a1-a032-821f8af74314

The problem appears to be that, in the browser, a `pino/browser.js` module is being loaded with a different implementation of `stdSerializers.errWithCause` which doesn't actually serialize the `cause` property.

raashish1601:
## Summary

In the browser build, `stdSerializers.errWithCause` was aliased to the plain error serializer. Because `Error#cause` is not enumerable, the browser serializer dropped nested causes even though callers explicitly selected the cause-aware serializer.

This adds a browser `errWithCause` serializer that recursively serializes nested `Error` causes while still preserving non-Error cause values.

Fixes #2057.

## Validation

- `npm run lint`
- `node test/browser.test.js`

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -39,7 +39,7 @@ const stdSerializers = {
   req: mock,
   res: mock,
   err: asErrValue,
-  errWithCause: asErrValue
+  errWithCause: asErrValueWithCause
 }
 function levelToValue (level, logger) {
   return level === 'silent'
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
CI is red

PR #2452, opened 2026-06-12
fix: escape child binding keys

mcollina:
## Summary

- serialize child and base binding keys with the existing JSON string encoder
- preserve quotes, backslashes, and control characters without breaking NDJSON framing
- add a regression test for a crafted child binding key containing quotes and newlines

## Tests

- `npm test`
- `node --test test/escaping.test.js`
- `git diff --check`

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -258,7 +258,7 @@ function asChindings (instance, bindings) {
       value = serializers[key] ? serializers[key](value) : value
       value = (stringifiers[key] || wildcardStringifier || stringify)(value, stringifySafe)
       if (value === undefined) continue
-      data += ',"' + key + '":' + value
+      data += ',' + asString(key) + ':' + value
     }
   }
   return data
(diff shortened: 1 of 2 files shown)


lib/tools.js
```diff
@@ -258,7 +258,8 @@ function asChindings (instance, bindings) {
       value = serializers[key] ? serializers[key](value) : value
       value = (stringifiers[key] || wildcardStringifier || stringify)(value, stringifySafe)
       if (value === undefined) continue
-      data += ',"' + key + '":' + value
+      const strKey = asString(key)
```
jsumners: The reason I didn't do it this way back when the report was originally filed is because it's the same code in multiple places, and particularly easy to overlook. But centralizing the overall algorithm was onerous. So 🤷‍♂️

jsumners (approved):

jsumners (approved):

PR #2455, opened 2026-06-21
fix(multistream): isolate per-stream write() errors so remaining streams still receive log data

LeSingh1:
## Problem

`pino.multistream().write()` iterates over all registered streams and calls `stream.write(data)` on each. There is no error isolation around individual writes: if any stream's synchronous `write()` throws, the iteration loop aborts immediately and **every stream that comes after the failing one in the sorted list never receives the log line**. Logs are silently dropped with no indication that anything went wrong.

Minimal reproduction:

```js
const pino = require('pino')

let received = 0
const goodStream = { write () { received++ } }
const badStream  = { write () { throw new Error('boom') } }

const multi = pino.multistream([
  { level: 'info', stream: badStream  },   // sorted first (same level)
  { level: 'info', stream: goodStream }
])

const log = pino({ level: 'info' }, multi)
log.info('hello')

console.log(received) // prints 0 — goodStream never received the line!
```

## Fix

Wrap each `stream.write(data)` call in a `try/catch` inside `multistream.write()`. The caught error is surfaced via a dedicated `EventEmitter` (`errorEmitter`) so callers can observe it with `res.on('error', fn)` — without interfering with the existing `res.emit()` method, which propagates events like `'message'` down to every child stream for pino's internal config protocol.

When no `'error'` listener is registered the error is swallowed, preserving the existing non-throwing contract of `multistream.write()`.

**What changed in `lib/multistream.js`:**
- Added an internal `errorEmitter` (`EventEmitter`) instance.
- Exposed `on` / `once` / `removeListener` / `off` on `res` delegating to `errorEmitter`.
- Wrapped `stream.write(data)` in `try/catch`; emits `'error'` on `errorEmitter` when there is at least one listener, otherwise swallows.

## New tests (in `test/multistream.test.js`)

1. **`multistream.write continues writing to remaining streams when one stream throws synchronously`** — asserts that a stream ordered after a throwing stream still receives all log lines and that an `'error'` event is emitted for each throwing write.
2. **`multistream.write does not throw when a stream write() throws and no error listener is registered`** — asserts `doesNotThrow` when no listener is attached.

## Test results

All 32 multistream tests pass (including the 2 new ones), and the full basic + serializer suites (126 tests) continue to pass.

## Checklist

- [x] Existing tests pass
- [x] New regression tests added
- [x] No change to the `res.emit()` stream-propagation behaviour
- [x] Backward compatible (callers that do not attach a `res.on('error')` listener see no behaviour change)

### DIFF
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -18,29 +18,36 @@ function multistream (streamsArray, opts) {
     })
   }
 
-  // A dedicated EventEmitter for surfacing per-stream write errors to callers.
-  // We keep it separate so that the existing `res.emit()` method (which
-  // propagates events like 'message' down to every child stream) is
-  // unaffected.
-  const errorEmitter = new EventEmitter()
+  const res = new EventEmitter()
 
-  const res = {
-    write,
-    add,
-    remove,
-    emit,
-    on: errorEmitter.on.bind(errorEmitter),
-    once: errorEmitter.once.bind(errorEmitter),
-    removeListener: errorEmitter.removeListener.bind(errorEmitter),
-    off: errorEmitter.off.bind(errorEmitter),
-    flushSync,
-    end,
-    minLevel: 0,
-    lastId: 0,
-    streams: [],
-    clone,
-    [metadata]: true,
-    streamLevels
+  res.write = write
+  res.add = add
+  res.remove = remove
+  res.flushSync = flushSync
+  res.end = end
+  res.minLevel = 0
+  res.lastId = 0
+  res.streams = []
+  res.clone = clone
+  res[metadata] = true
+  res.streamLevels = streamLevels
+
+  // res is now a full EventEmitter, but multistream has always also fanned
+  // every emitted event out to its child streams (for example the 'message'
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
I think the best outcome would be to move the res object to
be a full EventEmitter instead.

lib/multistream.js
```diff
@@ -17,19 +18,36 @@ function multistream (streamsArray, opts) {
...
+  res.emit = function (...args) {
+    const [event] = args
+    EventEmitter.prototype.emit.apply(this, args)
+    if (event !== 'error') {
+      for (const { stream } of this.streams) {
+        if (typeof stream.emit === 'function') {
+          stream.emit(...args)
+        }
+      }
+    }
```
mcollina, line 50: Ahum, this doesn't look a good idea after all. Do we need the return value to have an emit property?

If we can't remove this, the previous approach was in fact better. My bad.

mcollina (changes requested):

PR #2457, opened 2026-06-25
fix: log objects with an own __proto__ key instead of crashing

spokodev:
## Problem

Logging an object that has an own enumerable `__proto__` key crashes the logger and drops the log line:

```js
const pino = require('pino')
pino().info(JSON.parse('{"__proto__":{"x":1},"user":"bob"}'), 'incoming request')
// TypeError: serializers[key] is not a function
// nothing is written
```

Such an object is the normal result of parsing untrusted JSON. `JSON.parse('{"__proto__":{...}}')` defines `__proto__` as an own enumerable data property rather than setting the prototype, so any request body, queue message, or webhook payload routed straight into a log call can trigger it. `JSON.stringify` serializes the same object without issue, so the expected behavior is a single valid NDJSON line.

## Root cause

The serializers and stringifiers objects are merged in a way that leaves `Object.prototype` on their prototype chain. When the log object loop reaches `key === '__proto__'`, `serializers[key]` resolves to `Object.prototype`, which is truthy but not a function:

```js
if (serializers[key]) {        // Object.prototype is truthy
  value = serializers[key](value)  // throws: not a function
}
```

The same pattern affects the stringifier lookup in `_asJson`, the message key lookup, and the serializer and stringifier lookups for child bindings in `asChindings`.

## Fix

Guard each lookup with `typeof ... === 'function'` so inherited prototype members are ignored and only real serializer or stringifier functions are invoked. The `__proto__` key is then serialized as a plain data property, matching `JSON.stringify`.

## Verification

Added three regression tests in `test/basic.test.js` covering the top-level object path, the redaction path, and the child bindings path. Each fails on `main` with `TypeError: serializers[key] is not a function` and passes with this change.

```
$ npm run lint   # clean
$ npx borp --pattern test/basic.test.js   # all green
```

The full suite is green other than one pre-existing thread-stream async-flush test that times out waiting for a temp file, which fails the same way on an unmodified checkout in this environment and is unrelated to this change.

### DIFF
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -166,13 +166,13 @@ function _asJson (obj, msg, num, time) {
   for (const key in obj) {
     value = obj[key]
     if (Object.prototype.hasOwnProperty.call(obj, key) && value !== undefined) {
-      if (serializers[key]) {
+      if (typeof serializers[key] === 'function') {
         value = serializers[key](value)
       } else if (key === errorKey && serializers.err) {
         value = serializers.err(value)
       }
 
-      const stringifier = stringifiers[key] || wildcardStringifier
+      const stringifier = typeof stringifiers[key] === 'function' ? stringifiers[key] : wildcardStringifier
 
       switch (typeof value) {
         case 'undefined':
@@ -201,8 +201,8 @@ function _asJson (obj, msg, num, time) {
 
   let msgStr = ''
   if (msg !== undefined) {
-    value = serializers[messageKey] ? serializers[messageKey](msg) : msg
-    const stringifier = stringifiers[messageKey] || wildcardStringifier
+    value = typeof serializers[messageKey] === 'function' ? serializers[messageKey](msg) : msg
+    const stringifier = typeof stringifiers[messageKey] === 'function' ? stringifiers[messageKey] : wildcardStringifier
 
     switch (typeof value) {
       case 'function':
@@ -255,8 +255,9 @@ function asChindings (instance, bindings) {
       bindings.hasOwnProperty(key) &&
       value !== undefined
     if (valid === true) {
-      value = serializers[key] ? serializers[key](value) : value
-      value = (stringifiers[key] || wildcardStringifier || stringify)(value, stringifySafe)
+      value = typeof serializers[key] === 'function' ? serializers[key](value) : value
+      const stringifier = typeof stringifiers[key] === 'function' ? stringifiers[key] : (wildcardStringifier || stringify)
+      value = stringifier(value, stringifySafe)
       if (value === undefined) continue
       data += ',"' + key + '":' + value
     }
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

jsumners (approved):

mcollina:
CI is very red

spokodev:
The red is the `thread-stream async flush should call the passed callback` test — it fails with "Promise resolution is still pending but the event loop has already resolved". It reproduces on a clean `main` under Node 22, so it isn't from this PR (which only touches the serializer/stringifier guards in `lib/tools.js`). Looks like a pre-existing issue in that test's async teardown.

luantaraschi:
Ran into this from a different direction and thought the measurements were worth leaving here, since they touch the lines you are changing.

The `typeof ... === 'function'` guard closes the `__proto__` case, because `Object.prototype` is an object. It does not close the rest, because the other inherited members really are functions, so they pass the guard and get called as serializers and stringifiers.

On `main` today, with no configuration at all, just logging an object whose field carries one of those names:

| field | result |
|---|---|
| `toString` | log line is not valid JSON: `"toString":[object Undefined]` |
| `constructor` | log line is not valid JSON: `"constructor":[object Object]` |
| `valueOf` | throws `TypeError: Cannot convert undefined or null to object` |
| `hasOwnProperty` | throws the same |
| `propertyIsEnumerable` | throws the same |
| `toLocaleString` | throws `TypeError: Object.prototype.toLocaleString called on null or undefined` |
| `isPrototypeOf` | valid JSON, wrong value (`false`) |

I re-ran the first two rows with your patch applied on top and they come out the same, which is what pointed me at the guard rather than at something further down.

If it helps, the shape that covers all of them is giving the two maps a null prototype where they are built, rather than guarding every read: `stringifiers` comes from `redaction()` and from the `{}` literal in `pino.js`, and `serializers` from `instance[serializersSym]`. I have a change for the `redaction()` half in #2495, which stays out of `lib/tools.js` precisely so it does not collide with this.

Entirely your call whether to fold the rest into this PR or leave it for a follow-up once this lands. Happy to send the follow-up if you would rather keep this one focused, and happy to leave it alone if you have a different fix in mind.

mcollina:
Happy if you'd like to send a fresh version of this pr

spokodev:
Thanks @luantaraschi, great breakdown. I've sent a fresh version that fixes this at the root instead of guarding each read.

Two build sites were the cause: the serializers merge in `createArgsNormalizer` (`Object.assign({}, ...)`) and the no-redact `stringifiers = {}` in `pino.js`, both plain objects, so `serializers['__proto__']` and `stringifiers['toString']` resolved to inherited members. Both now use `Object.create(null)` and the read guards are gone. The tests cover the full set of member names, redacted and not.

#2495 still covers a case this one doesn't touch: a redact path whose first segment is a prototype member, which throws while building the redaction shape in `lib/redaction.js`. I left that file alone so the two don't collide.

mcollina (approved):
lgtm

PR #2471, opened 2026-07-12
fix: add setBindings to the browser logger

### LINKED ISSUE #2021: [bug] logger.`setBindings` undefined when using esm exported pino
Replica:
```ts
import { pino } from "pino" 

export const logger = pino({
  browser: {
    serialize: true,
    asObject: true,
  },
  customLevels: {
    d: 1,
    i: 2,
    w: 3,
    // ...
  }
})
```
produces `TypeError: logger.setBindings is not a function`

Alternative:
```diff
- import { pino } from "pino"
+ import pino from "pino" 
```
Use default export fixes the problem for now.

ankit1324:
Fixes #2021

## Problem

The browser logger (`browser.js`) never implemented `setBindings`, even though the TypeScript types declare it on `LoggerExtras` — so any code path that resolves the `browser` field of `package.json` throws `TypeError: logger.setBindings is not a function`. (To answer the question in the issue thread: it's a JS bug — a missing method in the browser implementation, not a typings problem. The named/default import distinction in the report was a red herring; both exports point at the same function, what differs is whether the bundler picked `pino.js` or `browser.js`.)

## Fix

Implement `setBindings` on the browser logger:

- merges the new bindings into the logger's `bindings` (children inherit the method through the prototype chain and merge into their own bindings),
- mirrors the new bindings into `_logEvent.bindings` so `browser.transmit` receives them, matching what `child()` does,
- rebuilds the level functions (same mechanism as the level setter) so subsequent log calls prepend the updated bindings.

Consistent with Node semantics: like `asChindings` in `lib/proto.js`, updating a parent's bindings does not retroactively affect already-created children.

## Testing

New `test/browser-set-bindings.test.js` (tape, same style as the other browser tests): method exists, bindings appear in subsequent logs, repeated calls merge, child loggers merge with their own bindings, works with `customLevels` (the issue's repro), and bindings are transmitted. All existing browser tests and lint pass.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -166,6 +166,15 @@ function pino (opts) {
   logger._serialize = serialize
   logger._stdErrSerialize = stdErrSerialize
   logger.child = function (...args) { return child.call(this, setOpts, ...args) }
+  logger.setBindings = function (newBindings) {
+    this.bindings = Object.assign({}, this.bindings, newBindings)
+    if (this._logEvent) {
+      this._logEvent.bindings.push(newBindings)
+    }
+    // reassigning the level rebuilds the log functions so the updated
+    // bindings are prepended to subsequent log calls
+    this.level = this._level
+  }
 
   if (transmit) logger._logEvent = createLogEventShape()
 
--- a/test/browser-set-bindings.test.js
+++ b/test/browser-set-bindings.test.js
@@ -1,0 +1,99 @@
+'use strict'
+const test = require('tape')
+const pino = require('../browser')
+
+test('setBindings is a function on the root logger', ({ end, is }) => {
+  const instance = pino({ browser: {} })
+
+  is(typeof instance.setBindings, 'function')
+  end()
+})
+
+test('setBindings adds bindings to subsequent log calls', ({ end, is }) => {
+  const instance = pino({
+    browser: {
+      asObject: true,
+      write: function (o) {
+        is(o.answer, 42)
+        is(o.msg, 'test')
+      }
+    }
+  })
+
+  instance.setBindings({ answer: 42 })
+  instance.info('test')
+  end()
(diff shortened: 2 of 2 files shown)


ankit1324:
For transparency: after opening this I noticed #2403 tackles the same issue and predates this PR. I'd normally close mine in favor of the earlier one, but the implementations differ in ways that matter, so leaving this open for the maintainers to pick:

- #2403 **replaces** the logger's bindings on each call; this PR **merges** them (`Object.assign`), matching Node's `setBindings`/`asChindings` semantics.
- This PR also mirrors the new bindings into `_logEvent.bindings` so `browser.transmit` receives them (same as `child()` does); #2403 doesn't cover transmit.
- Additional tests here: merging across calls, child loggers, `customLevels` (the original repro), and transmit.

Happy to close either way if the maintainers prefer building on #2403.

ankit1324:
Re the red CI: the failing test (`thread-stream async flush should call the passed callback`) fails identically on a clean `main` checkout — it's unrelated to this PR's `browser.js` changes. I bisected it to the `thread-stream` 4.1.0 release floating in via the `^4.0.0` range and filed #2473 with details (last green main CI was 2026-05-08, right before 4.1.0 was published). The browser tests here all pass.

mcollina:
Please rebase

ankit1324:
Rebased on main (now includes the #2470 flush-test fix). Browser tests and the previously failing transport test both pass locally.

ankit1324:
@mcollina Rebase done and CI is green now (14 passing, the transport failure is gone with #2470 on main). Ready for review whenever you get a chance — happy to address any feedback.

mcollina (approved):
lgtm

PR #2472, opened 2026-07-13
docs: clarify Node.js version support policy in LTS doc

### LINKED ISSUE #815: Add more clarification to the LTS doc
I like the language about legacy versions and how they are are supported in https://github.com/googleapis/google-cloud-node#supported-nodejs-versions . I think we could adopt some form of it.

youdie006:
Closes #815.

Extends `docs/lts.md` with a **Supported Node.js versions** section, per the discussion on the issue:
- Adapts the google-cloud-node "Supported Node.js Versions" phrasing that @jsumners liked.
- States explicitly (per @mcollina) that support for a Node.js line is only dropped in a new pino **major**.
- Adds best-effort EOL guidance and how to stay on an EOL Node by pinning to the last pino major that supported it (grounded in the doc's own Schedule table, e.g. `"pino": "8.x"` for Node 14).

Existing content is untouched; the change is additive.

---

*AI disclosure: this PR was drafted with Claude Code (AI-assisted). The doc content is grounded in the repo's existing LTS policy and Schedule table (no invented policy).*

### DIFF
--- a/docs/lts.md
+++ b/docs/lts.md
@@ -37,6 +37,35 @@ A "month" is defined as 30 consecutive days.
 
 [semver]: https://semver.org/
 
+<a name="supported-nodejs"></a>
+
+### Supported Node.js versions
+
+Pino follows the [Node.js release schedule][node-release]. A Pino release line
+is tested against, and supports, every Node.js release line that is in its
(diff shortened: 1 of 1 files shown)


mcollina (approved):
Lgtm

PR #2474, opened 2026-07-13
fix(multistream): resolve string level labels in write()

### LINKED ISSUE #1413: Configuration ignored when multiple transport targets

xianjianlf2:
## Problem

When a source relays a level label string via the metadata protocol (e.g. `pino-abstract-transport` forwarding lines produced with a `formatters.level` option), `lastLevel` is a string such as `"error"` instead of a number. The `dest.level <= level` comparisons in `write()` then evaluate against `NaN`, so the record matches no stream and is silently dropped.

## Fix

Normalize a string `lastLevel` through `streamLevels` before comparing, mirroring the resolution already done for stream levels in `add()`.

## Testing

Added a regression test that drives `multistream` via the metadata protocol with a string `lastLevel` (`"error"`) and asserts the record still reaches both the `debug` (`<= error`) and `error` streams, exactly as it would when `lastLevel` is the numeric `50`.

Closes #1413

### DIFF
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -48,7 +48,13 @@ function multistream (streamsArray, opts) {
   // we can exit early because the streams are ordered by level
   function write (data) {
     let dest
-    const level = this.lastLevel
+    // `lastLevel` is normally the numeric level, but a source may relay a
+    // level label string (e.g. pino-abstract-transport forwarding lines
+    // produced with a `formatters.level` option). Resolve it to a number so
+    // the level comparisons below don't silently drop the record.
+    const level = typeof this.lastLevel === 'string'
+      ? (this.streamLevels || streamLevels)[this.lastLevel]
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #2482, opened 2026-07-26
fix: invoke flush callback when transport worker is unref'd

### LINKED ISSUE #2481: logger.flush(cb) never invokes the callback on an idle event loop
When `logger.flush(cb)` is called on a transport-backed logger and the event loop has no other pending work, the callback is never invoked. The process exits cleanly with code 0, and the caller has no way to detect that the flush completion was silently swallowed.

The docs explicitly state: *"If there is a need to wait for the logs to be flushed, a callback should be used."* (`docs/api.md`), but the callback contract is not honored in this case.

## Reproduction

```js
'use strict'
const pino = require('pino')
const fs = require('fs')

const dest = '/tmp/pino-flush-repro.log'
fs.writeFileSync(dest, '')

  target: 'pino/file',
  options: { destination: dest }
})

logger.info({ marker: 'first-log' })

let callbackFired = false
logger.flush((err) => {
  callbackFired = true
  console.log('CALLBACK FIRED, err:', err)
})

process.on('exit', (code) => {
  console.log(`exit code=${code}, callback fired=${callbackFired}`)
})
```

**Actual output:**

exit code=0, callback fired=false

The log line does get written to disk (worker completes the work), but the main thread is never notified.

**Expected:** callback is invoked (with or without error), consistent with the Node.js callback contract.

## Workaround

Keeping the event loop alive during the flush lets the callback fire:

```js
const keepAlive = setInterval(() => {}, 100)
logger.flush((err) => {
  clearInterval(keepAlive)
  // ...
})
```

The `transport` worker is deliberately `unref`'d once ready (see `lib/transport.js`), so an idle loop drains before the worker delivers the flush-completion message back to the main thread.

## Related

- Test `thread-stream async flush should call the passed callback` (`test/transport/sync-false.test.js`) was recently patched to work around the same behavior on the test side (#2470 / c4d39d4), keeping the event loop alive during the awaited flush. That commit fixes the test but the underlying behavior in production remains: `logger.flush(cb)` on an otherwise idle event loop silently swallows the callback.
- Docs already warn that `flush()` does not work with `pino-pretty` due to cross-thread limitations. This is a **different** issue: the reproduction above uses `pino/file`, and the log data does reach disk. Only the completion signal is lost.

## Notes

I have a working local fix. Happy to open a PR — the natural place feels like `thread-stream`'s `requestWorkerFlush` (ref the worker while a flush callback is pending, unref once the callbacks map is empty), which would fix the issue at the root and benefit all `thread-stream` users, not just Pino. But I wanted to check with maintainers first whether that's the preferred direction, or whether a Pino-side workaround in `lib/proto.js`'s `flush()` would be more appropriate.

Hashim1999164:
## Summary
- Transport workers are unref'd after ready so the process can exit, which meant `logger.flush(cb)` never got its completion message on an idle event loop.
- Keep the worker referenced while a flush callback is in flight, and skip the ready-time unref if a flush is already pending.
- Add an idle-loop regression fixture and drop the old test-side keep-alive workaround.

Fixes #2481

## Test plan
- [x] `npx borp --timeout 60000 test/transport/sync-false.test.js test/syncfalse.test.js test/exit.test.js test/transport/core.test.js` (51 passed)
- [x] `node test/fixtures/transport-flush-idle.js` prints `callback-fired` and exits 0

### DIFF
--- a/lib/transport.js
+++ b/lib/transport.js
@@ -143,6 +143,29 @@ function buildStream (filename, workerData, workerOpts, sync, name) {
     sync
   })
 
+  // Keep the worker referenced while a flush(cb) is in flight. The worker is
+  // unref'd after ready so the process can exit, but that also means an idle
+  // event loop can drain before the flush-completion message is delivered.
+  // See https://github.com/pinojs/pino/issues/2481
+  const originalFlush = stream.flush.bind(stream)
+  let flushPending = 0
+  stream.flush = function (cb) {
+    if (typeof cb !== 'function') {
+      return originalFlush(cb)
+    }
+    flushPending++
+    if (flushPending === 1) {
+      stream.ref()
+    }
+    originalFlush(function (err) {
+      flushPending--
+      if (flushPending === 0) {
+        stream.unref()
+      }
+      cb(err)
+    })
+  }
+
   stream.on('ready', onReady)
   stream.on('close', function () {
     process.removeListener('exit', onExit)
@@ -152,7 +175,10 @@ function buildStream (filename, workerData, workerOpts, sync, name) {
 
   function onReady () {
     process.removeListener('exit', onExit)
-    stream.unref()
+    // Do not unref while a flush callback is waiting for the worker.
+    if (flushPending === 0) {
+      stream.unref()
+    }
 
     if (workerOpts.autoEnd !== false) {
       setupOnExit(stream)
--- a/test/fixtures/transport-flush-idle.js
+++ b/test/fixtures/transport-flush-idle.js
@@ -1,0 +1,34 @@
+'use strict'
+
+const fs = require('node:fs')
+const { join } = require('node:path')
+const { tmpdir } = require('node:os')
+const pino = require('../..')
+
+const dest = join(tmpdir(), `pino-flush-idle-${process.pid}.log`)
+fs.writeFileSync(dest, '')
+
+const transport = pino.transport({
+  target: 'pino/file',
+  options: { destination: dest }
+})
+
+const logger = pino(transport)
+logger.info({ marker: 'first-log' })
+
+let callbackFired = false
+logger.flush((err) => {
+  callbackFired = true
+  if (err) {
+    console.error(err)
+    process.exitCode = 1
+    return
+  }
+  console.log('callback-fired')
+})
+
+process.on('exit', () => {
+  if (!callbackFired) {
+    console.error('callback-missing')
+  }
+})
--- a/test/transport/sync-false.test.js
+++ b/test/transport/sync-false.test.js
@@ -3,8 +3,10 @@
 const test = require('node:test')
 const assert = require('node:assert')
 const os = require('node:os')
+const { tmpdir } = require('node:os')
 const { join } = require('node:path')
 const { readFile } = require('node:fs').promises
+const { unlink } = require('node:fs/promises')
 const { promisify } = require('node:util')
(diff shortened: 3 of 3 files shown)


aquie00t:
Thanks for jumping on this — the ref-during-flush approach is exactly right!

One design question for the maintainers before this lands: the root cause
actually lives in `thread-stream`, where `requestWorkerFlush` registers the
callback in `flushCallbacks` and posts the FLUSH message, but the worker is
already unref'd, so an idle event loop drains before the completion message
arrives.

Fixing it there — ref the worker while any callback is in `flushCallbacks`,
unref once the map empties — would:

- keep pino's `lib/proto.js` unchanged (no new state for tracking pending flushes)
- automatically fix the same behavior for other thread-stream consumers
  (pino-elasticsearch, custom transports, etc.)
- respect the encapsulation: pino calls `stream.flush(cb)` and trusts
  the stream to honor the callback contract

Not a blocker for this PR — just wanted to surface the alternative in case
the deeper fix feels cleaner. Happy to help either way.

Hashim1999164:
Thanks for the thread-stream perspective. Happy to keep this PR as the pino-side workaround if maintainers prefer shipping the fix here first, or reshape it if the preferred path is fixing ref/unref in `thread-stream` instead. Looking for maintainer guidance either way.

test/fixtures/transport-flush-idle.js
```diff
@@ -0,0 +1,34 @@
+'use strict'
+
+const fs = require('node:fs')
+const { join } = require('node:path')
+const { tmpdir } = require('node:os')
+const pino = require('../..')
+
+const dest = join(tmpdir(), `pino-flush-idle-${process.pid}.log`)
```
kilisamemarisaaa, line 8: This fixture leaves `pino-flush-idle-<pid>.log` in the system temp directory on every run. Because the parent test spawns a fresh child process each time, the PID-based names do not provide cleanup and the files accumulate across local/CI runs.

Could the parent test remove this path after `await child` (for example in a `finally` block, using `child.pid` to reconstruct it)? Cleaning up from this child before exit would keep work on the otherwise-idle event loop and weaken the regression this fixture is meant to prove.

Hashim1999164: The parent test already removes that file after the child exits.

`test/transport/sync-false.test.js` reconstructs `pino-flush-idle-${child.pid}.log` and unlinks it in a `finally` block so the fixture can stay idle through flush.

lib/transport.js
```diff
@@ -143,6 +143,29 @@ function buildStream (filename, workerData, workerOpts, sync, name) {
...
+      stream.ref()
+    }
+    originalFlush(function (err) {
+      flushPending--
+      if (flushPending === 0) {
+        stream.unref()
+      }
+      cb(err)
+    })
+  }
```
mcollina, line 167: This should be fixed on thre thread-stream side

Hashim1999164: Thanks. I will move the ref around flush into thread-stream so the worker stays referenced until the FLUSHED message is delivered.

I can drop this pino wrapper once thread-stream owns that behavior.

Hashim1999164:
heads up, someone already put this on thread-stream: https://github.com/pinojs/thread-stream/pull/249

i can drop the pino wrapper once that lands. bro i dont know if you want this one closed now or kept until theirs merges lmao

PR #2493, opened 2026-08-05
fix(multistream): fix clone add crash and missing end/remove methods

spellsaif:
This PR resolves multiple runtime exceptions and TypeScript typing mismatches associated with the multistream.js API in multistream.js.
  
  Specifically, it ensures cloned multistreams are fully compatible with their original counterpart's interface and behaves correctly as a builder pattern.
  
## Issues Addressed
  
  1. Crash on Cloned multistream.js with String Levels:
  Cloned multistream objects were missing the streamLevels mapping. Invoking multistream.js with string-based log levels (e.g. 'info') threw a TypeError: Cannot read properties of undefined (reading 'info').
  2. Missing Lifecycle multistream.js on Cloned Multistream:
  The cloned instance omitted the multistream.js method, causing crashes when streams were shut down.
  3. TypeScript Typings Mismatch:
  The pino.d.ts interface in pino.d.ts did not declare the public .end() and .remove() methods.
  4. Fluent API Discrepancy:
  Calling multistream.js or multistream.js on a cloned multistream incorrectly returned the original res instance instead of the cloned one (this), breaking clone isolation.             
  
## Changes
  
  • **multistream.js**:
      • Propagated streamLevels, end, and emit properties inside the multistream.js return block.
      • Updated multistream.js and multistream.js methods to return this instead of the closure-scoped original res.
  • **pino.d.ts**:
      • Exposes .end() and .remove() signatures on the pino.d.ts interface.
  • **multistream.test.js**:
      • Added a dedicated unit test suite (clone supports add, remove, and end methods correctly) to verify fluent returns, successful stream additions, and correct stream lifecycle     
      endings.
  

  ## Verification
  
  • Ran eslint checking and all style rules pass.
  • Executed the unit test suite and verified the newly added tests successfully validate the changes without regressions.

### DIFF
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -105,7 +105,7 @@ function multistream (streamsArray, opts) {
 
   function add (dest) {
     if (!dest) {
-      return res
+      return this
     }
 
     // Check that dest implements either StreamEntry or DestinationStream
@@ -141,7 +141,7 @@ function multistream (streamsArray, opts) {
 
     this.minLevel = streams[0].level
 
-    return res
+    return this
   }
 
   function remove (id) {
@@ -154,7 +154,7 @@ function multistream (streamsArray, opts) {
       this.minLevel = streams.length > 0 ? streams[0].level : -1
     }
 
-    return res
+    return this
   }
 
   function end () {
@@ -180,12 +180,14 @@ function multistream (streamsArray, opts) {
       write,
       add,
       remove,
+      emit,
+      flushSync,
+      end,
       minLevel: level,
       streams,
       clone,
-      emit,
-      flushSync,
-      [metadata]: true
+      [metadata]: true,
+      streamLevels
     }
   }
 }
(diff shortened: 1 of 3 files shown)


luantaraschi (approved):
I checked the clone behavior and the public type surface. A clone now carries the level map used by `add`, exposes the same lifecycle methods as the original multistream, and returns its own receiver from `add` and `remove`. The original multistream still returns itself through those same functions because they are called with the original object as `this`.

I ran `test/multistream.test.js` with all 32 tests passing. The complete type check also passed: seven TSTyche files with 91 assertions, the runtime type fixture, and `attw` all completed without errors. ESLint on the changed files and `git diff --check` passed as well.

spellsaif:
Awesome, thanks for the thorough review! Glad to hear everything is passing and the clone/type behavior looks good. :)

mcollina (approved):
lgtm

PR #2494, opened 2026-08-06
docs: add serializers documentation

### LINKED ISSUE #478: Missing documentation for serializers API
We are totally missing documentation regarding how to write a serializer. We have a note at http://getpino.io/#/docs/api?id=serializers-object about the function needing to be synchronous and return a JSONifiable object, but nothing about the actual API interface to implement.

bilashcse:
## What

Adds `docs/serializers.md`, a dedicated page documenting the serializer interface, and links it from the docsify sidebar and the README documentation list.

## Why

Closes #478. As noted there, the docs described the constraints on a serializer (synchronous, returns a JSONifiable value) but never the interface itself. @davidmarkclements suggested a `serializers.md` page in that thread; this is that page.

## What it covers

The function signature and the value a serializer receives; that serializers apply to own top-level keys of the merging object and are skipped when the value is `undefined`; the `errorKey` / `err` fallback; serializing the message via `messageKey`; child logger inheritance and override; a pointer to `pino.stdSerializers`; and that `Symbol.for('pino.*')` is deprecated in favour of `formatters.log`.

Behaviour described is taken from `lib/tools.js` (L166-173, L203-204), `lib/proto.js` (L115-136) and `pino.js` (L111-113).

Documentation only, no code changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@
 * [API ⇗](/docs/api.md)
 * [Browser API ⇗](/docs/browser.md)
 * [Redaction ⇗](/docs/redaction.md)
+* [Serializers ⇗](/docs/serializers.md)
 * [Child Loggers ⇗](/docs/child-loggers.md)
 * [Transports ⇗](/docs/transports.md)
 * [Diagnostics ⇗](/docs/diagnostics.md)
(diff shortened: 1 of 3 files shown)


jsumners (commented):
Serializers are documented at https://github.com/pinojs/pino/blob/eff2b30ea52cb73ec683541972ae57b331626bdf/docs/api.md?plain=1#L460-L474

PR #2496, opened 2026-08-11
browser: forward remaining log arguments to write in asObject mode

### LINKED ISSUE #2244: browser: logger does not pass arguments to write function
when logging this way:
```ts
const log = logger('context')

log.debug('bla', { di: 'da' })
```

the arguments do not end up anywhere in the write function:
```ts
...
     browser: {
      write: (logObj, ...logParams) => {
        const { level, msg, time } = logObj as Record<string, string>
        const timeFormatted = new Date(time).toISOString()
        const levelColor = levelColors[level as keyof typeof levelColors]
        console.log(`[${timeFormatted}] ${levelColor}${level} ${color.cyan} ${msg} ${color.white}`, ...logParams)
      },
      formatters: {
        level: label => {
          return {
            level: label,
          }
        },
      },
    },
...
```

`arguments` always has length 1. Neither `logObj` contains them, and logParams is always undefined.

SnowingFox:
Fixes #2244

## Problem

When a custom `browser.write` is used (which forces `asObject` mode), extra arguments passed to the log methods never reach the `write` function:

```js
const logger = pino({ browser: { write: (logObj, ...logParams) => { /* ... */ } } })
logger.debug('bla', { di: 'da' })
```

Here `{ di: 'da' }` is silently dropped: `logParams` is always empty because `asObject()` only returns the formatted log object.

`asObjectBindingsOnly` mode already forwards the remaining arguments (`return [formattedLogObject, ...argsCloned]`), but the regular `asObject` branch dropped them.

## Change

Mirror the `asObjectBindingsOnly` branch in `browser.js` and return the remaining (unconsumed) arguments alongside the formatted log object:

```js
return [formattedLogObject, ...argsCloned]
```

## Verification

Added a regression test in `test/browser.test.js`:

- `logger.info('test', { di: 'da' })` now calls `write(logObj, { di: 'da' })`.

The new test fails on `main` and passes with this change:

```
npx tape test/browser.test.js
```

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -449,7 +449,7 @@ function asObject (logger, level, args, ts, opts) {
     if (opts.redactFn) {
       formattedLogObject = opts.redactFn(formattedLogObject)
     }
-    return [formattedLogObject]
+    return [formattedLogObject, ...argsCloned]
   }
 }
 
--- a/test/browser.test.js
+++ b/test/browser.test.js
@@ -203,6 +203,24 @@ test('opts.browser.asObjectBindingsOnly passes the bindings but keep the message
   end()
 })
 
+test('opts.browser.write func passes additional arguments when asObject is enabled', ({ end, ok, is, deepEqual }) => {
+  const instance = pino({
+    browser: {
+      asObject: true,
+      write: function (o, ...args) {
+        is(o.level, 30)
+        is(o.msg, 'test')
+        ok(o.time)
+        deepEqual(args, [{ di: 'da' }])
+      }
+    }
+  })
+
+  instance.info('test', { di: 'da' })
+
+  end()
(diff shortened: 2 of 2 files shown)


luantaraschi (approved):
I reproduced the reported behavior against the parent commit: the custom `write` function received only the formatted log object. At this head it also receives the untouched `{ di: 'da' }` argument, while the formatted object still contains the expected level, message, and timestamp.

I also ran the complete `test/browser.test.js` file (196 assertions), ESLint on the changed files, and `git diff --check`. All passed. The change is limited to forwarding the arguments that remain after the log object is formatted, consistent with the existing `asObjectBindingsOnly` path.

PR #2499, opened 2026-08-11
test: use fs rename for cross-platform transpilation

marktran0710:
## What does this PR do?

Use `fs.promises.rename` instead of the Unix-only `mv` command in the TypeScript fixture transpilation helper.

## Why?

`npm run test-ci` failed on Windows because `mv` is not available as a native command.

## Testing

- `npm run lint`
- `npm run test-types`
- `npm run transpile`
- `npx borp test/transport/core.test.js --timeout 60000`
- `git diff --check`

All checks passed.

This change keeps the behavior the same while making the transpilation step cross-platform.

### DIFF
--- a/test/fixtures/ts/transpile.cjs
+++ b/test/fixtures/ts/transpile.cjs
@@ -5,6 +5,7 @@ const fs = require('node:fs')
 
 const existsSync = fs.existsSync
 const stat = fs.promises.stat
+const rename = fs.promises.rename
 
 // Hardcoded parameters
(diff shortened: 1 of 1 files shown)


luantaraschi (approved):
Tested this on Windows with Node 22.20.0. On the current main branch, `npm run transpile` fails because `mv` is unavailable. At this commit, the same command completes successfully. I also ran `npm run lint` and `git diff --check`, and both passed. Using `fs.promises.rename` keeps the file operation within Node and fixes the cross-platform failure. I did not find any issues with the change.

PR #2504, opened 2026-08-18
Add bindings() method to browser logger

### LINKED ISSUE #1177: bindings() and setBindings() are not present in the browser version of pino
Hello! 

I was attempting to use pino in the browser today and discovered that `.bindings()` is undefined. Additionally, `.setBindings()` is also `undefined`, but it's also undocumented in general (is it officially part of the API?).

I looked into adding this to browser.js but I'd definitely need some guidance before proceeding. For example, I'm not sure if  `bindings` changes cascade through child loggers or if they are fixed at the time of child creation, for example.

RT1811:
## Summary

Adds `bindings()` support to the browser logger to match the Node implementation.

- Adds `bindings()` to browser logger instances
- Preserves bindings inherited when child loggers are created
- Ensures later parent `setBindings()` calls do not affect existing children
- Keeps internal binding state separate from the public `bindings()` method
- Adds browser tests for bindings behavior

Fixes #1177

## Testing

- Added `test/browser-bindings.test.js`
- Existing `browser-set-bindings.test.js` tests pass
- `CI=1 npm test` passes

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -49,6 +49,9 @@ function levelToValue (level, logger) {
 const baseLogFunctionSymbol = Symbol('pino.logFuncs')
 const hierarchySymbol = Symbol('pino.hierarchy')
 
+const ownBindingsSymbol = Symbol('pino.ownBindings')
+const bindingsSymbol = Symbol('pino.bindings')
+
 const logFallbackMap = {
   error: 'log',
   fatal: 'error',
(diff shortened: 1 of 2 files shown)


RT1811:
Hi! I noticed the CI workflow expired while awaiting approval, so the tests haven’t run yet. Could a maintainer please restart the workflow and approve it when they have a chance?

I’m happy to address any feedback or test failures once the checks run. Thanks for your time!

PR #2506, opened 2026-08-22
fix(multistream): keep stream ids on clone so remove() works

chuanghiduoc:
## Problem

`clone(level)` rebuilt each stream entry with only `{ level, stream }`, dropping the `id` assigned by `add()`. Every inherited entry on the clone then had `id: undefined`, so the documented `remove(id)` could never match the entry the caller asked for:

```js
const base = multistream([A, B])
const clone = base.clone('trace')

const entry = clone.streams.find(s => s.stream === B)
console.log(entry.id)        // undefined
clone.remove(entry.id)       // findIndex(s => s.id === undefined) -> no-op
                             // (or strips an arbitrary undefined-id entry)
```

The instance-level test added in #2493 passed because it removed a stream that had been **added to the clone** (which gets a fresh id); inherited entries were never covered.

## Fix

Copy the source entry's `id` into cloned entries. Adding to a clone still assigns fresh ids from the shared counter, so uniqueness across base and clones is preserved.

## Testing

- extended `clone supports add, remove, and end methods correctly` in `test/multistream.test.js` to assert every cloned entry keeps a numeric id
- `node --test test/multistream.test.js`: 32 passing
- full suite: 537 unit tests passing (one unrelated `transport/big.test.js` timeout under parallel load; passes in isolation)

### DIFF
--- a/lib/multistream.js
+++ b/lib/multistream.js
@@ -172,6 +172,11 @@ function multistream (streamsArray, opts) {
     for (let i = 0; i < streams.length; i++) {
       streams[i] = {
         level,
+        // preserve the assigned id so remove() keeps working on clones;
+        // without it every inherited entry has `id: undefined` and
+        // remove() would strip an arbitrary entry instead of the
+        // requested one.
+        id: this.streams[i].id,
         stream: this.streams[i].stream
       }
     }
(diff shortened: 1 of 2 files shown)


official-burak (commented):
Copying `id` on clone is the right fix. `remove()` is `findIndex(s => s.id === id)`, so inherited entries with `id: undefined` either no-op or drop an arbitrary row.

The new test only asserts `typeof entry.id === 'number'`. It never calls `clone.remove(inheritedId)` and checks that stream is gone. The existing `clone supports add, remove, and end` case still only removes a stream that was `add`ed to the clone, which already had an id.

A remove-on-inherited-id assertion would actually cover the bug.

PR #2509, opened 2026-08-28
feat(browser): support nestedKey in asObject mode and browser.write

lx3133584:
## Problem

When using `pino` in browser / isomorphic environments with `browser.asObject: true` (or custom `browser.write`), the `nestedKey` configuration option was ignored. Logged payload objects were assigned directly to the top-level log object rather than nested under `nestedKey`, creating an inconsistency between Node.js and browser runtimes.

## Root Cause

- `setOpts` in `browser.js` did not store `nestedKey`.
- `asLogObject` unconditionally called `Object.assign(logObject, argsCloned.shift())` for all object arguments instead of nesting payload objects under `logObject[opts.nestedKey]` while keeping bindings at the top level.

## Fix

- Stored `nestedKey: opts.nestedKey || opts.browser.nestedKey || null` in `setOpts`.
- Updated `asLogObject` to assign payload objects (`lvl === 0`) to `logObject[opts.nestedKey]` while assigning binding objects (`lvl > 0`) to the top-level `logObject`.

## Testing

- Added unit tests in `test/browser.test.js` verifying that `nestedKey` nests payload objects under the configured key in root and child loggers.
- Verified all 198 browser tests and lint pass cleanly.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -142,6 +142,7 @@ function pino (opts) {
     levels,
     timestamp: getTimeFunction(opts),
     messageKey: opts.messageKey || 'msg',
+    nestedKey: opts.nestedKey || opts.browser.nestedKey || null,
     onChild: opts.onChild || noop,
     redactFn
   }
@@ -426,7 +427,10 @@ function asObject (logger, level, args, ts, opts) {
   if (opts.asObjectBindingsOnly) {
     if (msg !== null && typeof msg === 'object') {
       while (lvl-- && typeof argsCloned[0] === 'object') {
-        Object.assign(logObject, argsCloned.shift())
+        const target = (lvl === 0 && opts.nestedKey)
+          ? (logObject[opts.nestedKey] = logObject[opts.nestedKey] || {})
+          : logObject
+        Object.assign(target, argsCloned.shift())
       }
     }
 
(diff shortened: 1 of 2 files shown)


browser.js
```diff
@@ -439,7 +443,10 @@ function asObject (logger, level, args, ts, opts) {
     // deliberate, catching objects, arrays
     if (msg !== null && typeof msg === 'object') {
       while (lvl-- && typeof argsCloned[0] === 'object') {
-        Object.assign(logObject, argsCloned.shift())
+        const target = (lvl === 0 && opts.nestedKey)
+          ? (logObject[opts.nestedKey] = logObject[opts.nestedKey] || {})
```
kilisamemarisaaa, line 447: This reuses a same-named child binding as the payload container. On this head, `child({ payload: 'binding' }).info({ foo: 'bar' }, 'hello')` emits `{"payload":"binding","msg":"hello"}` in browser `asObject` mode, so the logged object is lost; the Node logger emits `payload: { foo: "bar" }`.

If the binding is an object, `Object.assign` mutates that shared binding instead: two successive logs with `{ first: 1 }` and `{ second: 2 }` make both already-collected records end up with `payload: { binding: true, first: 1, second: 2 }`.

Please create a fresh nested payload object for each log (overriding a colliding binding) rather than using `||`, and add regression cases for primitive/object collisions plus two consecutive log calls.

kilisamemarisaaa (commented):
I found one correctness issue in the collision path. I reproduced it against this head with Node 24.12.0 and compared the emitted objects with the regular Node logger behavior; details are inline.

PR #2519, opened 2026-09-04
fix logging edge cases

marktran0710:
## Summary

Fix several logging edge cases affecting JSON output, child bindings, transport level validation, and browser serializers.

## Changes

- Escape custom `messageKey` values to keep emitted logs valid JSON.
- Support child bindings named `hasOwnProperty` and null-prototype objects.
- Correct multi-target transport formatter validation to use the configured level mappings.
- Prevent browser serializers from mutating caller-owned log objects.
- Add regression tests for all fixes.

## Tests

- `npm run lint`
- `npm run test-types`
- Core and transport tests
- Browser and browser transmit tests
- Jest tests
- Eight-million-line stress test

All targeted tests passed.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -218,7 +218,9 @@ function pino (opts) {
         ? Object.keys(childSerializers)
         : serialize
       delete bindings.serializers
-      applySerializers([bindings], childSerialize, childSerializers, this._stdErrSerialize)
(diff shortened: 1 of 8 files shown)


mcollina (changes requested):
Please send each on its own pr

PR #2520, opened 2026-09-04
fix message key escaping

marktran0710:
## Summary

Escape custom `messageKey` values to ensure emitted log records remain valid JSON.

## Changes

- Precompute the escaped message key for each logger instance.
- Prevent malformed output and log injection when `messageKey` contains quotes or control characters.
- Add a regression test covering malicious key content.

## Tests

- `npm run lint`
- `node --test test/escaping.test.js`

### DIFF
--- a/lib/symbols.js
+++ b/lib/symbols.js
@@ -24,6 +24,7 @@ const stringifiersSym = Symbol('pino.stringifiers')
 const endSym = Symbol('pino.end')
 const formatOptsSym = Symbol('pino.formatOpts')
 const messageKeySym = Symbol('pino.messageKey')
+const messageKeyStrSym = Symbol('pino.messageKeyStr')
 const errorKeySym = Symbol('pino.errorKey')
 const nestedKeySym = Symbol('pino.nestedKey')
 const nestedKeyStrSym = Symbol('pino.nestedKeyStr')
@@ -62,6 +63,7 @@ module.exports = {
   endSym,
   formatOptsSym,
   messageKeySym,
+  messageKeyStrSym,
   errorKeySym,
   nestedKeySym,
   wildcardFirstSym,
--- a/lib/tools.js
+++ b/lib/tools.js
@@ -21,6 +21,7 @@ const {
   nestedKeySym,
   formattersSym,
   messageKeySym,
+  messageKeyStrSym,
   errorKeySym,
   nestedKeyStrSym,
   msgPrefixSym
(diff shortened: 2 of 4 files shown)


mcollina (approved):
lgtm

jsumners (commented):
I don't see why we need to put a `JSON.stringify` in this path.

marktran0710:
> I don't see why we need to put a `JSON.stringify` in this path.

Okay, I will remove it

marktran0710:
@jsumners Here is the reason why I add `JSON.stringify`
Keeping `JSON.stringify(messageKey)` here is intentional.

For a normal key:

```js
pino({ messageKey: 'msg' }).info('hello')
```

`JSON.stringify('msg')` returns `"msg"`, so the output remains:

```json
{"msg":"hello"}
```

For a key containing special characters:

```js
pino({ messageKey: 'x"}\n{"fake":true}' }).info('hello')
```

`JSON.stringify(messageKey)` safely produces:

```json
{"x\"}\n{\"fake\":true}":"hello"}
```

Without escaping, the raw key can break the JSON structure or inject additional log records.

The neighboring object keys are escaped through `asString(key)` during serialization. `messageKey` is a fixed logger option, so it is escaped once during initialization and cached for performance. The escaping behavior is the same; only the timing is different.

jsumners (commented):
This complicates things for no reason. The application author is in control of defining the message key. There's no way to set it to an invalid value except by choice.

marktran0710:
> This complicates things for no reason. The application author is in control of defining the message key. There's no way to set it to an invalid value except by choice.

So what is your suggestion? Remove `JSON.stringify` and use` [messageKeyStrSym]:messageKey`

jsumners:
> > This complicates things for no reason. The application author is in control of defining the message key. There's no way to set it to an invalid value except by choice.
> 
> So what is your suggestion? Remove `JSON.stringify` and use` [messageKeyStrSym]:messageKey`

I don't think anything needs to be done. I have not seen any reports of this being a problem.

marktran0710:
> > > This complicates things for no reason. The application author is in control of defining the message key. There's no way to set it to an invalid value except by choice.
> > 
> > 
> > So what is your suggestion? Remove `JSON.stringify` and use` [messageKeyStrSym]:messageKey`
> 
> I don't think anything needs to be done. I have not seen any reports of this being a problem.
okay, so i will closed it

PR #2521, opened 2026-09-05
docs: document the APIs that are unavailable in the browser

### LINKED ISSUE #2032: The `pino.destination()` function is not defined on a browser (bug and/or documentation issue)
This is related to or the same issue as https://github.com/pinojs/pino/issues/607, but I'm using Parcel rather than WebPack.

```
utils.js:25 Uncaught TypeError: (0 , _pinoDefault.default).destination is not a function
```

Issue #607 was closed and locked, but I don't feel like it was properly resolved. The issue here isn't really that `destination` doesn't *work* on a browser, it's understandable why that would be the case, it's that calling it throws an error and the fact that it isn't available isn't documented. The page linked from the other issue is https://getpino.io/#/docs/browser, which mentions the word "destination" only once but isn't talking about this method. The documentation for `destination()` at https://getpino.io/#/docs/api?id=pino-destination also don't mention that it isn't available in the browser.

The issue I ran into, is that I have some code that is shared between the backend and the frontend. In the browser this code doesn't need to do anything with the destination, it's just logging to the console, but the code that is setting up the backend logging is still in there. And when this code runs, that's a hard error and everything stops.

There are a few ways to work around this, I went with:
```js
if (!pino.destination) pino.destination = () => undefined; // Fix for browser
```

I could have also just put an `if (pino.destination)` around the location it's being used, or I could use try/catch, etc. This is basically fine, albeit a little annoying. But I don't think it should behave this way without warning/documentation.

**Suggested resolutions:**
- Document which features and functions are completely unavailable for the browser on that page. Include advice for code that needs to run in either environment.
- Note that a feature or function isn't available in browser in its section of the API docs.
- And/or, provide a shim such as the one above that simply no-ops the function to avoid causing an error, or logs a warning so the developer is aware of the issue, or throws an explicit error with a clear message.

kavish-19:
Fixes #2032.

`pino.destination()`, `pino.transport()` and `pino.multistream()` are not part of the browser build, so calling them in the browser throws:

```
TypeError: pino.destination is not a function
```

Neither `docs/browser.md` nor the API reference mentioned this, so the failure reads as a bundler problem rather than an intentional limit. That was the substance of #2032, where you replied:

> I think having a no-op would be problematic, but this certainly needs documenting. Would you like to open a PR?

The reporter never opened one, so here it is. Documentation only — no source changes, and no no-op shim, per your comment.

## What this adds

A **Node.js-only APIs** section in `docs/browser.md` covering:

- the three statics that are absent from the browser build, and why;
- that `pino.levels`, `pino.stdSerializers` and `pino.stdTimeFunctions` *are* available in both;
- two behaviours that were also undocumented — `logger.flush()` is a no-op whose callback is never invoked, and the `EventEmitter` methods are no-ops, so `level-change` never fires in the browser;
- guidance for isomorphic code, including that the browser `pino()` takes the options object only and silently ignores a second `destination` argument.

Plus five cross-references in `docs/api.md`, on `pino.destination()`, `pino.transport()`, `pino.multistream()`, `logger.flush([cb])` and `Event: 'level-change'`.

## Verification

Every claim was checked by running `browser.js` against `pino.js` rather than read off the source:

| Claim | Result |
| --- | --- |
| `destination` / `transport` / `multistream` absent in browser, present in Node | confirmed |
| `levels` / `stdSerializers` / `stdTimeFunctions` present in both | confirmed |
| browser `pino.destination()` throws `TypeError` | confirmed |
| browser `flush()` exists, callback never invoked (sync or after a tick) | confirmed |
| `level-change` never fires in browser | confirmed |
| `level-change` does fire in Node | confirmed (control) |
| browser `pino()` arity is 1, second argument ignored | confirmed |

That last one corrected a draft of this PR: I had assumed the workaround in #2032 (`pino.destination = () => undefined`) would leave the logger silently discarding lines. It doesn't — the browser build ignores the second argument and logging continues to the console. The text now documents the real behaviour.

`npm run lint` passes.

## Deliberately out of scope

`pino.symbols`, `pino.version` and `logger.bindings()` are also missing from the browser build, but #2508 and #2504 are changing that. Documenting them as absent would be stale on merge, so they are left alone.

Happy to adjust the wording, drop the isomorphic-code guidance if you would rather the docs stay purely descriptive, or split the `api.md` notes into a separate PR.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1093,6 +1093,9 @@ If there is a need to wait for the logs to be flushed, a callback should be used
 **Note:** `flush()` does not work when using `pino-pretty`. See
 [Flush Limitations with `pino-pretty`](/docs/asynchronous.md) for more details.
 
+**Note:** in the browser this method is a no-op and `cb` is never invoked. See
+[Node.js-only APIs](/docs/browser.md#nodejs-only-apis).
+
 * See [`destination` parameter](#destination)
 * See [Asynchronous Logging ⇗](./asynchronous.md)
 
@@ -1190,6 +1193,10 @@ register its own serializer upon instantiation the serializers of the parent wil
 
 The logger instance is also an [`EventEmitter ⇗`](https://nodejs.org/dist/latest/docs/api/events.html#events_class_eventemitter)
 
+**Note:** the `EventEmitter` methods are no-ops in the browser, so this event
+never fires there. See
+[Node.js-only APIs](/docs/browser.md#nodejs-only-apis).
+
 A listener function can be attached to a logger via the `level-change` event
 
 The listener is passed five arguments:
@@ -1245,6 +1252,8 @@ Exposes the cumulative `msgPrefix` of the logger.
 Create a Pino Destination instance: a stream-like object with
 significantly more throughput than a standard Node.js stream.
 
+**Note:** this is not available in the browser build. See [Node.js-only APIs](/docs/browser.md#nodejs-only-apis).
+
 ```js
 const pino = require('pino')
 const logger = pino(pino.destination('./my-file'))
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

